Issue #42, opened 2020-08-14
Faster OSS


blefaudeux:
## 🚀 Feature
Reduce the runtime cost tied to activating OSS

## Motivation
Right now the speed penalty of using OSS over a couple of nodes, for a typical ClassyVision load is 2-10x. 

## Pitch
There are a bunch of low hanging fruits, from oss_ddp to broadcast batching which could be used to reduce this impact

## Alternatives
Not using OSS I guess

PR #43, opened 2020-08-14
[feat] batch broadcast requests into a configurable buffer

blefaudeux:
# Before submitting

## What does this PR do?
Improves on #42 by batching the small broadcasts into a bigger one. There's a tradeoff in between doing more copies and incurring less latency on the communication side, so the broadcast buffer and a size above which the broadcast is direct are configurable.
This is a long lived PR so some listed commits are unrelated and come from merges with upstream master over time.

## PR review

## Did you have fun?


fairscale/optim/utils.py
```diff
@@ -4,12 +4,18 @@
...
 
 import io
-from typing import Any, Dict
+from typing import TYPE_CHECKING, Any, Dict, List
 
 import torch
 from torch._six import container_abcs
 import torch.distributed as dist
 
+if TYPE_CHECKING:  # pragma: no cover
```
msbaines, line 13: Why is this conditional?

fairscale/optim/oss.py
```diff
@@ -97,9 +109,32 @@ def step(self, closure: Optional[Callable[[], float]] = None) -> Optional[float]
...
+            # Batch smaller params in a broadcast buffer to speed up the communication
+            buffered_params: List[Parameter] = []
+            buffered_elements = 0
+
             for param_group in param_groups:
                 for param in param_group["params"]:
-                    dist.broadcast(tensor=param, src=rank, group=self.group)
+                    if param.numel() >= self._buffer.numel():
+                        # Big param block, broadcast directly
+                        dist.broadcast(tensor=param, src=rank, group=self.group)
```
msbaines, line 120: batch buffers need to be per-device to avoid device-to-device transfers

fairscale/optim/utils.py
blefaudeux: copy paste from oss.py, my guess was that it was to catch strict or loose type hinting

fairscale/optim/oss.py
blefaudeux: ah interesting, I did not think of that. Thanks, looking into it

fairscale/optim/oss.py
```diff
@@ -65,6 +76,7 @@ def __init__(self, params: _params_t, optim: Type[Optimizer] = SGD, group: Any =
 
         # Current device is set by the parameters allocated to this rank
         self._device = self.partition_parameters()[self.rank][0]["params"][0].device
+        self._buffer = torch.zeros(buffer_size).to(self._device)
```
blefaudeux, line 79: @msbaines  could you elaborate on the batch buffer which needs to be per-device ? I'm guessing given your comment that this is not enough, but it's not obvious to me

msbaines: The parameters could be stored on multiple devices (e.g. gpipe) so we need a batch buffer for each device.

msbaines: > The parameters could be stored on multiple devices (e.g. gpipe) so we need a batch buffer for each device.

See oss_ddp.py

blefaudeux:
This should come after the benchmark PR (https://github.com/facebookresearch/fairscale/pull/47), I merged early to be able to use the benchmarks here

benchmarks/oss.py
```diff
@@ -119,7 +119,7 @@ def closure():
     print(f"[{dist.get_rank()}] : Mean speed: {mean:.2f} +/- {std:.2f}")
 
     if use_oss and check_regression and dist.get_rank() == 0:
-        assert (mean - 3.0 * std) < reference_speed, "Speed regression detected"
+        assert (mean + 3.0 * std) > reference_speed, "Speed regression detected"
```
blefaudeux, line 122: this was actually wrong before, we want speed to be equal or better to the baseline, not lower...

benchmarks/oss.py
```diff
@@ -133,11 +133,12 @@ def closure():
     parser.add_argument("--epochs", action="store", default=10, type=int)
     parser.add_argument("--batch_size", action="store", default=32, type=int)
     parser.add_argument("--data_size", action="store", default=512, type=int)
-    parser.add_argument("--check_regression", action="store", default=True, type=bool)
-    parser.add_argument("--reference_speed", action="store", default=39.82, type=float)
+    parser.add_argument("--check_regression", action="store_true", default=False)
+    parser.add_argument("--reference_speed", action="store", default=33, type=float)
```
blefaudeux, line 137: a previous PR switched the optimizer from SGD to RMSProp, which is a tad slower. I checked that this 33fps value is not degraded by this PR

fairscale/optim/oss.py
```diff
@@ -44,12 +45,25 @@ class OSS(Optimizer):
             optimizer to shard (default: SGD)
         group (group):
             torch.distributed group (default: group.WORLD)
+        buffer_size (int, optional): number of elements to buffer before
```
msbaines, line 48: What does optional mean in this context? The parameter does not look like an optional.

fairscale/optim/oss.py
```diff
@@ -67,6 +81,12 @@ def __init__(self, params: _params_t, optim: Type[Optimizer] = SGD, group: Any =
         # Current device is set by the parameters allocated to this rank
         self._device = split_param_groups[self.rank][0]["params"][0].device
 
+        # Broadcast buffer settings
+        self._buffer: Optional[torch.Tensor] = None
```
msbaines, line 85: parameters cannot change after init so you could pre-process the params and pre-create the batch buffers.

fairscale/optim/oss.py
```diff
@@ -102,9 +122,41 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
 
         # Sync all the states
         for rank, param_groups in enumerate(self.partition_parameters()):
+            # Batch smaller params in a broadcast buffer to speed up the communication
+            buffered_params: List[Parameter] = []
+            buffered_elements = 0
+
             for param_group in param_groups:
+                # We assume that the whole param group is on the same device
```
msbaines, line 130: This is not a safe assumption. gpipe we have params on multiple devices within the same group.

msbaines (changes requested):
It is not safe to assume all params in a param_groups are on the same device.

fairscale/optim/oss.py
blefaudeux: ouch, yes of course, it's an old comment but I stuck to it in the logic, that's pretty bad. Sorry for the dumb submission, fixing that

fairscale/optim/oss.py
blefaudeux: I meant to write that people are free to pass it in or not, there's a default provided

blefaudeux:
Now deduplicating in between oss_ddp and oss @min-xu-ai, slightly tighter. I think that it would be nicer to adapt the buffering parameters to the loaded model, that's something we know at construction time (very large model  and few shards -> adapt the buffer size)

blefaudeux:
Something else that I'm planning is to do is to pre-sort the parameters per size at construction time, so that the step and oss_ddp logic is simpler, as suggested by Min

PR #46, opened 2020-08-20
[fix] OSS restore state to proper device

blefaudeux:
# Before submitting

## What does this PR do?
Fixes the state pull and push behavior to being something more easily predictable, when a state is pulled it gets moved to CPU (failsafe if this is a reasonably big model). Upon restore the `param_groups` were not restored to the proper device. It looks like there are still probable duplicates in that field though

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -149,7 +149,7 @@ class OSS(Optimizer):
 
         # Restore the global param_groups
         self.param_groups = recursive_copy_to_device(
-            state_dict["param_groups"], non_blocking=False, device=self._device
+            state_dict["param_groups"], non_blocking=True, device=self._device
         )
 
     def add_param_group(self, param_group: dict) -> None:
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -148,9 +148,7 @@ class OSS(Optimizer):
         self.load_local_state_dict(state_dict["state"][self.rank])
 
         # Restore the global param_groups
-        self.param_groups = recursive_copy_to_device(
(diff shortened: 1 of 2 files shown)


fairscale/optim/oss.py
```diff
@@ -148,7 +148,7 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
         self.load_local_state_dict(state_dict["state"][self.rank])
 
         # Restore the global param_groups
-        self.param_groups = state_dict["param_groups"]
+        self.param_groups = recursive_copy_to_device(state_dict["param_groups"], non_blocking=True, device=self._device)
```
min-xu-ai, line 151: Line too long?

Is there a test for this already? If so, should there be some assert in the test?

blefaudeux: no test, it's a good point, I'll add that

msbaines: We follow PyTorch coding style of 120 columns:

https://github.com/pytorch/pytorch/blob/master/.flake8#L3

It is enforced here:

https://github.com/facebookresearch/fairscale/blob/master/setup.cfg#L29

blefaudeux: it's supposed to be automatic with vscode+black, it's pure FB tooling and 120 lines, this formatting crap is so annoying.. I'll fix that in the next PR

blefaudeux: ah no ok, black is happy with that indeed, it's 121 cols but for some reason this passes (same for the init above, also 121). Anyway, all good then

PR #61, opened 2020-09-02
[fix] OSS pytorch-compliant state dict

### LINKED ISSUE #60: Rewrite state_dict in a more pytorch idiomatic way
## 🚀 Feature
Change the param_groups handling in the state dict, in order to follow more closely the default PyTorch assumptions
https://pytorch.org/docs/stable/optim.html#torch.optim.Optimizer.state_dict

## Motivation
- Some users may assume that the default pytorch optimizer interface with respect to the state dict is respected by fairscale/oss
- People familiar with pytorch optimizers would have an easier learning curve when peeking into OSS

## Pitch
Rewrite the exposed state dict in order to return "state" and "param_groups" in accordance to pytorch expectations, without duplications

## Alternatives
- rely on the python/pytorch memory model to remove duplicates in memory and while serializing
- add wrappers on the user side

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #60 .

## PR review

## Did you have fun?
Except for the realization that all my lint errors were due to isort and black being version-dependent in their behaviour (FFS), all good 🙃

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -4,8 +4,8 @@
 # LICENSE file in the root directory of this source tree.
 
 import copy
-import logging
 from itertools import chain
+import logging
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type
 
 import torch
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -4,8 +4,8 @@
 # LICENSE file in the root directory of this source tree.
 
 import copy
-from itertools import chain
 import logging
+from itertools import chain
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type
 
 import torch
@@ -58,14 +58,14 @@ class OSS(Optimizer):
         # Build the wrapped optimizer, responsible for a shard of the params
         self.group = group
         self.rank = dist.get_rank(group)
-        param_groups = self.partition_parameters()
-        self.optim = optim(param_groups[self.rank], **defaults)
+        split_param_groups = self.partition_parameters()
+        self.optim = optim(split_param_groups[self.rank], **defaults)
 
         # Optional consolidated optimizer state
         self._all_states: List[Dict[str, Any]] = []
 
         # Current device is set by the parameters allocated to this rank
-        self._device = self.partition_parameters()[self.rank][0]["params"][0].device
+        self._device = split_param_groups[self.rank][0]["params"][0].device
 
     def partition_parameters(self) -> List[List[dict]]:
         """Partitions parameters across distributed ranks.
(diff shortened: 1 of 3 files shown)


fairscale/optim/oss.py
```diff
@@ -140,7 +140,10 @@ def state_dict(self) -> Dict[str, Any]:
             len(self._all_states) > 0
         ), "The optimizer state is not materialized, please call consolidate_state_dict on every replica beforehand"
 
-        return {"state": self._all_states}
+        return {
+            "state": [s["state"] for s in self._all_states],
+            "param_groups": [s["param_groups"] for s in self._all_states],
```
msbaines, line 145: Remove param_groups from _all_states to avoid having a redundant copy?

msbaines (commented):
What if you un-sharded in state-dict and then re-sharded in load_state_dict? That way, we work exactly like PyTorch is expecting.

fairscale/optim/oss.py
blefaudeux: we don't own this stat dict object, it's returned and could have any lifetime, I don't see the issue with a redundant copy ?

blefaudeux:
> What if you un-sharded in state-dict and then re-sharded in load_state_dict? That way, we work exactly like PyTorch is expecting.

The pytorch expectation is not super clear to me, param_groups is a list (even if described as a dict), there does not seem to be a lot more constraints in https://pytorch.org/docs/stable/_modules/torch/optim/optimizer.html#Optimizer.state_dict than that.

You mean unrolling the list so that param_groups is `[dict()]` ? I can do that

fairscale/optim/oss.py
msbaines: A redundant copy won't consume memory since they are just references but I'm wondering if it will increase disk space. Would be good to check that. I'm also wondering what will happen on load. Will the two copies refer to the same objects or will there be two copies. That could cause OOM.

OSS created the state_dicts in _collect_sharded_states. There are no references outside OSS at this point.

blefaudeux: aaaah, I see what you mean, but it's not related to this PR, right ? In general I agree, _all_states could be changed to remove duplicates, I agree, but to me this is a relatively independent change, I was trying to address the get/load state interface only. In the same fashion, it feels to me like there must be duplicates in between `self.param_groups`, `self._all_states` and `self.optim.param_groups`

blefaudeux: re: lifetime, I was refering to the object you commented on, `return {}`, I thought that was the subject of your comment

msbaines: I'm worried about this potentially introducing a regression to fairseq. They are now using this in master and I tested with only a single copy of the parameters. Maybe just confirm that disk space does not increase and we don't get duplicate tensors for the params on load. Might be cool to add a memory benchmark (@andersonic added a memory assertion in one of the gpipe benchmarks). If the duplicates are referencing the same object then it is not an issue.

msbaines (approved):

PR #65, opened 2020-09-04
[feat] OSS flatten state dict

blefaudeux:
# Before submitting

## What does this PR do?
Changes the structure of the returned state dict with respect to the param_groups to make it closer to what a vanilla optimizer would return (un-shard them). Shard again when loading

## PR review

## Did you have fun?

### DIFF
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -119,7 +119,7 @@ def train(
     print(f"[{dist.get_rank()}] : Mean speed: {mean:.2f} +/- {std:.2f}")
 
     if use_oss and check_regression and dist.get_rank() == 0:
-        assert (mean - 3.0 * std) < reference_speed, "Speed regression detected"
+        assert (mean + 3.0 * std) > reference_speed, "Speed regression detected"
         assert max_memory < 1.05 * reference_memory, "Memory use regression detected"
         print("[Regression Test] VALID")
 
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -49,8 +49,8 @@ def train(
     # Data setup, dummy data
     def collate(inputs: List[Any]):
         return {
-            "inputs": torch.stack([i[0] for i in inputs]).to(rank),
-            "label": torch.stack([i[1] for i in inputs]).to(rank),
+            "inputs": torch.stack([i[0] for i in inputs]).to(torch.device(rank)),
+            "label": torch.stack([i[1] for i in inputs]).to(torch.device(rank)),
         }
 
     dataloader = DataLoader(
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -133,11 +133,12 @@ if __name__ == "__main__":
     parser.add_argument("--epochs", action="store", default=10, type=int)
     parser.add_argument("--batch_size", action="store", default=32, type=int)
     parser.add_argument("--data_size", action="store", default=512, type=int)
-    parser.add_argument("--check_regression", action="store", default=True, type=bool)
-    parser.add_argument("--reference_speed", action="store", default=39.82, type=float)
+    parser.add_argument("--check_regression", action="store_true", default=False)
+    parser.add_argument("--reference_speed", action="store", default=32.32, type=float)
     parser.add_argument("--reference_memory", action="store", default=4475, type=float)
 
     args = parser.parse_args()
+    print(f"Benchmark arguments: {args}")
 
     print("\nBenchmark vanilla optimizer")
     mp.spawn(
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -146,20 +146,19 @@ class OSS(Optimizer):
             len(self._all_states) > 0
         ), "The optimizer state is not materialized, please call consolidate_state_dict on every replica beforehand"
 
-        # Flatten the state_dict, save the partition
-        # The partition indexes the flat param_groups list,
-        # and logs the rank <> shard correspondence
-        partition: Dict[int, Tuple[int, int]] = {}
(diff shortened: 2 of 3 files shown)


blefaudeux:
cc @mannatsingh

fairscale/optim/oss.py
```diff
@@ -146,9 +146,22 @@ def state_dict(self) -> Dict[str, Any]:
             len(self._all_states) > 0
         ), "The optimizer state is not materialized, please call consolidate_state_dict on every replica beforehand"
 
+        # Flatten the state_dict, save the partition
+        # The partition indexes the flat param_groups list,
+        # and logs the rank <> shard correspondence
+        partition: Dict[int, Tuple[int, int]] = {}
```
msbaines, line 152: Why not use a list since the keys are 0 to number of partitions?

fairscale/optim/oss.py
```diff
@@ -146,9 +146,22 @@ def state_dict(self) -> Dict[str, Any]:
...
+        param_groups: List[Dict[Any, Any]] = []
+
+        i_global = 0
+        for i, s in enumerate(self._all_states):
+            param_groups.extend(s["param_groups"])
+            partition[i] = (i_global, i_global + len(s["param_groups"]))
+            i_global = partition[i][1]
```
msbaines, line 159: Might be simpler as:
```suggestion
            end = i_global + len(s["param_groups"])
            partition[i] = (i_global, end)
            i_global = end
```

fairscale/optim/oss.py
```diff
@@ -146,9 +146,22 @@ def state_dict(self) -> Dict[str, Any]:
             len(self._all_states) > 0
         ), "The optimizer state is not materialized, please call consolidate_state_dict on every replica beforehand"
 
+        # Flatten the state_dict, save the partition
+        # The partition indexes the flat param_groups list,
+        # and logs the rank <> shard correspondence
+        partition: Dict[int, Tuple[int, int]] = {}
+        param_groups: List[Dict[Any, Any]] = []
+
+        i_global = 0
```
msbaines, line 155: This name is confusing. Maybe call it `start` instead?

fairscale/optim/oss.py
```diff
@@ -146,9 +146,22 @@ def state_dict(self) -> Dict[str, Any]:
             len(self._all_states) > 0
         ), "The optimizer state is not materialized, please call consolidate_state_dict on every replica beforehand"
 
+        # Flatten the state_dict, save the partition
```
msbaines, line 149: The code doesn't seem to flatten the state dict. It just converts the list to a dictionary. Is this comment correct?

blefaudeux: the flattening starts l153, I can move the comment down

fairscale/optim/oss.py
blefaudeux: ah, it was meant as "global index", but start works I guess

fairscale/optim/oss.py
blefaudeux: sure, just three lines to increment a counter, that's why I was trying to keep this compact. I miss the `index++` notation in C++ at times to be honest

msbaines: readability > compactness

fairscale/optim/oss.py
msbaines: My concern is more that the comment does not seem to be totally consistent with the code. We don't really flatten state_dict, we just flatten the param_groups part of the state_dict. state itself is not flattened. Instead we convert the state field to a dictionary with integer keys. Why not flatten that also? If it can't be flattened, please add a comment as to why.

blefaudeux: ah ok, agreed, I thought it was more about the localization, I can definitely change the comment. I'm not sure of the benefit of flattening the state else, I don't know of any use case, so it was more a matter of not doing something not needed, but I can do the same for sure

fairscale/optim/oss.py
blefaudeux: taste and colours, I was just trying to give the initial reasoning. Completely fine by me with a longer version

benchmarks/oss.py
```diff
@@ -119,7 +119,7 @@ def closure():
     print(f"[{dist.get_rank()}] : Mean speed: {mean:.2f} +/- {std:.2f}")
 
     if use_oss and check_regression and dist.get_rank() == 0:
-        assert (mean - 3.0 * std) < reference_speed, "Speed regression detected"
+        assert (mean + 3.0 * std) > reference_speed, "Speed regression detected"
```
blefaudeux, line 122: this was wrong, fixed in another PR but might as well be fixed here (I was bumping into this locally)

blefaudeux: (we want the speed to increase, not decrease.. the test initially made sense when comparing runtime, now that it compares frames per second higher is better)

benchmarks/oss.py
```diff
@@ -133,11 +133,12 @@ def closure():
     parser.add_argument("--epochs", action="store", default=10, type=int)
     parser.add_argument("--batch_size", action="store", default=32, type=int)
     parser.add_argument("--data_size", action="store", default=512, type=int)
-    parser.add_argument("--check_regression", action="store", default=True, type=bool)
-    parser.add_argument("--reference_speed", action="store", default=39.82, type=float)
+    parser.add_argument("--check_regression", action="store_true", default=False)
+    parser.add_argument("--reference_speed", action="store", default=32.32, type=float)
```
blefaudeux, line 137: 39 was the default speed for SGD, I had changed that earlier to RMSProp when checking for the memory pressure, somehow this change was lost

msbaines (approved):

Issue #63, opened 2020-09-03
Add more documentation


blefaudeux:
## 🚀 Feature
- Surface all the features exposed by fairscale
- Add code snippets
- Add benchmark examples
- Add links towards other projects which are using fairscale already
- Add some thoughts on when and where use the different methods exposed by fairscale

## Motivation
Fairscale is hard to grok from the outside, lower the barrier of entry

## Pitch
Easier to onboard people, easier to increase traction, clean things up

## Alternatives
live in a cave

blefaudeux:
cc @msbaines @froody @min-xu-ai

froody:
We should also add guidance around picking optimal parameters. For example with gpipe, the right value for the 'chunks' parameter depends on the model size.

VitaliyLi:
would be also nice to add a discussion on when to use specific features, ie when to use pipelining and when to use tensor parallelism

blefaudeux:
> We should also add guidance around picking optimal parameters. For example with gpipe, the right value for the 'chunks' parameter depends on the model size.

some frameworks expose the per param group flops, that's an approximation (ie not all flops take the same time to compute), but would that still be a useful heuristic, or should the size (in memory) be taken into account ? for my general understanding

blefaudeux:
closed for now, to be revived if need be

PR #79, opened 2020-09-10
[feat] adding a README example for OSS

blefaudeux:
# Before submitting

## What does this PR do?
Improves on #63.

## PR review

## Did you have fun?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,16 +48,14 @@ def train(
     model.train()
     for e in range(epochs):
         for batch in dataloader:
-            def closure():
-                model.zero_grad()
-                outputs = model(batch["inputs"])
-                loss = loss_fn(outputs, batch["label"])
-                torch.distributed.all_reduce(loss, op=torch.distributed.ReduceOp.SUM)
-                loss /= world_size
-                loss.backward()
-                return loss
-
-            optimizer.step(closure)
+            # Train
+            model.zero_grad()
+            outputs = model(batch["inputs"])
+            loss = loss_fn(outputs, batch["label"])
+            torch.distributed.all_reduce(loss, op=torch.distributed.ReduceOp.SUM)
+            loss /= world_size
+            loss.backward()
+            optimizer.step()
 
 if __name__ == "__main__":
     # supposing that WORLD_SIZE and EPOCHS are somehow defined somewhere
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ fairscale supports:
 * optimizer state sharding (fairscale.optim.oss)
 
 ## Examples
+### Pipe
 
 Run a 4-layer model on 2 GPUs. The first two layers run on cuda:0 and the next two layers run on cuda:1.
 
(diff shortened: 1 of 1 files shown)


blefaudeux:
pnig review, if you don't mind

README.md
```diff
@@ -19,6 +20,59 @@ model = torch.nn.Sequential(a, b, c, d)
...
+    world_size: int,
+    epochs: int):
+
+    # DDP
+    dist_init(rank, world_size)
+
+    # Problem statement
+    model = myAwesomeModel()
+    dataloader = mySuperFastDataloader()
+    loss = myVeryRelevantLoss()
```
min-xu-ai, line 41: nice names!

README.md
```diff
@@ -19,6 +20,59 @@ model = torch.nn.Sequential(a, b, c, d)
...
+            def closure():
+                model.zero_grad()
+                outputs = model(batch["inputs"])
+                loss = loss_fn(outputs, batch["label"])
+                torch.distributed.all_reduce(loss, op=torch.distributed.ReduceOp.SUM)
+                loss /= world_size
+                loss.backward()
+                return loss
+
+            optimizer.step(closure)
```
min-xu-ai, line 60: using closure is not the most common way for pytorch optimizers, right? Perhaps this example can be simplified by not using closure?

min-xu-ai (approved):
nice!

README.md
blefaudeux: depends on people, but closures is (to my understanding) considered a little safer because the scope is tight. The scope in python is leaky (like, something defined in a for loop leaks outside of it), so closures bring some sanity. It also makes it compatible with optimizers which require multiple evaluations, both options (with and without closures) are in the pytorch doc https://pytorch.org/docs/stable/optim.html#taking-an-optimization-step

min-xu-ai: Completely agree. I was just suggesting that it is not the most common way people do it (AFAIK) (perhaps because most programmers are not very used to it). For an initial example, perhaps keep it simple is better. Of course, if you want to use this opportunity to advocate the usage of closure style, it is fine too. In that case, perhaps some comments in the code to explain what's going on with closure and why it is better?

blefaudeux: follow up: closure removed !

README.md
```diff
@@ -19,6 +20,57 @@ model = torch.nn.Sequential(a, b, c, d)
...
+
+    # Any relevant training loop, nothing specific to OSS. For example:
+    model.train()
+    for e in range(epochs):
+        for batch in dataloader:
+            # Train
+            model.zero_grad()
+            outputs = model(batch["inputs"])
+            loss = loss_fn(outputs, batch["label"])
+            torch.distributed.all_reduce(loss, op=torch.distributed.ReduceOp.SUM)
```
min-xu-ai, line 55: is this assuming ddp? perhaps in the simplest form, each rank was given the same batch and they don't need to reduce losses? Sorry, I didn't notice this until now. Maybe I am missing something?

README.md
blefaudeux: I think it's an orthogonal subject to zero, so probably best to not mix things up. Fine by me as it is, no worries, I'll keep using closures for my own code though :)

README.md
blefaudeux: yes assuming ddp, I thought that it was reasonable for zero to be useful ? for instance we torch.dist.broadcast the state shards, so ddp needs to be there. I'm actually not clear on whether it could be useful without ddp, one could probably imagine something, but my assumption was that de facto people interested would come from a ddp-enabled background ?

min-xu-ai: of course. I would love to try closure next time myself too.

My question above, though, is about the all_reduce on loss. Is that necessary? Is that due to the assumption that here we are doing ddp training?

min-xu-ai: our msg crossed. :-)

I think ddp is more useful with oss for sure. But without ddp it can useful to extend the model size too. (i.e. use 10 GPUs to train a model that is 5X bigger but not much speed gain in terms of sample/s.

In any case, this is find. I was just double checking.

PR #90, opened 2020-09-16
Multi-process pipe

froody:
# Before submitting

- [X] Did you write any new necessary tests?

## What does this PR do?
Adds support for distributing pipeline stages across multiple processes (and therefore multiple machines).
* Adds a `style` argument to the `Pipe` constructor, defaulting to `PipelineStyle.SingleProcess`, but also supporting `PipelineStyle.MultiProcess`
* Added support for lazy construction of modules (see lazy_construction for an example)
* Added two implementations of inter-process communication: one based on rpc with globally visible queues, one based on send/recv
* Copied all the relevant tests from tests/pipe to tests/pipe_process and modified them to exercise PipelineStyle.MultiProcess

## Did you have fun?


msbaines (approved):
Looks great! Please remove print messages and commented out code. Also, please 'import logging' if you want to have debug messages.

fairscale/nn/pipe/pipe.py
```diff
@@ -123,18 +201,11 @@ def split_module(
...
-            "module and sum of balance have different length "
-            f"(module: {len(module)}, sum of balance: {sum(balance)})"
-        )
-
-    if any(x <= 0 for x in balance):
-        raise BalanceError(f"all balance numbers must be positive integer (balance: {balance})")
+    check_balance(module, balance)
 
-    if len(balance) > len(devices):
+    if devices and len(balance) > len(devices):
```
sidgoyal78, line 206: minor nit: we could pass `style` as an argument to this function and avoid using `devices` to compare? 

Same comment for `verify_splitting` method.

sidgoyal78 (commented):
Thanks for the PR, Tom. I put a few inline comments.

fairscale/nn/pipe/pipe.py
froody: this is hard because these functions aren't methods of Pipeline, I think in this case it's ok because it's literally just checking the value of devices, not branching on some more complex behaviour.

sidgoyal78: Sounds good, thanks.

PR #97, opened 2020-09-17
[feat] Sharded DDP - small refactor and new features

### LINKED ISSUE #91: Privatize the sharded optimizer within the sharded DDP
## 🚀 Feature
Align the OSS_ddp architecture with this proposal https://github.com/pytorch/pytorch/issues/42849.
A main difference is that oss_ddp owns the construction of the sharded optimizer, it does not have to be exposed.

## Motivation
- Prototype something closer to the RFC
- Bring the sharding in the same place, so that we can eventually align the sharding from the optimizer and the model parameter sharding. Right now DDP gets the full model

## Pitch
Keep backward compatibilty: using OSS directly stays possible, but if targeting oss_ddp, then it owns the sharded optimizer. Eventually surface the sharding and align both

## Alternatives
Keep the current approach in two seperate steps (shard DDP / shard optim), but it does not seem very future proof

blefaudeux:
# Before submitting

## What does this PR do?
- Fixes #91, the sharded DDP creates the sharded optimizer by itself, and relies on its partition.
- Start to align the names with the [Pytorch/ZeRO RFC](https://github.com/pytorch/pytorch/issues/42849)
- Remove duplicated calls, cache some results (partition_parameters() for instance)
- Expose the partition as optimizer attributes
- Add a benchmark for Sharded DDP, not regression testing for now as this is evolving fast

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -21,7 +21,7 @@ from torch.nn import Parameter
 from fairscale.optim import OSS
 
 
-class OssDdp(nn.Module):
+class ShardedDataParallel(nn.Module):
     """Implements distributed data parallel training with optimizer state sharding.
 
     A simplified version of :class:`torch.nn.parallel.DistributedDataParallel`.
@@ -30,7 +30,8 @@ class OssDdp(nn.Module):
 
     Args:
         module (~torch.nn.Module): module to be parallelized
-        oss (fairscale.optim.OSS): shared state optimizer
+        optimizer (~torch.optim.Optimizer): optimizer to be used for training
+        optimizer_params(Dict): extra parameters for the optimizer
         world_size (int): number of parallel workers
         process_group (optional): the c10d process group to be used for
             distributed gradient reduction. If None, the default WORLD process group
@@ -86,7 +87,7 @@ class OssDdp(nn.Module):
     def optimizer(self) -> torch.optim.Optimizer:
         return self.sharded_optimizer
 
-    def train(self, mode: bool = True) -> "OssDdp":
+    def train(self, mode: bool = True) -> "ShardedDataParallel":
         pre_mode = self.module.training
         self.module.train(mode)
         if self.module.training:
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -71,6 +71,12 @@ class ShardedDataParallel(nn.Module):
         # Build the sharded optimizer
         self.sharded_optimizer = OSS(self.module.parameters(), optim=optimizer, group=process_group, **optimizer_params)
 
+        # The gradients will be routed to the shards directly, remove the top level buffers
+        for pg in self.sharded_optimizer.param_groups:
+            for param in pg["params"]:
+                if param.requires_grad:
+                    param.grad = None
+
         # Handle the heterogeneous communication / sharding. The sharded optimizer owns the partitions
         self.param_rank = self.sharded_optimizer.param_to_rank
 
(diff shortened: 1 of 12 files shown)


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -176,10 +163,9 @@ def reduce_params(params: List[Parameter], params_rank: int) -> None:
                         p.grad = buffer[offset : offset + sz].view_as(p).clone()
                     offset += sz
             else:
-                # zero the grads
+                # wipe the grads
                 for p in params:
-                    if p.grad is not None:
-                        p.grad.data.zero_()
+                    p.grad = None
```
min-xu-ai, line 168: perhaps could be done with map(lambda...) function too.

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -39,16 +38,30 @@ def run_one_step(rank, world_size, backend, device, temp_file_name):
...
     ddp.reduce()
+
+    # Check that all the grads have been populated, for the shard
+    if device == torch.device("cuda"):
+        torch.cuda.synchronize()  # flush any remaining cuda op, just in case
+
+    for pg in optimizer.optim.param_groups:
+        for param in pg["params"]:
+            if param.requires_grad:
+                assert param.grad.abs().sum().item() > 0.0, "The reduce step should have populated all the gradients"
```
min-xu-ai, line 59: nice change in the test!

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -9,26 +9,19 @@
...
 
-if TYPE_CHECKING:
-    from fairscale.optim import OSS
-    from torch import Tensor
-    from torch.nn import Parameter
-else:
-    OSS = Any
-    Tensor = Any
-    Parameter = Any
+from fairscale.optim import OSS
```
min-xu-ai, line 21: is OSS going to be renamed to ShardOptimizer at some point?

min-xu-ai (approved):
Looks really nice to me!

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: yes I would love to, at least to match the RFC entirely. On FB side this means that we would need to touch classy and lightning code, but I was not sure about fairseq so I erred on the side of caution and did not touch it. @msbaines , thoughts ?

fairscale/optim/oss.py
```diff
@@ -51,30 +53,37 @@ class OSS(Optimizer):
...
+        self._device = self.partition_parameters()[self.rank][0]["params"][0].device
 
         # Sync local and global param_groups keys
         for global_group, local_group in zip(self.param_groups, self.optim.param_groups):
             for k, v in local_group.items():
                 if k != "params":
                     global_group[k] = v
 
+    # Partition helpers
     def partition_parameters(self) -> List[List[dict]]:
```
blefaudeux, line 87: I think that was a little slow, because being re-run for all step(). Now it's cached and re-run only when we modify the param_groups

fairscale/optim/oss.py
```diff
@@ -83,21 +92,52 @@ def partition_parameters(self) -> List[List[dict]]:
...
+
+                for rank, params in enumerate(param_lists):
+                    param_group_rank = copy.copy(param_group)
+                    param_group_rank["params"] = params
+                    self._partition_parameters[rank].append(param_group_rank)
+
+        return self._partition_parameters
+
+    @property
+    def per_device_params(self) -> List[List[Parameter]]:
```
blefaudeux, line 114: trying to keep all the partition related tidbits in the same place, and caching whenever possible

fairscale/optim/oss.py
blefaudeux: could have been a property by the way, but I was not sure of the fairseq usecase, whether this was being accessed or not (cc @msbaines )

PR #102, opened 2020-09-22
Making oss gloo-compliant, check that it will stay the case

blefaudeux:
# Before submitting

## What does this PR do?
Fixes OSS being broken with the Gloo backend, rightfully so (in place operation on a tensor which requires_grad, undefined). Adds a CI test for Gloo so that we don't miss issues like this in the future

## PR review

## Did you have fun?

### DIFF
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -19,14 +19,14 @@ from torchvision.transforms import ToTensor
 from fairscale.nn.data_parallel import ShardedDataParallel
 from fairscale.optim.oss import OSS
 
-BACKEND = dist.Backend.NCCL if torch.cuda.is_available() else dist.Backend.GLOO  # type: ignore
 OPTIM = torch.optim.RMSprop
 
 
-def dist_init(rank, world_size):
+def dist_init(rank, world_size, backend):
+    print(f"Using backend: {backend}")
     os.environ["MASTER_ADDR"] = "localhost"
     os.environ["MASTER_PORT"] = "29501"
-    dist.init_process_group(backend=BACKEND, rank=rank, world_size=world_size)
+    dist.init_process_group(backend=backend, rank=rank, world_size=world_size)
 
 
 def get_problem(rank, data_size, batch_size):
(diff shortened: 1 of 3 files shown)


fairscale/optim/oss.py
```diff
@@ -153,12 +153,22 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
         for rank, param_groups in enumerate(self.partition_parameters()):
             for param_group in param_groups:
                 for param in param_group["params"]:
+                    # NOTE: Broadcast is in-place and not differentiable
+                    # Gloo will rightly assert on this operation for any tensor that requires grad.
+                    # We save and restore the grad requirement state to work around that, in our case
+                    # the grad is only useful on the source rank.
+                    requires_grad.append((param, param.requires_grad))
+                    param.requires_grad = False
                     requests.append(dist.broadcast(tensor=param, src=rank, group=self.group, async_op=True))
```
min-xu-ai, line 166: it seems that you can chain the assignment using future.then() function with a lambda function?

min-xu-ai (approved):
LGTM

benchmarks/oss.py
```diff
@@ -210,29 +211,31 @@ def closure():
...
         )
     else:
-        print("\nBenchmark vanilla optimizer")
-        mp.spawn(
-            train,
-            args=(args.world_size, args.epochs, args.batch_size, args.data_size, False, False),
-            nprocs=args.world_size,
-            join=True,
-        )
+        # print("\nBenchmark vanilla optimizer")
```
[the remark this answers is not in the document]
blefaudeux: by mistake, fixed in the commit after that normally ? unless I forgot to push..

fairscale/optim/oss.py
blefaudeux: ah I didn't know that it was supported with the futures that pytorch returns ? I'll have a look

min-xu-ai: also here: https://pytorch.org/docs/stable/futures.html#torch.futures.collect_all

wait_all() can be used in place of a loop.

blefaudeux: yes for wait_all, I guess that it's a loop under the hood but cleaner indeed ! I"m not sure that the then() formulation proposed will work though, lambdas are stateless in python so you typically cannot change a member value

min-xu-ai: I thought you can pass in to `then()` a closure which contains the reference to the `param` variable, then you can freely change its memory the original value. It could avoid the extra list entirely. But I could be wrong.

blefaudeux: not possible unfortunately, I just tried, the futures linked above are from the RPC framework, not pytorch.distributed

blefaudeux: the only methods supported [here](https://pytorch.org/docs/stable/distributed.html#synchronous-and-asynchronous-collective-operations) are .wait() and is_completed()

PR #120, opened 2020-10-01
[fix] OSS: Eager gradient release - free memory

blefaudeux:
# Before submitting

## What does this PR do?
- Release the gradients for the model parameters as soon as every rank computes its update, at this point only the gradients for this rank's shard are useful. This frees 100MB of memory for a small model (rn101) & 2 ranks, no other consequence.
- (unrelated) Adjust the circleci config to use the 4 allocated GPUs, probably a more realistic usecase

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -157,6 +157,8 @@ class OSS(Optimizer):
         self._sync_param_groups()
 
         # Run the optimizer step on this shard only:
+        self._free_other_grads()
+
         if closure is not None:
             loss = self.optim.step(closure=closure, **kwargs)  # type: ignore
         else:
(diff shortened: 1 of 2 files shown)


blefaudeux:
[This](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/511/workflows/08fa3bfb-6bfc-459e-917d-c390f7218565/jobs/1909) shows the apples to apples comparison (the regression test is tuned to the previous memory use, 4475MB for this configuration). 
The two next commits adjust CI to use 4 ranks, this is unrelated and consequently adjust the regression test, the two items are unrelated.

fairscale/optim/oss.py
```diff
@@ -367,3 +369,14 @@ def _broadcast_state_dict(self) -> None:
                 # Discard this tensor/rank, broadcast necessary for syncing
                 logging.debug("Discarding broadcast from rank %s", rank)
                 broadcast_object(empty_buffer, src_rank=rank, group=self.group, dist_device=self._device)
+
+    def _free_other_grads(self) -> None:
```
min-xu-ai, line 373: nice find. perhaps we can do this while doing reduce with other rank? I.e. we can immediately free the grad after they are used? feel free to defer to a separate change.

min-xu-ai (approved):
minor suggestion.

fairscale/optim/oss.py
blefaudeux: in sharded_ddp you mean ? Yes I think that there are tons of optimizations to be done there, for now it's pretty slow, which is to be expected. Here I just wanted to fix it in the general case, when step() is called with the normal DDP

Issue #117, opened 2020-09-29
[OSS] Faster broadcast in step()


blefaudeux:
## 🚀 Feature
Bucketize the param broadcasts, this step is a pain point empirically

## Motivation
Speed, no compromise other than some code complexity, easy decision.

## Pitch
Decide parameter buckets at construction time, bucketize the smallest params but keep the biggest ones async
 
## Alternatives
Status quo, too slow

## Additional context
Right now for a typical computer vision training task, OSS takes a 30% speed hit when going multi node

blefaudeux:
cc @min-xu-ai same as state dict consolidation, up for grabs or happy to dive into this, just a heads up

PR #122, opened 2020-10-01
[feat] OSS/SDP : bucketing

blefaudeux:
# Before submitting

## What does this PR do?
Tentative improvement on #117, bucket the smallest parameters in a flat tensor, directly communicate the other ones. The per-device parameters returned by OSS are now also sorted per rank, and then sorted per size, which simplifies the bucketing. End result is a big speedup for ShardedDataParallel, and arguably more readable broadcast and reduce calls.

The speed benefit across multiple nodes would need to be tested 

## PR review

## Did you have fun?


blefaudeux:
Test on 4-gpus: see [baseline](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/518/workflows/ce74d5d8-674a-4f62-bd08-4ebadbefed2b/jobs/1939/parallel-runs/0/steps/0-110) vs [PR](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/520/workflows/25652690-e1a7-4630-b920-81a71112d786/jobs/1947/parallel-runs/0/steps/0-110)

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -39,7 +39,7 @@ class ShardedDataParallel(nn.Module):
             distributed gradient reduction. If None, the default WORLD process group
             will be used.
         buffer_size (int, optional): number of elements to buffer before
-            performing reduce (default: 256M). Used to reduce multiple small
+            performing reduce (default: 1M). Used to reduce multiple small
```
min-xu-ai, line 42: Curious, how is 1M default determined?

fairscale/optim/oss.py
```diff
@@ -175,25 +197,63 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+            offset = 0
+            while i_p < len(rank_params) and offset + rank_params[i_p].numel() < buffer_size:
+                end = offset + rank_params[i_p].numel()
+                if rank == self_rank:
+                    buffer[offset:end].copy_(rank_params[i_p].data.view(-1))  # type: ignore
+                offset = end
+                i_p += 1
+
+            if i_p > 0:
+                dist.broadcast(tensor=buffer, src=global_rank, group=group, async_op=True)
```
min-xu-ai, line 232: corner case: this could be the last param, which means async_op needs to == True?

fairscale/optim/oss.py
```diff
@@ -175,25 +197,63 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
-                    requires_grad.append((param, param.requires_grad))
+            if len(rank_params) == 0:
+                continue
+
+            global_rank = OSS.get_global_rank(group, rank)
+
+            # Copy small gradients into per-GPU buffers and then async broadcast
+            i_p = 0
+            offset = 0
+            while i_p < len(rank_params) and offset + rank_params[i_p].numel() < buffer_size:
```
min-xu-ai, line 224: maybe I am not understanding the code right, but I thought you need an outer loop here to broadcast using the buffer potentially multiple times. I.e.

- say your buffer size is 10
- your param sizes are [2,3,3,4,4,5, 6]
- you need to pack 2,3,3 in the buffer once and then 4,4 again using the buffer and then potentially skip the single param buffer cases, etc.

I think it would be nice to refactor the code that does bucket send in a separate function and unit test it for different corner cases. What do you think?

min-xu-ai (commented):
I have some questions. could be that I am not reading it correctly.

fairscale/optim/oss.py
blefaudeux: it's not that practical to loop over the same buffer, because it means that it cannot be async anymore (or you need to allocate many more buffers), to me it's the key issue with this approach, ditching async or needed more buffers are big drawbacks. 

So yes the proposal is to change the logic, it buckets the smallest items (we can increase the buffer size if we want to cover more of them) and sends it async, and the rest is sent as-is (not bucketed), async also except for the last one which is a gatekeeper. 

Empirically it's significantly faster

fairscale/optim/oss.py
blefaudeux: good point, that could be what the unit test is reacting to, fixing that !

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: I think that it's something we should tweak probably, I just changed it after a discussion with @msbaines who was mentioning batching kb params, and because on 4 gpus and training a RN101 the zero savings are around 400MB, so buffering 256MB is probably not worth it.

fairscale/optim/oss.py
blefaudeux: (second point) yes for the factoring of this bucketing function, you're right, would be more readable and easier to test. Doing that

min-xu-ai: Ah, got it. Can you please document it if not already? I feel that this will be very model dependent. If the param list has a lot of small tensors, then it can be put some in a single buffer, but the rest will be sent one by one again.

With async, does the comm library merge small messages? If so, perhaps async means we don't really need any buffer? Maybe keep it simple for now and add/decide buffering strategy later? But let me know if you want to keep as it. I can approve this version if so.

blefaudeux: > Ah, got it. Can you please document it if not already? I feel that this will be very model dependent. If the param list has a lot of small tensors, then it can be put some in a single buffer, but the rest will be sent one by one again.

I think that if the model has many small params, we're toasted anyway, bucketing or not. We can kind of speed up the coms if we bucket, but that requires many buffers to keep requests mostly async, and then the whole zero point (memory saving) is not that valuable. In any case bucketing over and over is not a silver bullet I think

> With async, does the comm library merge small messages? If so, perhaps async means we don't really need any buffer? Maybe keep it simple for now and add/decide buffering strategy later? But let me know if you want to keep as it. I can approve this version if so.

NCCL does not seem to bucket by itself, traces show some of the broadcast calls in the 80us range 
Screen Shot 2020-10-01 at 4 04 34 PM
A motivation for this is that currently OSS is slow when many nodes are involved, around 30% throughput hit for computer vision workloads, and the broadcast step is a big culprit (purple bar in the attached trace). We can test for real whether these changes go in the right direction of course, the benchmarks on CircleCI are not enough

min-xu-ai (approved):

blefaudeux:
Testing on FBLearner : f222484263

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -126,85 +131,69 @@ def reduce(self) -> None:
...
+        for rank, rank_params in enumerate(params):
+            if len(rank_params) == 0:
+                continue
+
+            restore_require_grad = []
+            requests = []
+            global_rank = OSS.get_global_rank(group, rank)
+
+            # Copy small gradients into per-GPU buffers and then async reduce
+            i_p = OSS._bucket(buffer, rank_params, copy=(global_rank == self_rank), gradients=True)
```
msbaines, line 169: Please rename i_p. Reading this code for the first time, its not clear to me what i_p is from the context.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -126,85 +131,69 @@ def reduce(self) -> None:
...
+                requests.append(dist.reduce(tensor=buffer, dst=global_rank, group=group, async_op=True))  # type: ignore
+
+            # Directly reduce the next grads
+            for param in rank_params[i_p:]:
+                # NOTE: workaround Gloo / leaf variable requiring grad modified in place
+                if param.requires_grad:
+                    restore_require_grad.append(param)
+                    param.requires_grad = False
+
+                param.div_(world_size)  # type: ignore
```
msbaines, line 182: Why are dividing the param?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -126,85 +131,69 @@ def reduce(self) -> None:
...
+                requests.append(dist.reduce(tensor=param, dst=global_rank, group=group, async_op=True))  # type: ignore
```
msbaines, line 183: Why are we reducing the params? Shouldn't we be reducing the grads?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -126,85 +131,69 @@ def reduce(self) -> None:
...
+            # Copy small gradients into per-GPU buffers and then async reduce
+            i_p = OSS._bucket(buffer, rank_params, copy=(global_rank == self_rank), gradients=True)
+
+            if i_p > 0:
+                buffer.div_(world_size)  # type: ignore
+                requests.append(dist.reduce(tensor=buffer, dst=global_rank, group=group, async_op=True))  # type: ignore
+
+            # Directly reduce the next grads
+            for param in rank_params[i_p:]:
+                # NOTE: workaround Gloo / leaf variable requiring grad modified in place
```
msbaines, line 177: This comment is less useful than the original comment. Please restore the original comment.

Can we avoid this hack with `with torch.no_grad():`?

Do we have a test for this? Can this only be done for gloo.

fairscale/optim/oss.py
```diff
@@ -49,45 +49,58 @@ class OSS(Optimizer):
...
+            the size of the buffer used to batch the small parameter tensors (default 1M).
     """
 
     #: The optimizer used for a given shard
     optim: Optimizer
 
     in_super_constructor: bool
 
-    def __init__(self, params: _params_t, optim: Type[Optimizer] = SGD, group: Optional[Any] = None, **default: Any):
+    def __init__(
```
msbaines, line 61: Please use 120 character length.

fairscale/optim/oss.py
```diff
@@ -395,3 +424,64 @@ def _free_other_grads(self) -> None:
             for p in partition:
                 for t in p["params"]:
                     t.grad = None
+
+    @staticmethod
+    def _scatter(buffer: torch.Tensor, params: List[Parameter], gradients: bool = False) -> None:
+        """Put as many parameters as possible from the parameter list in a given buffer.
```
msbaines, line 430: This comment seems wrong. It seems to be describing gather.

fairscale/optim/oss.py
```diff
@@ -395,3 +424,64 @@ def _free_other_grads(self) -> None:
             for p in partition:
                 for t in p["params"]:
                     t.grad = None
+
+    @staticmethod
+    def _scatter(buffer: torch.Tensor, params: List[Parameter], gradients: bool = False) -> None:
```
msbaines, line 429: I feel the gradients argument adds complexity. Why not just pass in a list of gradients instead?

fairscale/optim/oss.py
```diff
@@ -395,3 +424,64 @@ def _free_other_grads(self) -> None:
             for p in partition:
                 for t in p["params"]:
                     t.grad = None
+
+    @staticmethod
```
msbaines, line 428: This abstraction of _scatter doesn't seem necessary. I think the code would be simpler without it.

fairscale/optim/oss.py
```diff
@@ -175,26 +185,44 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+    def get_global_rank(group: Any, rank: int) -> int:
+        if group is dist.group.WORLD:
+            return rank
+        else:
+            global_rank = dist.distributed_c10d._get_global_rank(group, rank)  # type: ignore
+        return global_rank
+
+    @staticmethod
+    def _bucket(buffer: torch.Tensor, params: List[Parameter], copy: bool, gradients: bool = False) -> int:
```
msbaines, line 206: This abstraction also seems like it adds complexity. Open-coded, the bool arguments become unecessary.

msbaines (changes requested):
It feels like this should have been two changes.

Should the benchmark results be updated?

fairscale/nn/data_parallel/sharded_ddp.py
msbaines: Can we add a test to that catches this.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: wow, good catch, I'm not sure why it was not visible in the test

fairscale/optim/oss.py
blefaudeux: ah this is interesting, I was wondering about that when writing this down. Ok, changing this

fairscale/optim/oss.py
blefaudeux: same as above, the motivation was to remove code duplication

fairscale/optim/oss.py
blefaudeux: it removed duplicated code in between oss and shardedddp, that was the only point basically. Without that the broadcast and the reduce code chunks are 50 lines each doing almost the same thing

fairscale/optim/oss.py
blefaudeux: I'm using everything the contribute guide is telling me to do (including editorconfig, pre-commit hook), can we fix that in the tooling ?

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: no for 'with torch.no_grad():', the assert still triggers. Yes could be done for gloo only, but it adds a branching in the code ?

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: wrong edit, I was testing OSS only basically I overlooked that, thanks for catching this

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -51,7 +51,7 @@ def __init__(
         world_size: int,
         broadcast_buffers: bool,
         process_group: Any = None,
-        buffer_size: int = 2 ** 28,
+        buffer_size: int = 2 ** 20,
```
msbaines, line 54: You may want to consider having two parameters. A max size for the buffer and a min size below which a parameter is bucketed. Those feel like two separate concerns. Otherwise, you end up bucketing parameters that are very close to buffer_size. That would be a waste.

blefaudeux: it's not the same mechanism as before, the params are sorted by size and only the smallest ones are bucketed, and this happens only once so that the whole thing can stays async. This results in a huge perf win for shardedDDP

blefaudeux: the worst case is if the smallest parameter is just below the buffer size, this means *one* extra copy for no gain. All other cases that I can think of are what looks like reasonable tradeoffs. The benchmark is with a ResNet101, hopefully close to something typical for vision

fairscale/optim/oss.py
blefaudeux: removed

fairscale/optim/oss.py
blefaudeux: I think it was folded by black because of the extra argument, I have the exact black version from ci, not much more that I can do

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: if you meant: "do we have a test which would break if this is not in place", the answer is yes

blefaudeux:
> It feels like this should have been two changes.
The initial point was to remove duplication in between oss and shardedDDP, so it's hard to split this

> Should the benchmark results be updated?
OSS SDP was not regression tested, because super slow as a baseline (6fps vs 20fps for vanilla and oss). After this change I think that we could change the benchmarks indeed, and take the opportunity to use a configuration with a better throughput, but I would prefer to make this another diff

blefaudeux:
Optimizer step() time vs  Bucket size

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -76,6 +72,17 @@ def __init__(
...
 
+        # Allocate reduce buffers
+        # - Never use a bigger buffer than the number of model params
+        buffer_size = min(buffer_size, sum(p.numel() for p in self.module.parameters()))
+        self._reduce_buffers: Dict[torch.device, List[torch.Tensor]] = {}
+
+        # - One buffer per rank per device
+        for device, per_device in self.sharded_optimizer.per_device_params.items():
+            self._reduce_buffers[device] = [
+                torch.zeros(buffer_size, dtype=per_device[0][0].dtype, device=device) for _ in range(len(per_device))
```
msbaines, line 83: maybe use a variable for per_device[0][0].dtype?

fairscale/optim/oss.py
```diff
@@ -382,16 +387,95 @@ def _broadcast_state_dict(self) -> None:
...
+                i_bucketed += 1
+
+            if i_bucketed > 0:
+                future = dist.broadcast(tensor=buffer, src=global_rank, group=group, async_op=True)
+                if rank != self_rank:
+                    # This request will need to be unrolled
+                    bucket_requests.append((future, rank))
+
+            # Directly broadcast the rest
+            for param in params[i_bucketed:]:
```
msbaines, line 451: Are the params size ordered? How do we know that there aren't more small params.

fairscale/optim/oss.py
```diff
@@ -382,16 +387,95 @@ def _broadcast_state_dict(self) -> None:
...
+                # We save and restore the grad requirement state to work around that, in our case
+                # the grad is only useful on the source rank.
+                if param.requires_grad:
+                    restore_require_grad.append(param)
+                    param.requires_grad = False
+
+                requests.append(dist.broadcast(tensor=param, src=global_rank, group=group, async_op=True))
+
+        # Unroll the initial packed small parameters
+        for gate, rank in bucket_requests:
```
msbaines, line 463: earlier we are calling this a future and here we call it a gate. might be simpler to use consistent names.

fairscale/optim/oss.py
```diff
@@ -382,16 +387,95 @@ def _broadcast_state_dict(self) -> None:
...
+            if len(params) == 0:
+                continue
+
+            global_rank = OSS.get_global_rank(group, rank)
+
+            # Copy small parameters into per-GPU buffers
+            i_bucketed = 0  # the number of tensors packed in the buffer
+            offset = 0
+
+            while i_bucketed < len(params) and offset + params[i_bucketed].numel() < buffer_size:
```
msbaines, line 437: Are params size ordered?

fairscale/optim/oss.py
```diff
@@ -382,16 +387,95 @@ def _broadcast_state_dict(self) -> None:
...
+    @staticmethod
+    def get_global_rank(group: Any, rank: int) -> int:
+        if group is dist.group.WORLD:
+            return rank
+        else:
+            global_rank = dist.distributed_c10d._get_global_rank(group, rank)  # type: ignore
+        return global_rank
+
+    @staticmethod
+    def _consume_async_work(queue: List[Any]) -> None:
```
msbaines, line 412: Nit: might be more readable to just inline this.

msbaines (changes requested):
In an earlier iteration, there was a bug where we were reducing params instead of grads. This bug was not caught by our tests. Can we add a test to verify that DDP is working correctly.

fairscale/optim/oss.py
blefaudeux: per PR "The per-device parameters returned by OSS are now also sorted per rank, and then sorted per size, which simplifies the bucketing."

fairscale/optim/oss.py
blefaudeux: ordered ahead of time, it was explained in the PR

blefaudeux:
> In an earlier iteration, there was a bug where we were reducing params instead of grads. This bug was not caught by our tests. Can we add a test to verify that DDP is working correctly.

yes good point, we don't have a lot of tests on DDP to be honest, kind of the MVP. The issue would have been caught by the regression testing (which checks that we have the exact same loss in between pytorch <> oss <> oss_sdp on a toy problem), but better to have something more specific, I agree. Doing that

fairscale/optim/oss.py
blefaudeux: yeah, in the beginning the spirit was "let's remove all this duplicated code", now that everything is unrolled I agree that I can at least make it consistent. Taste and colors, I personally prefer tight functions in general so that the structure is more apparent

fairscale/optim/oss.py
blefaudeux: when unrolling the requests this acts as a gate (these "futures" are not really futures anyway, we cannot act on them), so I thought that was clearer to the reader. Apparently not, changing that

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -52,7 +59,8 @@ def run_one_step(rank, world_size, backend, device, temp_file_name):
     optimizer = ddp.optimizer
     model = ddp.module
 
-    input_tensor = torch.rand((64, 2)).to(device)
+    # Different input per rank, allows for checking that the gradients have been properly reduced
+    input_tensor = (torch.ones((64, 2)) * rank).to(device)
```
blefaudeux, line 63: this brings the inputs on one rank to 0, which kills one of the gradients (unless it has been properly synced), which makes it a better unit test.

blefaudeux:
> > In an earlier iteration, there was a bug where we were reducing params instead of grads. This bug was not caught by our tests. Can we add a test to verify that DDP is working correctly.
> 
> yes good point, we don't have a lot of tests on DDP to be honest, kind of the MVP. The issue would have been caught by the regression testing (which checks that we have the exact same loss in between pytorch <> oss <> oss_sdp on a toy problem), but better to have something more specific, I agree. Doing that

fixed

tests/nn/data_parallel/test_sharded_ddp.py
msbaines: I'm not sure this adjustment would have caught the earlier bug. Could we check the reduced gradient values are what we expect?

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -64,12 +72,16 @@ def run_one_step(rank, world_size, backend, device, temp_file_name):
     for pg in optimizer.optim.param_groups:
```
msbaines, line 72: The cuda.synchronize() is not necessary and should be removed.

fairscale/optim/oss.py
msbaines: Please add a comment to that affect or add 'ordered' to the variable name. It's nice if you can reason about the code without having to confirm that the params are ordered. Assumptions should be stated to avoid the code breaking during a future refactoring.

tests/nn/data_parallel/test_sharded_ddp.py
msbaines: What is the purpose of the `.abs().sum() / input_tensor.numel()` below ?

msbaines (changes requested):
Just some minor concerns about the test case.

tests/nn/data_parallel/test_sharded_ddp.py
blefaudeux: this means that the gradient on the weights is zero on the first layer on one of the ranks, so that if the grads are not reduced it is caught. Commenting out the ddp.reduce() or changing the ddp code to work on params instead of gradients does break this test. Now if you want another test, anything is possible of course

msbaines (approved):
LGTM. Thanks!

fairscale/optim/oss.py
```diff
@@ -382,16 +387,93 @@ def _broadcast_state_dict(self) -> None:
...
+            global_rank = OSS.get_global_rank(group, rank)
+
+            # Copy small parameters into per-GPU buffers
+            i_bucketed = 0  # the number of tensors packed in the buffer
+            offset = 0
+
+            # Since all the parameters are already sorted per increasing size, we only need to consider the first ones.
+            while i_bucketed < len(params) and offset + params[i_bucketed].numel() < buffer_size:
+                end = offset + params[i_bucketed].numel()
+                if rank == self_rank:
```
joshim5, line 437: @blefaudeux  Naman and I were trying to figure out why we were getting different results on Fairscale when using model parallelism. It looks like this diff broke the behavior I had fixed. 

Proprer behavior is restored if you change this line to `if rank == global_rank:`

cc @ngoyal2707 who is opening a fix

blefaudeux: oh, sorry about that :( could we add a proper unit test or regression test so that I don't break things again ?

min-xu-ai: The global_rank code is subtle, a unit test that prevent regression is non-trivial but probably very much worth it.

joshim5: @min-xu-ai I don't have a unit test, but I do have a regression/integration test. It's described in the PR description here: https://github.com/facebookresearch/fairscale/pull/121

Basically, run fairseq w/ and w/o OSS and check if the training logs are the same.

min-xu-ai: Don't worry. I was thinking a unit test within this repo since we change the code here from time to time and we don't want to wait until your test run to find out regression.

joshim5: It could still be added here! Run the script and compare the output. Sounds like Benjamin has a unit test though

PR #127, opened 2020-10-07
fix for megatron + oss

ngoyal2707:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -434,14 +434,14 @@ class OSS(Optimizer):
             # Since all the parameters are already sorted per increasing size, we only need to consider the first ones.
             while i_bucketed < len(params) and offset + params[i_bucketed].numel() < buffer_size:
                 end = offset + params[i_bucketed].numel()
-                if rank == self_rank:
+                if global_rank == self_rank:
                     buffer[offset:end].copy_(params[i_bucketed].data.view(-1))  # type: ignore
                 offset = end
                 i_bucketed += 1
 
             if i_bucketed > 0:
                 future = dist.broadcast(tensor=buffer, src=global_rank, group=group, async_op=True)
-                if rank != self_rank:
+                if global_rank != self_rank:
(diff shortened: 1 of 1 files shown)


blefaudeux (approved):
lgtm, sorry for the oversight

joshim5 (approved):
Naman and I discussed on chat, lgtm. Naman found that this fixes the inconsistencies w/ and w/o OSS. Tagging @blefaudeux since this fixes #122

fairscale/optim/oss.py
```diff
@@ -434,14 +434,14 @@ def _broadcast_params(
             # Since all the parameters are already sorted per increasing size, we only need to consider the first ones.
             while i_bucketed < len(params) and offset + params[i_bucketed].numel() < buffer_size:
                 end = offset + params[i_bucketed].numel()
-                if rank == self_rank:
+                if global_rank == self_rank:
```
msbaines, line 437: This function might be simpler if it didn't pass group and self_rank and instead just used self. That way, the correctness would be obvious. self_rank is ambiguous as to whether it is global or local.

```suggestion
                if global_rank == self.global_rank:
```

msbaines (approved):
Thanks for fixing. Some minor nits.

It would be nice to add a test so that this doesn't break in the future.

blefaudeux:
(context) we discussed the regression testing with Joshua already, I've a task up for that T77211997

blefaudeux:
Adding a test today, @joshim5 @ngoyal2707 would you prefer to merge as-is or pull in the suggestion from @msbaines ? I can also bring that in if you prefer

ngoyal2707:
> Adding a test today, @joshim5 @ngoyal2707 would you prefer to merge as-is or pull in the suggestion from @msbaines ? I can also bring that in if you prefer

lets merge as-is? and we can follow up with tests and other changes? cause current implementation breaks silently.

PR #129, opened 2020-10-08
[fix] OSS unit test to check data group

blefaudeux:
# Before submitting

## What does this PR do?
Catches the error in #127 in a more systematic basis

## PR review

## Did you have fun?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,12 @@ run_oss_benchmark: &run_oss_benchmark
       name: Run OSS Benchmark
       command: |
         python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 21.2 --reference_memory 4220 --reference_loss 0.63
-        python benchmarks/oss.py --gloo --optim_type oss
+
+run_oss_gloo: &run_oss_gloo
+- run:
+    name: Run OSS Benchmark
+    command: |
+      python benchmarks/oss.py --gloo --optim_type oss
 
 
 # -------------------------------------------------------------------------------------
@@ -253,6 +258,9 @@ jobs:
       - <<: *run_pipe_benchmark
 
       - <<: *run_oss_benchmark
+
+      - <<: *run_oss_gloo
+
 
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ run_oss_benchmark: &run_oss_benchmark
 
 run_oss_gloo: &run_oss_gloo
 - run:
-    name: Run OSS Benchmark
+    name: Run OSS with Gloo
     command: |
       python benchmarks/oss.py --gloo --optim_type oss
 
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -438,7 +438,7 @@ class OSS(Optimizer):
 
             if i_bucketed > 0:
                 future = dist.broadcast(tensor=buffer, src=global_rank, group=self.group, async_op=True)
-                if global_rank != self.rank:
+                if global_rank != self.global_rank:
                     # This request will need to be unrolled
                     bucket_requests.append((future, rank))
 
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -431,7 +431,7 @@ class OSS(Optimizer):
             # Since all the parameters are already sorted per increasing size, we only need to consider the first ones.
             while i_bucketed < len(params) and offset + params[i_bucketed].numel() < buffer_size:
                 end = offset + params[i_bucketed].numel()
-                if global_rank == self.rank:
+                if global_rank == self.global_rank:
                     buffer[offset:end].copy_(params[i_bucketed].data.view(-1))  # type: ignore
                 offset = end
                 i_bucketed += 1
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -194,7 +194,7 @@ class OSS(Optimizer):
                 device,
                 device_params,
             ) in self.per_device_params.items():  # all the params on this device (inc all ranks)
-                self._broadcast_params(self._broadcast_buffers[device], device_params, self.group, self.global_rank)
+                self._broadcast_params(self._broadcast_buffers[device], device_params)
 
         return loss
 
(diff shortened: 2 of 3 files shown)


blefaudeux:
you can check the commit just before the last one, this test does catch a faulty rank comparison

tests/optim/test_oss.py
```diff
@@ -334,3 +334,62 @@ def test_collect_shards():
...
+        model = torch.nn.Sequential(torch.nn.Linear(input_width, hidden), torch.nn.Linear(hidden, target_width)).to(
+            device
+        )
+
+        # With SGD, Momentum is required to get a state to shard
+        optimizer = optim.OSS(model.parameters(), lr=0.1, momentum=0.99, group=pg, broadcast_buffer_size=2 ** 20)
+        check(optimizer)
+
+    # Model not-fitting in the broadcast bucket
+    if rank in sub_group_ranks:
```
msbaines, line 380: This if seems redundant. Why not merge into the if above?

tests/optim/test_oss.py
```diff
@@ -334,3 +334,62 @@ def test_collect_shards():
...
+                loss.backward()
+
+                return loss
+
+            original_loss = optimizer.step(closure=closure)
+            post_update_loss = loss_fn(model(inputs), target)
+
+            assert (
+                original_loss - post_update_loss
+            ).norm() > 1e-6, f"Model update failed - difference {(original_loss - post_update_loss).norm()}"
```
msbaines, line 367: Why take the norm? Why not just abs?

If you check against an exact value the test will catch even more errors.

I'm not sure this is a strong enough test of correctness.

If OSS is working correctly then all workers should have the same params after step. Why not verify that all workers compute the same norm for the params?

tests/optim/test_oss.py
```diff
@@ -334,3 +334,62 @@ def test_collect_shards():
...
+    epochs, batch, input_width, hidden, target_width = 5, 3, 20, 10, 5
+    loss_fn = torch.nn.L1Loss().to(device)
+
+    def check(optimizer):
+        # Just run a couple of random epochs, check that the model is properly updated
+        for _ in range(epochs):
+            target = torch.rand((batch, target_width), device=device)
+            inputs = torch.rand((batch, input_width), device=device)
+
+            def closure():
```
msbaines, line 354: The test might be more realistic if you used DDP to reduce gradients.

.circleci/config.yml
```diff
@@ -101,7 +101,12 @@ run_oss_benchmark: &run_oss_benchmark
       name: Run OSS Benchmark
       command: |
         python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 21.2 --reference_memory 4220 --reference_loss 0.63
-        python benchmarks/oss.py --gloo --optim_type oss
+
+run_oss_gloo: &run_oss_gloo
```
msbaines, line 105: What is the rationale for breaking this out?

msbaines (changes requested):

tests/optim/test_oss.py
blefaudeux: I don't get that, could you elaborate ?

tests/optim/test_oss.py
blefaudeux: that was an afterthought, can be cleaned up indeed

tests/optim/test_oss.py
blefaudeux: > Why take the norm? Why not just abs?
I'm not sure why this is important ? it just meant "if they are too different", any norm would do

> If you check against an exact value the test will catch even more errors.
> 
> I'm not sure this is a strong enough test of correctness.
it does catch the error that the linked PR fixed, which was the initial point, so at least it's not useless. Now it's probably possible to do better

> If OSS is working correctly then all workers should have the same params after step. Why not verify that all workers compute the same norm for the params?
sounds like a good idea, I agree on that one

tests/optim/test_oss.py
blefaudeux: one issue with DDP is that I think that it requires CUDA, so it's more complicated to test with 3+ gpus so that there's a big enough sub-group where the optimizer works. Another point is that this is just meant to test the fact that multiple groups are properly handled by OSS, it's not a performance test, so I don't understand what DDP brings.

.circleci/config.yml
blefaudeux: it's not really a benchmark, just a correctness test because gloo will assert on some operations where NCCL will not. The setup (world size) is not the same for speed's sake, so the numbers in the log were not comparable.

This way all the values in oss benchmark are actual comparisons, I was planning to add [this alternative implementation](https://github.com/facebookresearch/fairscale/tree/oss_sharded_experimental) to the mix at some point and the logs looked[ like this](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/565/workflows/a87b1c8f-d4c1-4cd0-87f3-a704fe9e8d21/jobs/2128/parallel-runs/0/steps/0-110), hard for external people to see what's going on

fairscale/optim/oss.py
```diff
@@ -434,14 +431,14 @@ def _broadcast_params(
             # Since all the parameters are already sorted per increasing size, we only need to consider the first ones.
             while i_bucketed < len(params) and offset + params[i_bucketed].numel() < buffer_size:
                 end = offset + params[i_bucketed].numel()
-                if global_rank == self_rank:
+                if global_rank == self.global_rank:
```
blefaudeux, line 434: this was just changed on the previous commit to show that the test caught this mistake, not changed overall

tests/optim/test_oss.py
msbaines: I think it should be possible to all_reduce CPU tensors and should only add 1  or 2 more lines of code.

Tests also serve as examples of how to use the API so its nice if they match how the API is used in practice (when possible). If it only requires 1 or 2 more LOC it might be worth it.

msbaines (approved):

tests/optim/test_oss.py
blefaudeux: just to be sure, you mean all_reducing on the loss so that it looks like a normal training loop, or reducing all the gradients from here ?

fairscale/optim/oss.py
joshim5: Doesn't it make sense to have two tests? One that checks the correct usage and one that checks incorrect usage?

tests/optim/test_oss.py
joshim5: Agree with @msbaines. The challenge was getting this to work for DDP+MP, so the test should check for these combined.
@msbaines can also comment but I think all reducing on the loss is most realistic

blefaudeux: The committed version reduces on the loss only (and does catch your problem), we sync'ed 1:1 with Mandeep but that was too late, I initially missed it. I think that I understand that DDP would be nice here also, it needs to be there in the benchmark and tutorials in any case, but in this test it restricts the unit testing to platforms with 3+ GPUs, which are not that common. 

I can write a follow up unit test with DDP which only triggers if the HW fits, so that at least the CI catches it

PR #130, opened 2020-10-08
[minor] OSS: bring DDP in the benchmark

blefaudeux:
# Before submitting

## What does this PR do?
A suggestion from @msbaines on another PR made me think of that, the benchmark job that we had was not using DDP so not super realistic. The fake pictures were also the same on each rank, which prevents catching some sync issues, fixing that

## PR review

## Did you have fun?

### DIFF
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -88,7 +88,7 @@ def train(
             optimizer=OPTIM,
             optimizer_params={"lr": 1e-4, "momentum": 0.9},
             world_size=world_size,
-            broadcast_buffers=False,
+            broadcast_buffers=True,
         )
         ddp.train()
         optimizer = ddp.optimizer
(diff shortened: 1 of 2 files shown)


blefaudeux:
(the values in the benchmark regression test will need to be updated, that's expected. The current commit will fail because of that)

benchmarks/oss.py
```diff
@@ -91,6 +94,7 @@ def train(
         optimizer = ddp.optimizer
         model = ddp
     else:
+        model = DDP(model, device_ids=[rank])
```
min-xu-ai, line 97: should we toggle find_unused_parameters to True/False and see its impacts?

min-xu-ai (approved):
LGTM!

benchmarks/oss.py
blefaudeux: ahh, good point ! it was not yet up for review actually, looks like there's something wrong with SDP, looking into that :) I'll test this option

min-xu-ai: oops, sorry, didn't notice that! :-)

blefaudeux:
looks like something is wrong in SDP depending on the world size, and this benchmark change exposed that

PR #151, opened 2020-10-19
[test] moe: add a more thorough MOELayer routing test

msbaines:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/tests/nn/moe/test_moe_layer.py
+++ b/tests/nn/moe/test_moe_layer.py
@@ -73,6 +73,56 @@ def test_forward(device):
     assert torch.allclose(input, output)
 
 
+# Test Gate which round-robin routes tokens to experts
+class RoundRobinGate(torch.nn.Module):
+    def __init__(self, model_dim, num_experts):
+        super().__init__()
+        self.model_dim = model_dim
+        self.num_experts = num_experts
+
+    def forward(self, input):
+        g, s, m = input.shape
+        capacity = 2 * s // self.num_experts
+        output = torch.zeros(g, s, self.num_experts, capacity, dtype=input.dtype, device=input.device)
+        expert = 0
+        c = 0
+        for i in range(s):
+            for j in range(g):
+                output[j][i][expert][c] = 1.0
+            expert += 1
+            if expert == self.num_experts:
+                c += 1
+                expert = 0
+
+        return 0.0, output, output.bool()
+
+
+@pytest.mark.mpi
+@pytest.mark.parametrize("device", ["cpu"])
+def test_forward_routing(device):
+    model_dim = 8
+    num_experts = dist.get_world_size()
+    input = torch.randn(3, 4, 16, model_dim).to(device)
+    gate = RoundRobinGate(model_dim, num_experts)
+    expert = torch.nn.Linear(model_dim, model_dim, bias=False)
+    # Use scaling matrix (each rank has a different scale)
+    scale = dist.get_rank() + 1
+    expert.weight = torch.nn.Parameter(torch.eye(model_dim) * scale)
+    moe = MOELayer(gate, expert).to(device)
+    output = moe(input)
+    assert output.shape == input.shape
+    # Verify that each token was sent to the correct expert by checking its scale.
+    g, s, t, m = input.shape
+    expert = 0
+    for i in range(s):
+        for j in range(t):
+            for k in range(g):
+                torch.allclose(input[k][i][j], output[k][i][j] * (expert + 1))
+            expert += 1
+            if expert == num_experts:
+                expert = 0
+
+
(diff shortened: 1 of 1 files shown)


tests/nn/moe/test_moe_layer.py
```diff
@@ -73,6 +73,56 @@ def test_forward(device):
...
+# Test Gate which round-robin routes tokens to experts
+class RoundRobinGate(torch.nn.Module):
+    def __init__(self, model_dim, num_experts):
+        super().__init__()
+        self.model_dim = model_dim
+        self.num_experts = num_experts
+
+    def forward(self, input):
+        g, s, m = input.shape
+        capacity = 2 * s // self.num_experts
```
min-xu-ai, line 85: it seems that "//" should be "/" because if it is not divided evenly, then you should get an error anyway. "//" will mask error?

tests/nn/moe/test_moe_layer.py
```diff
@@ -73,6 +73,56 @@ def test_forward(device):
...
+        output = torch.zeros(g, s, self.num_experts, capacity, dtype=input.dtype, device=input.device)
```
min-xu-ai, line 86: I wonder if you can do something like:

```
o = []
for I in range(self.num_experts):
  x = np.array((g, s, 1, capacity))
  x[:,:,:,self.num_experts*I:self.num_experts*(I+1)] = 1.0
  o.append(x)
o = np.stack(o)
output = torch.tensor(o)
```

I didn't write the exact np code above, but just rough idea. That would avoid explicitly looping over all indexes and the meaning is more readable too?

min-xu-ai (commented):
some comments.

tests/nn/moe/test_moe_layer.py
msbaines: Added an assertion.

msbaines (commented):
Fixed. PTAL.

tests/nn/moe/test_moe_layer.py
```diff
@@ -73,6 +73,56 @@ def test_forward(device):
...
+
+    def forward(self, input):
+        g, s, _ = input.shape
+        assert s % self.num_experts == 0
+        capacity = 2 * s // self.num_experts
+        output = torch.zeros(g, s, self.num_experts, capacity, dtype=input.dtype, device=input.device)
+        expert = 0
+        c = 0
+        for i in range(s):
+            output[:, i, expert, c] = 1.0
```
min-xu-ai, line 91: it seems `expert = i % self.num_experts` and `c = i // self.num_experts`? No need to have separate variables and ++ them?

tests/nn/moe/test_moe_layer.py
```diff
@@ -73,6 +73,56 @@ def test_forward(device):
...
+    gate = RoundRobinGate(model_dim, num_experts)
+    expert = torch.nn.Linear(model_dim, model_dim, bias=False)
+    # Use scaling matrix (each rank has a different scale)
+    scale = dist.get_rank() + 1
+    expert.weight = torch.nn.Parameter(torch.eye(model_dim) * scale)
+    moe = MOELayer(gate, expert).to(device)
+    output = moe(input)
+    assert output.shape == input.shape
+    # Verify that each token was sent to the correct expert by checking its scale.
+    g, s, t, m = input.shape
```
min-xu-ai, line 115: same here, m can be replaced with _.

tests/nn/moe/test_moe_layer.py
```diff
@@ -73,6 +73,56 @@ def test_forward(device):
...
+    expert = 0
+    for i in range(s):
+        for j in range(t):
+            for k in range(g):
+                torch.allclose(input[k][i][j], output[k][i][j] * (expert + 1))
```
min-xu-ai, line 120: expert = (i * s + j * t)//num_experts?

tests/nn/moe/test_moe_layer.py
```diff
@@ -73,6 +73,56 @@ def test_forward(device):
...
+    expert.weight = torch.nn.Parameter(torch.eye(model_dim) * scale)
+    moe = MOELayer(gate, expert).to(device)
+    output = moe(input)
+    assert output.shape == input.shape
+    # Verify that each token was sent to the correct expert by checking its scale.
+    g, s, t, m = input.shape
+    expert = 0
+    for i in range(s):
+        for j in range(t):
+            for k in range(g):
```
min-xu-ai, line 119: similarly perhaps this loop can be avoided by use ":" indexing?

min-xu-ai (approved):
LGTM. minor suggestions to make it consistent with your latest change.

PR #152, opened 2020-10-20
[feat][minor] OSS benchmark - pick the model via args

blefaudeux:
# Before submitting

## What does this PR do?
Helpful in debugging a bug that I have right now on another branch, and could be practical for some folks who want an easy way to benchmark OSS across typical vision workloads. It would be nice to add NLP workloads in there also, as a follow up

## PR review

## Did you have fun?

### DIFF
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -16,7 +16,7 @@ import torch.nn as nn
 from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.utils.data import DataLoader
 from torchvision.datasets import FakeData
-from torchvision.models import resnet18, resnet50, resnet101, resnet152
+from torchvision.models import *  # noqa: F401
 from torchvision.transforms import ToTensor
 
 from fairscale.nn.data_parallel import ShardedDataParallel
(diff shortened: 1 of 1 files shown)


benchmarks/oss.py
```diff
@@ -30,9 +30,12 @@ def dist_init(rank, world_size, backend):
     dist.init_process_group(backend=backend, init_method="tcp://localhost:29501", rank=rank, world_size=world_size)
 
 
-def get_problem(rank, data_size, batch_size, device):
-    # Standard RN101
-    model = resnet101(pretrained=False, progress=True).to(device)
+def get_problem(rank, data_size, batch_size, device, resnet: str):
+    # Select the desired model on the fly
+    print(f"Using {resnet} for benchmarking")
+    model = {"resnet18": resnet18, "resnet50": resnet50, "resnet101": resnet101, "resnet152": resnet152}[resnet](
```
min-xu-ai, line 36: should we use importlib to import models directly from `torchvision.models`, that would avoid some glue code?

min-xu-ai (commented):
one suggestion.

benchmarks/oss.py
blefaudeux: ahh I didn't know that, would you have a pointer ?

min-xu-ai: an example:

```
import importlib


m1 = importlib.import_module('example.submodule')
print(m1)

m2 = importlib.import_module('.submodule', package='example')
print(m2)

print(m1 is m2)

```

you can turn string into import modules I think.

blefaudeux: the models are not modules really, so it does not work OOB, but looking into the syntax

min-xu-ai (approved):
LGTM!

PR #156, opened 2020-10-21
Pipe rpc

froody:
# Before submitting

## What does this PR do?

Adds support for:
* Reused layers (e.g. for weight sharing)
* Lazily-constructed layers
* Single-process control via PipeRPCWrapper

Also lays the foundation for asynchronous pipeline work by introducing an event loop for each rank/worker to process either activations or gradients as they arrive.

## PR review

## Did you have fun?


benchmarks/pipe.py
```diff
@@ -11,16 +11,26 @@
 from torch.distributed import rpc
 import torch.multiprocessing as mp
 import torch.nn as nn
+from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.utils.data import DataLoader
 import torchtext
 from torchtext.data.utils import get_tokenizer
 
+# from deepspeed.pipe import PipelineModule
```
msbaines, line 19: Remove.

tests/nn/model_parallel/commons.py
```diff
@@ -35,6 +36,12 @@
 from fairscale.nn.model_parallel import initialize_model_parallel
 from fairscale.nn.model_parallel.random import model_parallel_cuda_manual_seed
 
+try:
+    import torch_ucc  # noqa: F401
```
msbaines, line 40: This import seems unused.

benchmarks/pipe.py
```diff
@@ -11,16 +11,26 @@
...
 from fairscale.nn.model_parallel import initialize_model_parallel
-from fairscale.nn.pipe import pipe
+from fairscale.nn.model_parallel.initialize import get_data_parallel_group, get_pipeline_parallel_group
+from fairscale.nn.pipe import LazyModule, pipe
 from fairscale.optim import GradScaler
+from fairscale.optim.oss import OSS
 from tests.nn.model_parallel.commons import dist_init, get_worker_map
 
+try:
+    import torch_ucc  # noqa: F401
```
msbaines, line 29: This import does not seem to be used.

benchmarks/pipe.py
```diff
@@ -276,7 +289,15 @@ def train(lm_dataloader, model, criterion, optimizer, vocab_size, args):
 
     num_params = reduce(operator.add, (reduce(operator.mul, x.size()) for x in model.parameters()))
     if model.group:
-        print(f"training model, #prams = {num_params}, group: {model.group.rank()}, sizes {model.group.size()}")
+        total = torch.Tensor([num_params]).cuda()
+        torch.distributed.all_reduce(total, group=model.group)
+        print(
```
msbaines, line 294: use logging instead of print

tests/nn/model_parallel/commons.py
```diff
@@ -99,10 +113,20 @@ def spawn_for_all_world_sizes(test_func, world_sizes=get_world_sizes(), args=[])
         mp.spawn(test_func, args=(world_size, *args), nprocs=world_size, join=True)
 
 
-def helper(rank, world_size, func, args):
+def helper(rank, world_size, func, args, error_queue):
```
msbaines, line 116: Maybe a better name and a comment. Ditto for fixer. The terse names make understanding the code challenging.

tests/nn/model_parallel/test_layers.py
```diff
@@ -307,6 +307,10 @@ def run_test_row_parallel_linear(rank, model_parallel_size):
         print(" >> passed the test :-)")
 
 
+def a_barrier():
```
msbaines, line 310: The code felt cleaner when it called torch.distributed.barrier() directly.

tests/nn/moe/test_moe_layer.py
```diff
@@ -23,12 +23,14 @@
...
 if "OMPI_COMM_WORLD_SIZE" in os.environ:
-    dist.init_process_group(backend=dist.Backend.MPI)
+    pass  # dist.init_process_group(backend=dist.Backend.MPI)
 
 
 def setup_module(module):
     if "OMPI_COMM_WORLD_SIZE" not in os.environ:
         dist.init_process_group(backend=BACKEND, rank=0, world_size=1)
+    else:
+        dist.init_process_group(backend=dist.Backend.MPI)
```
msbaines, line 33: This doesn't work. I get an error about MPI being initialized multiple times.

msbaines (commented):
Looks good. Mostly nits.

tests/nn/moe/test_moe_layer.py
froody: ah, fixed, tested locally and it passes

fairscale/nn/pipe/messages.py
```diff
@@ -0,0 +1,137 @@
...
+import numpy as np
+import torch
+from torch import Tensor
+
+from fairscale.nn.model_parallel import get_pipeline_parallel_group
+
+from .types import MESSAGE_GENERATION_START, InputDevice, PipeMessage, Tensors, TransportConfig
+
+# FIXME Why is 256 ok for training but not for tests?
+MESSAGE_TENSOR_SIZE = 1024  # 256
```
msbaines, line 15: Why does this need to be fixed-size?

fairscale/nn/pipe/messages.py
```diff
@@ -0,0 +1,137 @@
...
+
+MessageQueues: List[Queue] = [Queue() for _ in range(MESSAGE_GENERATION_START)]
+
+
+def pyobject_to_tensor(obj: Any) -> Tensor:
```
msbaines, line 20: We have similar code in fairscale/optim/utils.py. Maybe we should create a fairscale.utils package to avoid duplication.

fairscale/nn/pipe/messages.py
```diff
@@ -0,0 +1,137 @@
...
+    else:
+        return tuple(t.to(input_device) for t in tensors)
+
+
+def rpc_push_queue(message: PipeMessage) -> None:
+    globals()["MessageQueues"][message.queue_name].put(message)
+
+
+def send_message(config: TransportConfig, message: PipeMessage, sync: bool = False) -> None:
+    if config.use_rpc:
```
msbaines, line 57: Might be nice to avoid the multiple config.use_rpc checks and just have a separate RPCMessageQueue class.

fairscale/nn/pipe/messages.py
```diff
@@ -0,0 +1,137 @@
...
+        return tensors
+    else:
+        return tuple(t.to(input_device) for t in tensors)
+
+
+def rpc_push_queue(message: PipeMessage) -> None:
+    globals()["MessageQueues"][message.queue_name].put(message)
+
+
+def send_message(config: TransportConfig, message: PipeMessage, sync: bool = False) -> None:
```
msbaines, line 56: I don't see a test for this or any callers in this PR. Is this dead code or is their a follow-up PR that uses this?

msbaines (commented):
If possible it would be nice if this could be broken up. Its hard to review all at once. It would also be useful to have a description of the design in one of the files as a comment or docstring.

benchmarks/pipe.py
```diff
@@ -460,10 +530,13 @@ def bench_single_process(args):
     blob = make_model_and_data(args, None, new_data=new_data)
     model = blob["model"]
 
-    balance = generate_balance(min(num_devices, 8), len(model))
-    p = pipe.Pipe(
-        model, balance, chunks=args.chunks, pipelined_backward=args.pipelined_backward, checkpoint=args.checkpoint
-    )
+    if args.deepspeed:
+        p = PipelineModule(layers=model, num_stages=min(num_devices, 4))
```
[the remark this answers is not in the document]
froody: hah you told me to delete the import, it never worked though so I'll remove it

fairscale/nn/pipe/messages.py
froody: Because send/recv have to agree on the size, so the two options are send the size first, or just pre-allocate a large enough buffer for all cases, so I chose the latter

fairscale/nn/pipe/messages.py
froody: oh I didn't know about that, created fairscale/utils/object.py

fairscale/nn/pipe/messages.py
froody: good point

fairscale/nn/pipe/messages.py
froody: It's called from pipeline.py and async_schedule.py

msbaines:
Added @blefaudeux to review fairscale.utils changs.

fairscale/optim/utils.py
```diff
@@ -52,11 +53,8 @@ def broadcast_object(
 
     if dist.get_rank() == src_rank:
         # Emit data
-        buffer = io.BytesIO()
-        torch.save(obj, buffer)
-        data = bytearray(buffer.getbuffer())
-        length_tensor = torch.LongTensor([len(data)]).to(dist_device)
-        data_send_tensor = torch.ByteTensor(data).to(dist_device)
+        data_send_tensor = pyobject_to_tensor(obj).to(dist_device)
```
blefaudeux, line 56: ah interesting, I didn't know of this way ! Could you elaborate on the benefits vs. BytesIO ? cc @mannatsingh, because this code comes from Classy initially. Else if that works better in any way that's fine by me, the code is a little cleaner I think

mannatsingh: Ah yeah this looks like a function within fairsclae which does something similar to what we do in classy, but through pickle and numpy (https://github.com/facebookresearch/fairscale/blob/63f7796a72f9afbcba64f9bf0df11753ecc4558d/fairscale/nn/pipe/pipeline.py#L108). Although I don't think we're supposed to pickle tensors. So if obj contains tensors this might become problematic?

froody: Ok, I'm going to revert my changes to fairscale.optim

mannatsingh: Just mentioned this to @blefaudeux over chat - I think the only issue with pickling tensors is the device handling bit. 
So I guess my statement isn't completely true - "I don't think we're supposed to pickle tensors". If the code works with both GPUs and CPUs, there's no need to get rid of that. Just something to be mindful of :)

msbaines (approved):

sidgoyal78 (commented):
Thanks for the PR. Looks great. The modules for async are already being used for ampnet/xpipe.

sidgoyal78 (approved):

PR #157, opened 2020-10-21
[feat] ShardedDataParallel with autoreduce

blefaudeux:
# Before submitting

## What does this PR do?
Stopgap solution before Pytorch's DDP become flexible enough to accommodate different reduction patterns out of the box, another DDP dedicated to the sharded optimizer which automatically reduces gradients to the appropriate ranks and releases the grad buffers 

Key features that this PR brings or maintains in a different form:
- [x] automatic gradient reduction to the appropriate ranks
- [x] overlap the gradient reduction and the backward pass
- [x] keep the tunable bucketing of small gradients
- [x] keep the reduce calls asynchronous (non-blocking) 

cc @mrshenli 

## PR review

## Did you have fun?


blefaudeux:
I need to make sure that all inputs are properly handled, incoming update (cc @SeanNaren). Open for feedback on anything though, that would be helpful

blefaudeux:
also cc @mannatsingh  and @ananthsub for visibility, if this flies through it would be great if we can integrate this DDP engine to Classy and Lightning, the memory savings and speedups are significant

blefaudeux:
I'll clean the PR up by removing all the support code in benchmark for instance later on, for now on please don't hesitate to comment on the sharded_ddp. I could not make the tensors match exactly DDP's after the grad reduce, they're close but not the same. If I all_reduce instead of reduce the results are again very slightly different, both from sharded_ddp/reduce and pytorch's DDP.

blefaudeux:
Now depends on #166, which removes some unrelated changes

blefaudeux:
Finally a little cleaner, sorry for all the commits and PRs, I was trying to offload unrelated changes. A remaining open question for me now is whether we keep the previous ShardedDDP iteration or not

blefaudeux:
ping @msbaines, would you mind having a look ? testing this today on fb infra

blefaudeux:
Back to the drawing board, testing that with Classy on FB infra shows that the inter node communication is too slow / scaling is very bad. The perf and memory savings are very good on a single node, so the principles are sound as expected, but should be improved

blefaudeux:
- Benchmarked on FB infra, similar if not better scaling behaviour as DDP
- Gradient accumulation is not handled, @msbaines @myleott shall I add that ?

blefaudeux:
Cleaning that up, @SeanNaren came up with a better implementation to make sure that the reduce is done before stepping (for now, in the future the async lane could be prolonged), backporting it here

blefaudeux:
> * Benchmarked on FB infra, similar if not better scaling behaviour as DDP
> * Gradient accumulation is not handled, @msbaines @myleott shall I add that ?

Gradient accumulation now handled.

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -42,7 +44,7 @@ def run_one_step(rank, world_size, backend, device, temp_file_name):
     torch.manual_seed(rank)
     np.random.seed(rank)
 
-    def check(broadcast_buffers: bool) -> None:
+    def check(broadcast_buffers: bool, grad_acccumulation: bool = False) -> None:
```
SeanNaren, line 47: I think this should be `grad_accumulation` rather than `grad_acccumulation`

blefaudeux:
ping reviewers, not sure if you would like something to be changed

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
+        self._grad_to_rank = [
+            _get_global_rank(self.process_group, self._find_rank(p)[1])
+            for p in filter(lambda x: x.requires_grad, self.base_model.parameters())
+        ]
+
+        # Setup the reduce buckets
+        # reduce_buffer dimensions: [optimizer, device, rank params]
+        # bucket_strategy dimensions: [optimizer, device, rank tuple -should bucket, offset_start, offset_end-]
+        # bucket_state dimensions: [optimizer, device, rank - current offset, max offset-]
+        self._reduce_buckets: Dict[OSS, Dict[torch.device, List[Dict[str, Any]]]] = {}
```
msbaines, line 118: Might be better for readability if these were classes.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -116,122 +185,180 @@ def no_sync(self) -> Generator:
...
-
-            if i_bucketed > 0:
-                buffer.div_(world_size)
-                bucket_requests.append(
-                    (
-                        dist.reduce(tensor=buffer, dst=global_rank, group=group, async_op=True),  # type: ignore
-                        rank,
-                    )
-                )
+    def _mark_inputs_requires_grad(*inputs: Any) -> Any:
```
msbaines, line 189: This will not work for non-float tensors.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
+            for device, per_rank_params in sharded_optimizer.per_device_params.items():
+                for r in range(self.world_size):
+                    self._bucket_state[sharded_optimizer][device][r] = (
+                        0,
+                        self._bucket_state[sharded_optimizer][device][r][1],
+                    )
 
-        # Flag used to make sure we only reduce gradients one time in the execution engine
-        self.need_reduction = False
+        # mark tensors as "requires_grad" to register the Gatekeeper in the autograd graph
```
msbaines, line 147: This will not work for non-float input tensors. The inputs in NLP are not floats.

Can you add a test for the non-float case.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
-        self.authoritative_rank = 0
+        self.global_rank = _get_global_rank(self.process_group, self.rank)
+
+        # Expose the same attributes as PytorchDDP, some frameworks rely on them.
+        # See https://pytorch.org/docs/stable/_modules/torch/nn/parallel/distributed.html#DistributedDataParallel
+        self.is_multi_device_module = len({p.device for p in self.base_model.parameters()}) > 1
+        distinct_device_types = {p.device.type for p in self.base_model.parameters()}
+        assert len(distinct_device_types) == 1, (
+            "ShardedDataParallel's input module must be on "
+            "the same type of devices, but input module parameters locate in {}."
```
min-xu-ai, line 104: locate in -> locate on?

fairscale/nn/data_parallel/sharded_ddp.py
min-xu-ai: add a docstring on why this function is needed?

min-xu-ai (commented):
This is very cool. Thanks for doing this heroic work! I am thinking of several things to do in the future. Do you mind I file them as issues?

1. if the buckets are currently the same size as the grads, perhaps we should introduce a mode where a fixed sized buckets are used to reduce memory consumption?
2. we should make sure this works without OSS?
3. after 1, to avoid python GIL overhead, perhaps we should move reducing into its own python process and use queues to send reduction data to that process for parallel reduction?
4. ensure this works with checkpointing

what do you think?

fairscale/optim/oss.py
```diff
@@ -181,20 +184,16 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+        self._free_other_grads()  # Depending on the DDP engine used, gradients specific to other ranks may still be loaded
+
         # Sync all the updated shards in between the ranks
-        with torch.no_grad():
-            for (
-                device,
-                device_params,
-            ) in self.per_device_params.items():  # all the params on this device (inc all ranks)
-                self._broadcast_params(self._broadcast_buffers[device], device_params)
+        for (device, device_params,) in self.per_device_params.items():  # all the params on this device (inc all ranks)
```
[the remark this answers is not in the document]
blefaudeux: Ah, I thought that it was useless actually, we're broadcasting which is not differentiable. I can put it back in, I don't know if there are perf implications in removing the autograd machinery

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
-        module: nn.Module,
-        optimizer: Type[torch.optim.Optimizer],
-        optimizer_params: Dict[str, Any],
-        world_size: int,
-        broadcast_buffers: bool,
+        base_model: nn.Module,
+        sharded_optimizer: Union[OSS, List[OSS]],
         process_group: Any = None,
-        buffer_size: int = 2 ** 19,
+        broadcast_buffers: bool = True,
```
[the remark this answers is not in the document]
blefaudeux: it's the same variable in DDP, I was trying to keep things aligned https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
+        reduce_buffer_size: int = 2 ** 19,
     ):
         super().__init__()
 
-        self.module = module
-        self.world_size = world_size
+        self.base_model = base_model
+        self.sharded_optimizers = [sharded_optimizer] if isinstance(sharded_optimizer, OSS) else sharded_optimizer
```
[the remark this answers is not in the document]
blefaudeux: lighting folks are testing that, but it's true that we could add a test case here, good point !

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
+            the sharded optimizer(s) which will decide the gradient partitioning
+    Keyword Args:
+        process_group (torch.nn.Optimizer):
+            optimizer to shard (default: SGD)
+        process_group (group):
+            torch.distributed group (default: group.WORLD)
+        broadcast_buffers (bool):
+            whether to broadcast model buffers in between ranks at the beginning of each forward pass
+        reduce_buffer_size (int):
+            the size of the per-device-per-rank buffers used for the reduce operation
```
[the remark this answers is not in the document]
blefaudeux: @min-xu-ai it's a buffer per device and per rank, that's what I was trying to explain. Basically one buffer per i<>j link. Ideally we should just reuse the buffer from OSS which is used for the same thing, but broadcast, maybe worth an update ? Yes for the default size, good point !

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
+from typing import Any, Callable, Dict, Generator, List, Tuple, Union
 
 import torch
-from torch import Tensor, nn
+from torch import nn
 import torch.distributed as dist
 from torch.nn import Parameter
 
-from fairscale.optim import OSS
+from fairscale.optim.oss import OSS
```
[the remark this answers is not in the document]
blefaudeux: ah no idea, it was an autocomplete. The diff compares to the old file but this was written from scratch since different logic, so I did not realize that the import differed. Fixing that

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: @msbaines @min-xu-ai essentially this needs 1 tensor to be marked, so that the Gatekeeper is registered in the autograd graph. I would really love to get rid of that, other solutions mentioned [here for instance](https://discuss.pytorch.org/t/custom-autograd-function-backward-pass-not-called/9144) seem deprecated

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: > This is very cool. Thanks for doing this heroic work! I am thinking of several things to do in the future. Do you mind I file them as issues?
> 
> 1. if the buckets are currently the same size as the grads, perhaps we should introduce a mode where a fixed sized buckets are used to reduce memory consumption?

the buckets are fixed size right now, it's a parameter actually. They are a lot smaller than the grads, it's only to collect the smallest ones, most of them are reduced directly. Two follow ups that I see are that we could use tensor views to remove the need for a copy, and/or reuse the buckets that OSS also has, they should be the same

> 2. we should make sure this works without OSS?

I don't see how that is possible, the whole point is that it reduces the gradients to the right ranks, you need a sharded optimizer for that ?

> 3. after 1, to avoid python GIL overhead, perhaps we should move reducing into its own python process and use queues to send reduction data to that process for parallel reduction?
ah yes, it would be nice to measure that, there's probably some perf to win there, I agree

> 4. ensure this works with checkpointing
yes, good point ! there's a context manager to accumulate/block the reduce, so I'm guessing that it should work already, but would be nice to test

blefaudeux:
> This is very cool. Thanks for doing this heroic work! I am thinking of several things to do in the future. Do you mind I file them as issues?
> 
> 1. if the buckets are currently the same size as the grads, perhaps we should introduce a mode where a fixed sized buckets are used to reduce memory consumption?

the buckets are fixed size right now, it's a parameter actually. They are a lot smaller than the grads, it's only to collect the smallest ones, most of them are reduced directly. Two follow ups that I see are that we could use tensor views to remove the need for a copy, and/or reuse the buckets that OSS also has, they should be the same

> 2. we should make sure this works without OSS?

I don't see how that is possible, the whole point is that it reduces the gradients to the right ranks, you need a sharded optimizer for that ?

> 3. after 1, to avoid python GIL overhead, perhaps we should move reducing into its own python process and use queues to send reduction data to that process for parallel reduction?

ah yes, it would be nice to measure that, there's probably some perf to win there, I agree

> 4. ensure this works with checkpointing

yes, good point ! there's a context manager to accumulate/block the reduce, so I'm guessing that it should work already, but would be nice to test

blefaudeux:
re: GIL @min-xu-ai , afterthought: in that case (autograd hooks) I think that the calls are done from the reducer, in C++, so there should be no GIL contention. Could still be worth checking that though

blefaudeux:
@reviewers thanks for the feedback, fixing the remarks and converting to draft until I can figure out a way to get rid of the requires_grad step

min-xu-ai:
> 
> 
> re: GIL @min-xu-ai , afterthought: in that case (autograd hooks) I think that the calls are done from the reducer, in C++, so there should be no GIL contention. Could still be worth checking that though

I was thinking the callbacks are in python, so they may have GIL contention esp. when the number of params is large, right?

blefaudeux:
> I was thinking the callbacks are in python, so they may have GIL contention esp. when the number of params is large, right?

I'm not super clear on that, they're called through the cpp engine so I'm not sure that the main GIL is involved. In any case it could be a follow up if need be ?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,178 @@
...
-        # Flag used to make sure we only reduce gradients one time in the execution engine
-        self.need_reduction = False
+        # mark tensors as "requires_grad" to register the Gatekeeper in the autograd graph
+        inputs = ShardedDataParallel._mark_inputs_requires_grad(*inputs)
 
-        # We can also forcibly accumulate grads locally and only do the
-        # gradients-reduce at some later time
-        self.accumulate_grads = False
+        # Register the gatekeeper, reduce wise
+        inputs = Gatekeeper.apply(self._reduce_work_handles, *inputs)
```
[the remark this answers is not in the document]
blefaudeux: it's a good point, I'll try to handle that directly here, simpler on the long run. I'm rewriting that part given feedback, also removing the requires_grad requirement

min-xu-ai:
> 
> 
> > I was thinking the callbacks are in python, so they may have GIL contention esp. when the number of params is large, right?
> 
> I'm not super clear on that, they're called through the cpp engine so I'm not sure that the main GIL is involved. In any case it could be a follow up if need be ?

Of course!

fairscale/nn/data_parallel/sharded_ddp_legacy.py
```diff
@@ -0,0 +1,237 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
```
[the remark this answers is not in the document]
min-xu-ai: if it is not used, I vote for removal.

fairscale/nn/data_parallel/sharded_ddp.py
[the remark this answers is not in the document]
min-xu-ai: Yeah, sounds good. You can also keep the parameter the same but rename the member variable to make it a bit more readable. Totally up to you.

fairscale/nn/data_parallel/sharded_ddp.py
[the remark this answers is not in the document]
blefaudeux: I just added a unit test, resolving the conversation, thanks Min for raising that

fairscale/nn/data_parallel/sharded_ddp.py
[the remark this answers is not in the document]
blefaudeux: doing that, good point

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: Fixed, new take which removes this faulty requirement

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,109 +4,126 @@
...
-        return self
+        _ = list(map(lambda x: x.wait(), work_handles))
+
+    def sync_buffers(self) -> None:
+        """
+        Sync all the param buffers in between ranks (including for instance batch norm statistics).
+        """
+        with torch.no_grad():
+            for buffer in self.base_model.buffers(recurse=True):
+                dist.broadcast(buffer.data, self.reference_global_rank, self.process_group, async_op=True)
```
mrshenli, line 126: any reason for not waiting for buffers?

blefaudeux: on the default stream, so the next operations will be queued anyway ? I can wait if that's safer, could be that it's a wrong assumption

blefaudeux: (fixed)

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,263 @@
...
+                                flat.param.grad = None
+                            else:
+                                # this rank is the owner, unroll the results
+                                assert flat.param.grad is not None
 
-        # Unroll the initial packed small gradients, as soon as possible
-        for future, rank in bucket_requests:
-            future.wait()
+                                flat.param.grad.data.copy_(
+                                    bucket.buffer[flat.start : flat.stop].view_as(flat.param.data)
```
mrshenli, line 223: maybe using `non_blocking=True` will help a bit here.

mrshenli (commented):
I didn't see anything that might lead to slowness when using larger buckets. The implementation of `reduce_bucket` makes sense to me. 

One thing to mention is that, if you are using multiple gpus per process, the GIL contention might cause some speed penalty, as the autograd engine has one thread per GPU. But this should be orthogonal to bucket sizes.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: ah, I checked that in the doc in the past and it was mentioning that it was useful for cpu<>gpu, but I can test this :) thanks !

blefaudeux:
ping @msbaines @min-xu-ai @myleott I've tried to address the previous review concerns

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+            # Reset all the grad reduce and bucket state flags
+            self._grad_to_be_reduced = [True] * len(self._grad_to_be_reduced)
+
+        def reduce_direct(*_: Any) -> None:
+            # Skip gradient reduction, do not alter status flags
+            if not self.accumulate_grads and self._grad_to_be_reduced[index]:
+                assert param.grad is not None
+
+                # Make sure that this is not fired twice
+                self._grad_to_be_reduced[index] = False
```
msbaines, line 174: What if its a shared parameter (e.g. shared embedding table)?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
-                        dist.reduce(tensor=buffer, dst=global_rank, group=group, async_op=True),  # type: ignore
-                        rank,
-                    )
+                # Make sure that this is not fired twice
+                self._grad_to_be_reduced[index] = False
+
+                # Copy to the flat buffer, update the buffer state
+                bucket = optimizer.buckets[param.device][dst_rank]
+
+                assert bucket.append(param, use_gradient=True), "%s - %s - %s" % (
```
msbaines, line 208: Better assert message.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: good point, I'm not super clear with that actually, looking into it. How does the BW proceed in that case, the grads are updated twice ?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+        broadcast_buffers: bool = True,
     ):
         super().__init__()
 
-        self.module = module
-        self.world_size = world_size
+        self.base_model = base_model
+        self.sharded_optimizers = [sharded_optimizer] if isinstance(sharded_optimizer, OSS) else sharded_optimizer
+        self.enable_broadcast_buffers = broadcast_buffers
+        self.accumulate_grads = False
```
joshim5, line 62: This is confusing out of context - maybe note that this will be temporarily changed in a context manager on `no_sync`?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+        self.world_size = dist.get_world_size(self.process_group)
+        self.reference_global_rank = OSS.get_global_rank(self.process_group, 0)  # picking rank 0 as the reference
         self.rank = dist.get_rank(self.process_group)
-        self.broadcast_buffers = broadcast_buffers
-        self.authoritative_rank = 0
+        self.global_rank = OSS.get_global_rank(self.process_group, self.rank)
+
+        # Expose the same attributes as PytorchDDP, some frameworks rely on them.
+        # See https://pytorch.org/docs/stable/_modules/torch/nn/parallel/distributed.html#DistributedDataParallel
+        self.devices = {p.device for p in self.base_model.parameters()}
```
joshim5, line 73:
```suggestion
        self.device = {p.device for p in self.base_model.parameters()}
```
Looks like this variable is called `self.device` in [the link you posted](https://pytorch.org/docs/stable/_modules/torch/nn/parallel/distributed.html#DistributedDataParallel).

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+        ).format(distinct_device_types)
+        self.device_type = list(distinct_device_types)[0]
+
+        # Scafolding to be able to reduce the grads during the BW pass
+        self._param_iterator = chain(*[optim.param_bucket_strategy.keys() for optim in self.sharded_optimizers])
+        self._grad_to_be_reduced = [True for _ in self._param_iterator]
+        self._grad_accs: List[Callable] = []
+        self._setup_backward_hooks()
+
+        # Make sure that all ranks start with the same model
```
joshim5, line 90: Why wouldn't they?

joshim5: Discussed over chat - this is just in beginning of training in case there are different seeds.
Small optimization could be to make this optional. For example, if we're loading from a saved checkpoint, we can assume that the parameters are already synced. For large models, this can save a minute of time and make startup faster.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+        Two possible backward hooks for a given parameter: either directly reduce to the appropriate rank,
+        or contribute to a bucket and reduce when the bucket is full.
+
+        Either way a delayed action is necessary and is passed as a callback.
         """
-        assert self.module.training, "Cannot call reduce in eval"
 
-        if not self.need_reduction or self.accumulate_grads:
-            return
+        def gatekeeper() -> None:
```
joshim5, line 163: Maybe add a comment explaining this method? I'm not familiar with this call, so I can't comment on correctness

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+            Variable._execution_engine.queue_callback(optimizer._consume_work_handles)
+
+            # Reset all the grad reduce and bucket state flags
+            self._grad_to_be_reduced = [True] * len(self._grad_to_be_reduced)
```
joshim5, line 167: Could add a comment explaining that this is a pseudo-future

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+
+        def reduce_direct(*_: Any) -> None:
```
joshim5, line 169: nice, this is very clear

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+                            callback=unwrap,
+                        )
+                    )
+
+                    # If all the reduce operations have been called, add the gatekeeper
+                    if len(optimizer.work_handles) == optimizer._max_work_handles:
+                        gatekeeper()
 
-    def _sync_buffers(self) -> None:
+        return reduce_bucket if should_bucket else reduce_direct
```
joshim5, line 247: Might be cleaner to define reduce_direct and reduce_bucket as instance methods and call the appropriate one depending on should_bucket?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+
+    def _setup_backward_hooks(self) -> None:
         """
-        Sync all the param buffers in between ranks.
-        TODO: Could be worth bucketing ?
+        Attach a reduce function to each grad-requiring parameter.
```
joshim5, line 251: What happens if I initialize a model and then later call `param.requires_grad=False` later after we've already wrapped the model? (I assume this usecase is legal)

joshim5 (commented):
This looks great. I left comments with a few suggestions to make things clearer, as well as a few questions about corner cases.
As we discussed on chat, it would be nice to support this without an optimizer. The workflow in Fairseq and ClassyVision is to wrap both `model` and `criterion` in DDP -- the second one doesn't have an optimizer. Additionally, in Fairseq, the optimizer abstraction is seperate from the model and doesn't exist until after the model is completely setup with DDP. Currently, Fairseq does:
```
model = YourFavoriteModel()
model = DDP(model)
optimizer = YourFavoriteOptimizer(model.parameters(), ...)
```

We will need to refactor this to:
```
model = YourFavoriteModel()
optimizer = YourFavoriteOptimizer(model.parameters(), ...)
model = DDP(model, optimizer)
```

I assume that other frameworks follow the first design, as it makes sense to do all model setup at once. This could be resolved if `model = ShardedDDP(model)` just adds properties to the model, which are then accessed by `OSS(model, model.parameters())`. This also makes it easy to switch OSS for another optimizer -- everything would still work; the sharding logic would just be ignored.

fairscale/nn/data_parallel/sharded_ddp.py
joshim5: If it's the same param, there is only one param object and one gradient, right? So should be fine. If you look at `model.parameters()` the shared param should only be there once - the parameters and grads are agnostic to the architecture itself.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+            model to be wrapped
+        sharded_optimizer (OSS, or list of OSS):
+            the sharded optimizer(s) which will decide the gradient partitioning
+
+    Keyword Args:
+        process_group (torch.nn.Optimizer):
+            optimizer to shard (default: SGD)
+        process_group (group):
+            torch.distributed group (default: group.WORLD)
+        broadcast_buffers (bool):
```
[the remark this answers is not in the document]
blefaudeux: ah yes sorry, I'll try to explain better. It's an extra step basically, same for DDP, "do you want your model buffers to be broadcasted at the beginning of each FW ?" would be a better phrasing

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
+        self.is_multi_device_module = len(self.devices) > 1
+        distinct_device_types = {p.device.type for p in self.base_model.parameters()}
+        assert len(distinct_device_types) == 1, (
+            "ShardedDataParallel's input module must be on "
+            "the same type of devices, but input module parameters are located on {} different device types."
+        ).format(distinct_device_types)
+        self.device_type = list(distinct_device_types)[0]
+
+        # Scafolding to be able to reduce the grads during the BW pass
+        self._param_iterator = chain(*[optim.param_bucket_strategy.keys() for optim in self.sharded_optimizers])
```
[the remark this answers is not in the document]
blefaudeux: tentatively done

fairscale/optim/oss.py
```diff
@@ -88,22 +88,26 @@ def __init__(
...
+        self.buckets: Dict[torch.device, List[Bucket]] = {}
         for device, per_device in self.per_device_params.items():
             # Allocate one buffer per rank and per device to group the small parameters
-            self._broadcast_buffers[device] = [
-                torch.zeros(broadcast_buffer_size, dtype=per_device[0][0].dtype, device=device)
+            self.buckets[device] = [
+                Bucket(buffer=torch.zeros(broadcast_buffer_size, dtype=per_device[0][0].dtype, device=device))
                 for _ in range(len(per_device))
             ]
+        self.param_bucket_strategy: Dict[torch.Tensor, bool] = {}
```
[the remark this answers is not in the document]
blefaudeux: done

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,266 @@
...
-        # Build the sharded optimizer
-        self.sharded_optimizer = OSS(self.module.parameters(), optim=optimizer, group=process_group, **optimizer_params)
+    def sync_all_params(self) -> None:
+        """
+        Sync the complete model states in between the ranks
+        """
+        with torch.no_grad():
+            work_handles = [
+                dist.broadcast(t, src=self.reference_global_rank, group=self.process_group, async_op=True)
+                for t in self.base_model.state_dict().values()
```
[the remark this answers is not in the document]
blefaudeux: the name is not appropriate, we're not just sync'ing the params actually, we're syncing the whole model (this is the exact same step as in DDP) to make sure that all ranks start with the same state. This is useful if for instance a seed is set per rank, then the model initialization would differ. I'm trying no to break existing workflows which have been relying on DDP basically, DDP does that so do I. Is that ok ?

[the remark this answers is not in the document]
blefaudeux: (see _sync_params_and_buffers(), in https://pytorch.org/docs/stable/_modules/torch/nn/parallel/distributed.html#DistributedDataParallel), I've even a unit test which checks that -and which rightly triggered in the previous commit-. I'll try to rephrase this part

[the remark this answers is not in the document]
blefaudeux: fixed

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -4,234 +4,268 @@
...
+        sharded_optimizer: Union[OSS, List[OSS]],
         process_group: Any = None,
-        buffer_size: int = 2 ** 19,
+        broadcast_buffers: bool = True,
     ):
         super().__init__()
 
-        self.module = module
-        self.world_size = world_size
+        self.base_model = module
```
SeanNaren, line 59: Would it be possible to change `self.base_model` to `self.module` as well? This would keep it consistent (this is also how we access base class across DDP etc in lightning).

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: it's a little bit more complicated than that, because they will be called as hooks, and in that case the only allowed parameter is 'grad' (so I cannot pass the destination rank for instance, [see](https://pytorch.org/docs/stable/autograd.html?highlight=register_hook#torch.Tensor.register_hook) ). That's why they are closures here and saved in the pseudo future (work handle list), because they capture these extra parameters, I don't think that it's doable as a method. One thing to remember is that I don't even know in which order they will be called, so there's no extra context that I use to guess these. It should be greatly simplified with proper futures, this is being worked on I think (cc @mrshenli to make sure I'm not completely wrong on that)

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: it's a good question, I had not thought of that. This case would not always work, and the reverse would never work:
- switch to requires_grad = False:
--  it would work if the param is not bucketed (it will not be part of the autograd and the hook will not be triggered)
-- it would not work if the param is bucketed, because it means that the bucket will never be full so never sent
- switch to requires_grad = True: will never work since this param will have no hook attached to it. I'm not sure how DDP handles that @mrshenli ?

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: done

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: done, now optional

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: I added a comment and renamed to show that it's a bool

fairscale/nn/data_parallel/sharded_ddp.py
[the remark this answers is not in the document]
blefaudeux: added extra comments

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: hmm, actually I think that's doable but some parts would be more complicated (find out where the hook is called from given the grad), especially since said grads are typically dropped if this rank does not own them (so I cannot have a static LUT)

msbaines (approved):
Very nice!

LGTM.

PR #159, opened 2020-10-21
[feature] OSS: Use MNIST to benchmark

### LINKED ISSUE #158: OSS: improve on the default benchmark, pass in something realistic
## 🚀 Feature
Instead of optimizing on random data, pick up something sensible from torchvision

## Motivation
Easier to compare to existing workloads to people, easier to make sure that we test on something realistic

## Pitch
Pick a small dataset from torchvision and use it

## Alternatives
- keep optimizing on random data
- create synthetic data on the fly

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #158, uses a real case (classify numbers) to benchmark DDP vs. OSS/DDP vs. OSS/ShardedDDP 

## PR review

## Did you have fun?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ run_oss_benchmark: &run_oss_benchmark
   - run:
       name: Run OSS Benchmark
       command: |
-        python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 13.7 --reference_memory 4390 --reference_loss 0.302
+        python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 800 --reference_memory 1120 --reference_loss 0.049
 
 run_oss_gloo: &run_oss_gloo
 - run:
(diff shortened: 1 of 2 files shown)


blefaudeux:
(I need to update the regression thresholds, this is expected)

min-xu-ai (approved):
LGTM.

We need to be careful using small model & data for benchmark though. I have found that even with cifar, its DDP performance is very different from real model & data, like imagenet or vissl.

blefaudeux:
> LGTM.
> 
> We need to be careful using small model & data for benchmark though. I have found that even with cifar, its DDP performance is very different from real model & data, like imagenet or vissl.

the default model is the same, resnet101, it's just that the previous optimization was on noise (FakeData) so I found it hard to relate to LR and loss for instance. I don't think that it makes a lot of sense overall though, I agree, but it's also something which needs to be reliably run on CI, so I thought that was a good compromise. I initially went for Fashion-MNIST by the way, but we discussed with @Vittorio-Caggiano and decided to stick to flight proven MNIST.

PR #167, opened 2020-10-23
[feat] OSS clip grads, bridge sharded states

### LINKED ISSUE #98: [OSS] Expose `clip_grad_norm` for the sharded state
## 🚀 Feature
Expose a method capable of returning the grad norm and clipping them which works on the sharded state

## Motivation
Similar to https://pytorch.org/docs/master/generated/torch.nn.utils.clip_grad_norm_.html but with OSS and sharded gradients

## Pitch
Could be done within ShardedDataParallel, reduce the norms and normalize the shards

## Alternatives
Not exposing this, could be problematic for some usecases

## Additional context
Exposed in a similar context by fairseq_optimizer

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #98, adds the option to clip the grads in a similar way to what torch already exposes. Required for OSS since each rank only has a partial view to all the gradients. 

## PR review

## Did you have fun?


fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+        # Filter out the grad-less params, concatenate params from all devices
+        local_params = [
+            filter(lambda x: x.grad is not None, device_params[self.rank])
+            for device_params in self.per_device_params.values()
+        ]
+
+        # Compute the norm on this grad set
+        if norm_type == inf:
+            local_norm = max(
+                p.grad.detach().abs().max().to(self._device)  # type: ignore   # mypy seems wrong here, None grads have been filtered out
```
joshim5, line 237: Hey @blefaudeux, this implementation probably won't work for model parallelism - the gradients need to be all reduced before applying this operation.

Are you planning to maintain support for model parallelism? I'm happy to contribute a PR! One approach is to mark each parameter in the `mpu` with a bit showing whether it's part of model parallelism and then allreduce just those gradients.

fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+                        for local_params_device in local_params
+                        for p in local_params_device
+                    ]
+                ),
+                norm_type,
+            )
+
+        # Sync all the norms from all ranks
+        norms = [torch.zeros_like(local_norm) for k in range(self.world_size)]
+        dist.all_gather(norms, local_norm, group=self.group, async_op=False)
```
joshim5, line 256: This gather assumes that each rank in the group has the same gradients. Otherwise, you'll end up with different clipping even if the ranks are the same. This will lead to strange behavior in the model parallelism case that is difficult to track down.

Continuing on comment from above, if we allreduce the model parallelism gradients above, then gradients should be the same for the same rank and your assumption will hold in model parallelism.

blefaudeux: > This gather assumes that each rank in the group has the same gradients. 

hmm, that was not the intent, each rank computes the norm on the gradient it has (they differ), then they all_gather so that everyone gets the overall max ?

fairscale/optim/oss.py
blefaudeux: I've little visibility in the model parallelism part, I'm really trying to keep it into account but as you can see it's not obvious. I did have a specific DDP in mind when writing that (for instance #157), which removes gradients from shards which don't own it and auto-reduce (not all-reduce), I realize now that the computation is wrong if it's not the case but then the interface will get a little clumsy

fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+
+        .. warning: This needs to be called on all ranks, since synchronization primitives will be used
+        """
+
+        # Compute the max norm for this shards's worth of gradients
+        max_norm = float(max_norm)
+        norm_type = float(norm_type)
+
+        # Filter out the grad-less params, concatenate params from all devices
+        local_params = [
```
blefaudeux, line 229: @joshim5 this should consider all the params from all the devices on this rank, so that was my tentative acknowledgement that model parallel could be in use

fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+                        for p in local_params_device
+                    ]
+                ),
+                norm_type,
+            )
+
+        # Sync all the norms from all ranks
+        norms = [torch.zeros_like(local_norm) for k in range(self.world_size)]
+        dist.all_gather(norms, local_norm, group=self.group, async_op=False)
+        total_norm = torch.max(torch.stack(norms))
```
msbaines, line 257: I don't think the logic above is correct. For regular norms, you could do something like this:
```suggestion
        total_norm = local_norm ** norm_type
        total_norm = dist.all_reduce(total_norm, group=self.group)
        total_norm = total_norm ** (1/norm_type)
```
You would need different logic for inf norm.

tests/optim/test_oss.py
```diff
@@ -429,3 +429,49 @@ def test_multiple_groups():
...
+        optimizer.clip_grad_norm(0.3, norm_type=1.0)
+        for params in optimizer.per_device_params.values():
+            for param in filter(lambda x: x.grad is not None, params[rank]):
+                assert torch.norm(param.grad, p=1.0) < 0.3, f"param grad norm above clip : {param.grad}"
+        return loss
+
+    _ = optimizer.step(closure=closure)
+
+
+def test_gradient_clipping():
```
msbaines, line 469: I think you need a test that verifies correctness. You want to verify that shared and un-sharded total_norm is the same.

fairscale/optim/oss.py
```diff
@@ -10,6 +10,7 @@
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type
 
 import torch
+from torch._six import inf
```
msbaines, line 13:
```suggestion
from math import inf
```

msbaines (changes requested):
Please add a more detailed description in the commit message and doc string.

Also, please add a more rigorous test of correctness. Perhaps, verifying that unsharded and sharded total_norm is the same for a few different norm types.

msbaines (changes requested):
Also, please update API docs and tutorials.

fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
 
         return loss
 
+    def clip_grad_norm(self, max_norm: float, norm_type: float = 2.0) -> torch.Tensor:
```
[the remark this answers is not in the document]
msbaines: I think we should support all the same types as torch.norm: 'frob', 'nuc', inf,  int, float

fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+                p.grad.detach().abs().max().to(self._device)  # type: ignore   # mypy seems wrong here, None grads have been filtered out
+                for local_params_device in local_params
+                for p in local_params_device
+            )
+
+        else:
+            local_norm = torch.norm(  # type: ignore
+                torch.stack(
+                    [
+                        torch.norm(p.grad.detach(), norm_type).to(self._device)  # type: ignore
```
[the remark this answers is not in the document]
blefaudeux: that was a week ago so I lost some context, but I guess that torch.norm is not known.. I can add a signature about that, a bit sad that we have to duplicate torch interface in fairscale

fairscale/optim/oss.py
joshim5: @msbaines: Is this correct? `self.group` is the Data Parallel group -- we shouldn't be summing across data parallel workers for the clip norm. Instead, we should be summing across model parallel units. 

When making this change, it's important not to overcount. One way is to flag each model parallel weight with a bit saying whether it's part of a model parallel calculation or just a repeated parameter.

Consider these two trivial examples:
1) Model parallel units don't have any model parallel operations; they have the exact same parameters. When computing the clip norm, if you allreduce over model parallel units, you'll double count!
2) Model parallel units only have model parallel operations. In this setting, you're okay.

fairscale/optim/oss.py
joshim5: Okay, in that case I think it would work if you did the gather. If you switch to allreduce as Mandeep is proposing (better) then you'll need to be careful here.

fairscale/optim/oss.py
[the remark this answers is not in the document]
msbaines: Yes. Please add torch.norm to our stubs.

fairscale/optim/oss.py
```diff
@@ -201,6 +202,72 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+
+        Returns:
+            Total norm of the parameters (viewed as a single vector).
+
+        .. note: This is analogous to `torch.nn.utils.clip_grad_norm_` but handles the partitioning and multiple devices per rank
+        under the hood.
+
+        .. warning: This needs to be called on all ranks, since synchronization primitives will be used
+        """
+
```
joshim5, line 223: As discussed over chat, I'll add in the model parallel case. In the meantime, could you add the below so that there aren't any silent bugs?

```suggestion
if self.group != dist.group.WORLD:
  raise NotImplemented("Clip norm not yet supported for model parallelism (coming soon!)")
```

fairscale/optim/oss.py
msbaines: This feedback was specific to Data Parallel. Normally, after DDP backward, each worker has access to all_reduced gradients for every parameter. So a local total_norm calculation will be consistent across workers. With optimizer state sharding, you can avoid all_reduce-ing gradients and instead just reduce on the worker with the shard. But if you do that, workers no longer have access to all gradients. That is the rationale for an OSS-specific clip_grads.

To get a correct total_norm, each worker needs to calculate it own norm and then combine it with the local_norms from other workers to get a total_norm.

blefaudeux: @msbaines you're right, I had inf in mind but that's mostly wrong, fixing this

fairscale/optim/oss.py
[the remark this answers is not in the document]
blefaudeux: done, it does not solve mypy being completely lost with the filtered iterables, but that's something else

fairscale/optim/oss.py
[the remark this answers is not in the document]
blefaudeux: @msbaines https://pytorch.org/docs/stable/_modules/torch/nn/utils/clip_grad.html#clip_grad_norm_ supports int/float/inf

blefaudeux:
hmm not sure what breaks on CI, works on my machine (famous last words). Looking into that

tests/optim/test_oss.py
```diff
@@ -422,10 +439,89 @@ def closure():
...
 def test_multiple_groups():
     world_size = 6
+    temp_file_name = tempfile.mkstemp()[1]
+
+    mp.spawn(
+        run_test_multiple_groups, args=(world_size, temp_file_name), nprocs=world_size, join=True,
+    )
+
+
+def run_gradient_clipping(rank, world_size, tempfile_name):
```
blefaudeux, line 455: this is the new test, checking that we get the same gradient norms vs. pytorch

blefaudeux:
ping reviewers, if you have a minute

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+        Arguments:
+            max_norm (float or int): max norm of the gradients
+            norm_type (float or int): type of the used p-norm. Can be ``'inf'`` for infinity norm.
+
+        Returns:
+            Total norm of the parameters (viewed as a single vector).
+
+        .. note: This is analogous to `torch.nn.utils.clip_grad_norm_` but handles the partitioning and multiple devices per rank
+            under the hood. The default torch util is not applicaple here, because each rank only has a partial view of all the grads
```
joshim5, line 219: nit: applicable spelled incorrectly

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+            in the model, so calling it in the OSS context would lead to different scaling being applied per subset of model parameters
+
+        .. warning: This needs to be called on all ranks, since synchronization primitives will be used
+
+        .. warning: Model paralelism -groups other than world- are not yet supported
+        """
+
+        if self.group != dist.group.WORLD:
```
joshim5, line 227: great! thanks for adding this

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+        local_params = itertools.chain(
+            *[
+                list(filter(lambda x: x.grad is not None, device_params[self.rank]))
+                for device_params in self.per_device_params.values()
+            ]
+        )
+
+        # Compute the norm on this grad set
+        local_norm = torch.norm(
+            input=torch.stack([torch.norm(input=p.grad.detach().to(self._device), p=norm_type) for p in local_params]),  # type: ignore
```
joshim5, line 244:
```suggestion
            input=torch.stack([torch.norm(input=p.grad.detach(), p=norm_type).to(self._device) for p in local_params]),  # type: ignore
```
Assuming the grads are first on CPU (correct me if I'm wrong), this could save memory while potentially being faster.

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+            p=norm_type,
+        )
+
+        # Sync all the norms from all ranks
+        if norm_type == inf:
```
joshim5, line 249: where `inf` defined?

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+
+        # Filter out the grad-less params, concatenate params from all devices
+        local_params = itertools.chain(
+            *[
+                list(filter(lambda x: x.grad is not None, device_params[self.rank]))
+                for device_params in self.per_device_params.values()
+            ]
+        )
+
+        # Compute the norm on this grad set
```
joshim5, line 242: I noticed you already break out the logic seperately for inf-norm below. In that case, why not avoid the `torch.stack` altogether in that case? This will be faster and reduce memory pressure.
```suggestion
        # Compute the norm on this grad set
        if norm_type == inf:
          local_norm = max(p.grad.detach().abs().max().to(self_device) for p in local_params)
```

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+        local_norm = torch.norm(
+            input=torch.stack([torch.norm(input=p.grad.detach().to(self._device), p=norm_type) for p in local_params]),  # type: ignore
+            p=norm_type,
+        )
+
+        # Sync all the norms from all ranks
+        if norm_type == inf:
+            norms = [torch.zeros_like(local_norm) for k in range(self.world_size)]
+            dist.all_gather(norms, local_norm, group=self.group, async_op=False)
```
joshim5, line 251: There is no need to gather all the tenors, since PyTorch distributed supports a `MAX` reduction. As below, you can do
`dist.all_reduce(total_norm, op=ReduceOp.MAX, group=self.group)`

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+        else:
+            total_norm = local_norm ** norm_type
+            dist.all_reduce(total_norm, group=self.group)
+            total_norm = total_norm ** (1.0 / norm_type)
+
+        clip_coef = torch.tensor(max_norm, dtype=total_norm.dtype, device=total_norm.device) / (total_norm + 1e-6)
+
+        if clip_coef < 1:
+            for device, device_params in self.per_device_params.items():
+                clip_coef = clip_coef.to(device)
```
joshim5, line 262: As written, this is going to involve unneeded device-to-device transfer. For example, `clip_coef` starts on CPU, then device 0, then device 1, etc. The device_i -> device_{i+1} is unnecessary and will reduce performance.

joshim5:
```suggestion
```

fairscale/optim/oss.py
```diff
@@ -201,6 +203,70 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+
+                for p in filter(lambda x: x.grad is not None, device_params[self.rank]):
+                    p.grad.detach().mul_(clip_coef)  # type: ignore
```
joshim5, line 265:
```suggestion
                    p.grad.detach().mul_(clip_coef.to(p.grad.device))  # type: ignore
```
Continuing from comment above, I recommend moving to device here so that you simply do CPU -> device_i each time.

joshim5 (changes requested):
Recommending a few changes to improve performance of the implementation.

fairscale/optim/oss.py
blefaudeux: from math ?

fairscale/optim/oss.py
blefaudeux: makes sense, thanks for the heads up I totally forgot about that

fairscale/optim/oss.py
blefaudeux: the grads may or may not be on cpu (most probably not actually), this was mostly to make sure that the multi-device case was handled properly, but in any case I think that your suggestion makes a lot of sense, not so small detail.. thanks

fairscale/optim/oss.py
joshim5: I missed the import, thanks

fairscale/optim/oss.py
blefaudeux: it's surprising to me that this would have a meaningful effect, we're talking about a single-item tensor which is moved n-times, n being the number of devices (~8 max ?). Updating the PR still but I'm wondering whether I'm missing something ?

fairscale/optim/oss.py
blefaudeux: doing that, I'm not entirely convinced to be honest but still fine by me :)

examples/tutorial_oss.py
```diff
@@ -44,26 +45,39 @@ def train(rank: int, world_size: int, epochs: int, use_oss: bool):
     dataloader = getData()
     loss_fn = getLossFun()
 
-    base_optimizer_arguments = {"lr": 1e-4}  # any optimizer specific arguments, LR, momentum, etc...
+    optimizer: Optional[Union[OSS, torch.optim.SGD]] = None
+
     if ~use_oss:
```
min-xu-ai, line 50: use ~ is very strange here, "not use_oss" is better?

examples/tutorial_oss.py
```diff
@@ -44,26 +45,39 @@ def train(rank: int, world_size: int, epochs: int, use_oss: bool):
...
             model.zero_grad()
             outputs = model(data)
             loss = loss_fn(outputs, target)
-            loss /= world_size
             loss.backward()
+
+            # if you want to clip the gradients / get the current max:
+            max_norm = 1000.0
+            norm_type = 1
+            if ~use_oss:
```
min-xu-ai, line 73: same, ~ is strange. it will always return True?
```
>>> ~False
-1
>>> ~True
-2
```

fairscale/optim/oss.py
```diff
@@ -201,6 +203,67 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
+            dist.all_reduce(total_norm, op=torch.distributed.ReduceOp.MAX, group=self.group)
+        else:
+            local_norm = torch.norm(
+                input=torch.stack([torch.norm(input=p.grad.detach(), p=norm_type).to(self._device) for p in local_params]),  # type: ignore
+                p=norm_type,
+            )
+
+            total_norm = local_norm ** norm_type
+            dist.all_reduce(total_norm, group=self.group)
+            total_norm = total_norm ** (1.0 / norm_type)
```
min-xu-ai, line 255: I don't quite understand these 3 lines. add a comment?

stubs/torch/__init__.pyi
```diff
@@ -596,6 +596,7 @@ class Tensor:
     def mode(self, dim: _int=-1, keepdim: _bool=False) -> Tuple[Tensor, Tensor]: ...
     @overload
     def mode(self, dim: Union[str, None], keepdim: _bool=False) -> Tuple[Tensor, Tensor]: ...
+    def mul_(self, value: Union[_float, _int, Tensor]): ...
```
min-xu-ai, line 599: thanks for adding!

min-xu-ai (commented):
nice! some questions below

fairscale/optim/oss.py
blefaudeux: n = sum(a^p)^1/p -> n_total = all_reduce(n^p)^(1/p), basically the sum is factorizable if everything is ^p

blefaudeux: adding a comment

examples/tutorial_oss.py
blefaudeux: Ah this is not from me actually, from @Vittorio-Caggiano I think, I just removed a line and fixed linting, didn't look beyond that. I'll change that

Vittorio-Caggiano: yes, my bad

min-xu-ai (approved):
LGTM!

PR #176, opened 2020-11-05
[refactor] moe: compute softmax after argmax

msbaines:
According to @TimDettmers:

So the flow from the balancing loss to the gate is the same. The
difference is when we multiply the softmax probability by the
expert we have a softmax gradient flow from the expert to the gate.
The problem with that is that (1) softmax gradients are small since
the gradient is p*(1-p), (2) if the gradient is close to 0 then the
deltas/errors coming from the expert are multiplied by this value on
the way to the gate. This means that a probability close to zero
should be avoided or otherwise errors from the 2nd expert will not
update the gate and the gate "does not know about the improvement to
the expert". Adam will help with this expect a little bit, since the
running differences for such an expert are small but we only get on
average every k/num_experts samples an running differences update
for adam. So adam might not be able to adjust quickly enough for a
situation like this.

The second piece is since we normalize in the softmax by the sum of
all exponentiated terms, and each term is positive, the normalizing
terms grows with the size of the number of experts. So if we have
1024 experts, it would be very likely that most of the probability
terms are close to zero and hence have very small gradients. If we
just take the top-k values and normalize them, then the normalizing
term is much smaller and the minimum probability is still sizable to
allow gradient flow from expert to gate.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/moe/top2gate.py
+++ b/fairscale/nn/moe/top2gate.py
@@ -28,17 +28,15 @@ def gumbel_rsample(shape: Tuple, device: torch.device) -> Tensor:
 
 def top2gating(logits: torch.Tensor) -> Tuple[Tensor, Tensor, Tensor]:
     """Implements Top2Gating on logits."""
-    gates = F.softmax(logits, dim=2)
-
-    # gates has shape of GSE
-    num_tokens = gates.shape[1]
-    num_experts = gates.shape[2]
+    # logits has shape of GSE
+    num_tokens = logits.shape[1]
+    num_experts = logits.shape[2]
     # capacity = 2S/E
     capacity = 2 * num_tokens // num_experts
     assert num_tokens % num_experts == 0
 
     # Create a mask for 1st's expert per token
-    indices1_gs = torch.argmax(gates, dim=2)
+    indices1_gs = torch.argmax(logits, dim=2)
     mask1 = F.one_hot(indices1_gs, num_classes=num_experts)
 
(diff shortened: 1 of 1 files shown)


min-xu-ai (commented):
maybe I am missing something. the new version of the code doesn't need a softmax anymore? Or is it done elsewhere?

msbaines:
> maybe I am missing something. the new version of the code doesn't need a softmax anymore? Or is it done elsewhere?

Softmax is done "by hand", see line 67-69.

fairscale/nn/moe/top2gate.py
```diff
@@ -71,8 +64,8 @@ def top2gating(logits: torch.Tensor) -> Tuple[Tensor, Tensor, Tensor]:
     # Normalize gate probabilities
     mask1_float = mask1.float()
     mask2_float = mask2.float()
-    gates1_gs = torch.einsum("gse,gse->gs", gates, mask1_float)
-    gates2_gs = torch.einsum("gse,gse->gs", gates, mask2_float)
+    gates1_gs = torch.einsum("gse,gse->gs", torch.exp(logits), mask1_float)
+    gates2_gs = torch.einsum("gse,gse->gs", torch.exp(logits), mask2_float)
```
min-xu-ai, line 68: will torch be smart enough and perform a single torch.exp on logits here or it will be done twice on the same data?

PR #181, opened 2020-11-10
[ShardedOptim/ShardedDDP] Distributed Grad Scaler (for torch AMP)

### LINKED ISSUE #180: [OSS] Handle torch AMP auto skip inf/nan steps
## 🚀 Feature
Using torch amp can lead to step() being skipped in the presence of inf or nans. An issue with that with OSS is that each rank will see a different step, meaning that some ranks could skip and others try to step (typically vanishing gradient on the way up and underflow with fp16). We need to handle that properly

## Motivation
Without that using torch amp with OSS can easily deadlock, see for instance https://github.com/facebookresearch/fairscale/pull/179

## Pitch
A somewhat elegant and not too far reaching solution to that could be to add a config flag to OSS, following which each rank will check for inf/nan on its grads, and signal all ranks that they should skip step() in that case. Users willing to apply OSS and torch amp would be advised to switch this flag on to prevent deadlocks

## Alternatives
Prevent mixed precision and OSS, but that makes OSS a lot less compelling

## Additional context
#179 and https://pytorch.org/docs/stable/notes/amp_examples.html?highlight=mixed%20precision#id2

blefaudeux:
# Before submitting
- [x]  Was this discussed/approved via a Github issue? (no need for typos, doc improvements)

## What does this PR do?
Fixes #180, supersedes the grad scaler so that it syncs the is_finite status across all the ranks. Trying to use the vanilla GradScaler would hang, because of the underflow test which would differ in between ranks and conditions the step(). This enables torch AMP with its scaler with minimal changes (just use ShardedGradScaler instead of torch's)

Credit: Most of the code or code ideas are from Sean (@seannaren) 

## PR review

## Did you have fun?


blefaudeux:
The changes are actually minimal, but depends on other PRs which are not landed.

blefaudeux:
the speed seems impacted even if it does not make a lot of sense, maybe a rebasing issue. Fixing that

blefaudeux:
Rebasing on current master, somehow I messed up the begging of this branch

PR #185, opened 2020-11-11
cpu support for benchmarks/oss.py

yuanyuan0057:
# Before submitting

- [X] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?

## What does this PR do?
Fix issues of benchmark/oss.py when trying to run it from cpu. Make two changes:
1. set device_ids to None when calling DDP as required by [PyTorch DDP](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html?highlight=distributeddataparallel#torch.nn.parallel.DistributedDataParallel)
2. remove max_allocated_memory when cpu is in use. For cuda, if not args.cpu will be true and max_allocated_memory will still be displayed

## PR review
Talked with @blefaudeux on the issue.

## Did you have fun?

### DIFF
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -104,7 +104,10 @@ def train(
         optimizer = model.sharded_optimizer
 
     else:
-        model = DDP(model, device_ids=[rank], find_unused_parameters=False)  # type: ignore
+        if args.cpu:
+            model = DDP(model, device_ids=None, find_unused_parameters=False)  # type: ignore
+        else:
+            model = DDP(model, device_ids=[rank], find_unused_parameters=False)  # type: ignore
         optimizer = (
             OSS(params=model.parameters(), optim=OPTIM, lr=1e-4, momentum=0.9)
             if optim_type == OptimType.oss_ddp
@@ -211,10 +214,7 @@ def train(
 
     training_stop = time.monotonic()
     img_per_sec = n_items / (training_stop - training_start) * args.epochs
-    max_memory = torch.cuda.max_memory_allocated(rank) / 2 ** 20
-
     logging.info(f"[{dist.get_rank()}] : Training done. {img_per_sec:.2f} img per sec inc. checkpoint")
-    logging.info(f"[{dist.get_rank()}] : Peak memory {max_memory:.1f}MiB")
 
     # Compute the mean and average img per second
     mean = sum(measurements) / len(measurements)
--- a/benchmarks/oss.py
+++ b/benchmarks/oss.py
@@ -105,9 +105,10 @@ def train(
 
     else:
         if args.cpu:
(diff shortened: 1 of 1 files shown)


benchmarks/oss.py
```diff
@@ -211,10 +214,7 @@ def closure():
 
     training_stop = time.monotonic()
     img_per_sec = n_items / (training_stop - training_start) * args.epochs
-    max_memory = torch.cuda.max_memory_allocated(rank) / 2 ** 20
```
min-xu-ai, line 214: should we still have this log when it is not on CPU?

benchmarks/oss.py
```diff
@@ -104,7 +104,10 @@ def train(
         optimizer = model.sharded_optimizer
 
     else:
-        model = DDP(model, device_ids=[rank], find_unused_parameters=False)  # type: ignore
+        if args.cpu:
+            model = DDP(model, device_ids=None, find_unused_parameters=False)  # type: ignore
```
min-xu-ai, line 108: perhaps conditionally set the device_ids's value to a variable and avoid duplicating the entire model = DDP() line in the if and else cases?

benchmarks/oss.py
yuanyuan0057: We already have that on line 210.:     
if not args.cpu:
        torch.cuda.synchronize(rank)
        max_memory = torch.cuda.max_memory_allocated(rank) / 2 ** 20
        logging.info(f"[{dist.get_rank()}] : Peak memory {max_memory:.1f}MiB")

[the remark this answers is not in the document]
yuanyuan0057: Same as previous reply. It was already set previously in line 210:
    if not args.cpu:
        torch.cuda.synchronize(rank)
        max_memory = torch.cuda.max_memory_allocated(rank) / 2 ** 20
        logging.info(f"[{dist.get_rank()}] : Peak memory {max_memory:.1f}MiB")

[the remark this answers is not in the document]
blefaudeux: true that, the diff presentation here is wrong, I'm not sure why it shows these lines as being deleted ?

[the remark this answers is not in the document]
yuanyuan0057: Yes, I deleted this line. This line is after the one in line 210. So on cuda, this line was evaluated twice.

benchmarks/oss.py
yuanyuan0057: Sure. This is done in the second commit

blefaudeux (approved):
Looks good to me, thanks for the PR Yuanyuan !

PR #193, opened 2020-11-18
Workaround sporadic crash on shutdown in test_python_autograd_function[PipelineStyle.MultiProcess]

froody:
Works around https://github.com/pytorch/pytorch/issues/47161 by calling rpc.shutdown() before exiting. My theory is that the race condition occurs when an rpc thread runs during shutdown, so I assume that calling rpc.shutdown() will prevent that. Local testing seems to indicate that it works.


blefaudeux (approved):
Theory sounds legit, that's all I know ! Thanks for the quick fix

PR #207, opened 2020-11-24
[doc] make the basic example more "compilable"

stas00:
I realize that this is a prototype example, this PR makes it just slightly easier to follow by adding the missing import for `dist` and writing out `dist_init` as the reader may not know what is it supposed to be.

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ def train(
     world_size: int,
     epochs: int):
 
-    # DDP
+    # DDP init example
     dist.init_process_group(backend='nccl', init_method="tcp://localhost:29501", rank=rank, world_size=world_size)
 
     # Problem statement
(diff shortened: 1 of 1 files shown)


blefaudeux:
Thanks a lot for the fix ! I think that this came from an example where 'dist_init' was defined as a helper function, but not the case here so that's confusing indeed.. Looks good to me

README.md
```diff
@@ -64,7 +65,7 @@ def train(
     epochs: int):
 
     # DDP
-    dist_init(rank, world_size)
+    dist.init_process_group(backend='nccl', init_method="tcp://localhost:29501", rank=rank, world_size=world_size)
```
blefaudeux, line 68: If you don't mind adding a comment stating that this setting is an example (port number for instance?)

blefaudeux (approved):
LGTM, thanks a lot for the fix !

README.md
blefaudeux: Never mind, this is just an example, should be fine. Merging as is, thanks again

PR #223, opened 2020-12-03
[fix] Fix iGPT buckets with ShardedDDP

### LINKED ISSUE #222: [ShardedDDP] Bucketing seems to break with ImageGPT
## 🐛 Bug
It looks like the reduce when BW breaks with this setup, although it's fine when bucketing is disabled.
See https://github.com/SeanNaren/image-gpt for context

## To Reproduce

```
pip install https://github.com/facebookresearch/fairscale/archive/master.zip
pip install https://github.com/PyTorchLightning/pytorch-lightning/archive/master.zipgit clone https://github.com/SeanNaren/image-gpt.git
cd image-gpt/ && git checkout feature/cuda && pip install -e .
cd ..# You might need to reduce the number of GPUS in the config
python src/run.py --dataset mnist train configs/s_gen.yml --sharded
```
-> Assert that buckets overflow

## Expected behavior
Should run drama-free 

## Additional context
cc @seannaren

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #222. Overall I think that the bucketing is a little more elegant (the unroll of the bucket, be it after reduce or broadcast is factorized instead of being duplicated), the buckets store the destination and current rank once and for all so that callbacks are smaller, but I'm wondering whether tackling bucket-as-tensor-view directly would not be nicer. I'm also not sure of what to add as a unit test to catch this iGPT case, main idea is that some of the autograd hooks do not fire somehow (this is beyond ShardedDDP and a case that we need to handle anyway), so the buckets were not flushed in between the phases, open for suggestions.

## PR review
cc @seannaren @mrshenli

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -229,22 +229,6 @@ class ShardedDataParallel(nn.Module):
                 )
 
                 if bucket.full():
-
-                    def unwrap() -> None:
-                        for flat in bucket.params:
-                            if dst_rank != self.global_rank:
-                                # this rank is not the owner, release the grad
-                                flat.param.grad = None
-                            else:
-                                # this rank is the owner, unroll the results
-                                assert flat.param.grad is not None
-
-                                flat.param.grad.data.copy_(
-                                    bucket.buffer[flat.start : flat.stop].view_as(flat.param.data), non_blocking=True
-                                )
-
-                        bucket.reset()
-
                     bucket.buffer /= self.world_size
 
                     optimizer.work_handles.append(
@@ -252,7 +236,7 @@ class ShardedDataParallel(nn.Module):
                             handle=dist.reduce(
                                 tensor=bucket.buffer, dst=dst_rank, group=self.process_group, async_op=True,
                             ),
-                            callback=unwrap,
+                            callback=bucket.unroll,
                         )
                     )
 
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -112,6 +112,9 @@ class ShardedDataParallel(nn.Module):
             # for the subsequent FW to be correct
             self.sync_buffers(blocking=True)
 
+        # Reset all the grad reduce and bucket state flags
+        self._grad_to_be_reduced = [True] * len(self._grad_to_be_reduced)
+
         # Normal FW on the base model
         return self.module(*inputs, **kwargs)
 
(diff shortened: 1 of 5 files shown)


min-xu-ai (commented):
is it possible to add a test to exercise this case? if this is needed urgently, fine to me to merge with a test to be added later.

blefaudeux:
> is it possible to add a test to exercise this case? if this is needed urgently, fine to me to merge with a test to be added later.
yes I would like to, I'm not sure how to trigger that behaviour properly though, I alluded to that in the PR description. Basically some of the BW hooks do not fire (even if the params are trainable), it makes sense in some situations I think so ShardedDDP should be robust to that. I'll try to come up with an example, working on that

min-xu-ai (approved):

min-xu-ai:
> yes I would like to, I'm not sure how to trigger that behaviour properly though, I alluded to that in the PR description. Basically some of the BW hooks do not fire (even if the params are trainable), it makes sense in some situations I think so ShardedDDP should be robust to that. I'll try to come up with an example, working on that

yeah, I think we should collect all these kind of corner cases, like some params only used in some iterations, etc.

blefaudeux:
I've a unit test which works well now: just use the real GPT2, and sure enough this triggers. Problem is, it uncovered a real big issue with the bucketing, in that if some BW hooks do not fire (proof is this can happen) *and* the corresponding params were part of a bucket, then the bucket is never sent. The "fix" in this PR is not correct actually, the only fix that I can figure out right now is not to bucket.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -179,9 +182,6 @@ def gatekeeper() -> None:
             # and execute the delayed actions (release gradients, unroll the buckets)
             Variable._execution_engine.queue_callback(optimizer._consume_work_handles)
 
-            # Reset all the grad reduce and bucket state flags
```
blefaudeux, line 182: this was risky in that maybe that we never go there, if all grads are not materialized. Moved to the beginning of the FW, which will always happen

fairscale/utils/testing.py
```diff
@@ -211,3 +212,81 @@ def replacement(*args: Any, **kwargs: Any) -> None:
...
+
+        x = self.ln_1(x)
+        a, _ = self.attn(x, x, x, attn_mask=attn_mask, need_weights=False)
+        x = x + a
+        m = self.mlp(self.ln_2(x))
+        x = x + m
+        return x
+
+
+class GPT2(nn.Module):
```
blefaudeux, line 238: does not take a lot of space, and makes GPT available for testing without any dependency, it's a proven good corner case.

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -233,4 +236,49 @@ def test_two_optimizers():
     backend = "gloo"
     temp_file_name = tempfile.mkstemp()[1]
     device = "cpu"
-    mp.spawn(run_test_two_inputs, args=(world_size, backend, device, temp_file_name), nprocs=world_size, join=True)
+    mp.spawn(run_test_two_optimizers, args=(world_size, backend, device, temp_file_name), nprocs=world_size, join=True)
+
+
+def run_test_gpt2(rank, world_size, backend, device, temp_file_name):
```
blefaudeux, line 242: this test correctly asserts with the prior bucketing implementation, as @Seannaren first experienced

blefaudeux:
Landing since approved, I think that it's important that master is always working and this enforces it. I'll need to get back to the bucketing part though, either do without and clean up the codebase or devise a better way which works

PR #229, opened 2020-12-04
[feat]: AdaScale work with lr_scheduler and tests, examples

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


fairscale/optim/adascale.py
```diff
@@ -334,9 +340,16 @@ def zero_grad(self) -> None:
         return self._optimizer.zero_grad()
 
     def state_dict(self) -> Dict:
-        """Proxy function to optimizer, checkpointing needs this."""
+        """ Proxy function to optimizer, checkpointing needs this.
+
+            Note: Do NOT checkpoint in the middle of gradient accumulation since
```
blefaudeux, line 345: `.. note:` ? (see https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html)

tests/optim/test_single_node_adascale.py
```diff
@@ -94,6 +94,9 @@ def test_grad_accum_gpu():
 def test_state_checkpointing():
     """ Test state checkpointing on GPU since that's the common case.
 
+        Note, we don't support checkpointing in the middle of gradient accumulation
```
blefaudeux, line 97: same

blefaudeux (approved):
LGTM, minor nitpicking on the rst syntax to get prettier html out of it, but that's all

fairscale/optim/adascale.py
min-xu-ai: Great idea! Will address both places!

README.md
```diff
@@ -114,6 +114,21 @@ AdaScale can be used to wrap a SGD optimizer and to be used in DDP (Distributed
...
+from fairscale.optim import AdaScale
+
+...
+optim = AdaScale(SGD(model.parameters(), lr=0.1))
+...
+step = 0
+for sample in dataloader:
+    ...
+    step += optim.gain()
+    update_lr(step)  # This update_lr() function is unchanged from single GPU training.
```
mikerabbat, line 128: The standard PyTorch learning rate schedulers (in `torch.optim.lr_scheduler`) have a different syntax. Instead of passing `step` as an argument, they use a syntax like:
```
>>> optimizer = torch.optim.SGD(model.parameters(), lr=0.1, momentum=0.9)
>>> scheduler = torch.optim.lr_scheduler.CyclicLR(optimizer, base_lr=0.01, max_lr=0.1)
>>> data_loader = torch.utils.data.DataLoader(...)
>>> for epoch in range(10):
>>>     for batch in data_loader:
>>>         train_batch(...)
>>>         scheduler.step()
```
I have an idea of how we could fix this pretty easily. We should be keeping track of the scale-invariant step number internally to AdaScale (accumulate the gain every time `step()`) is called. Then we could add a method to AdaScale that returns the number of times to call `scheduler.step()` after calling `optim.step()`.

One other point - the example should probably include the call to `optim.step()`

mikerabbat (commented):
Not sure if this is a big deal, but the way learning rate scheduling is shown in the examples is different from how it is done with the standard PyTorch schedulers.

I don't know how standardized this is, though. Other packages (fairseq, classy_vision) appear to implement their own scheduling logic which is more similar to the one in the documentation.

README.md
min-xu-ai: I see. That's a great observation! I didn't know about the standard way either. I will give this a try in the example. Stay tuned for new version of this PR later.

min-xu-ai:
Hi, thanks for the suggestions reviewers! Can you take another look?

docs/requirements.txt
```diff
@@ -3,4 +3,4 @@ sphinx==3.2.1
 sphinx_rtd_theme==0.4.3
 sphinxcontrib-programoutput==0.16
 git+https://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme
-torch==1.6.0
+torch>=1.6.0
```
blefaudeux, line 6: oh, nice catch

fairscale/optim/adascale.py
```diff
@@ -38,27 +38,36 @@
...
         adascale = AdaScale(optim)
 
+        step = 0
         for epoch in ...:
             for batch in ...:
                 optim.zero_grad()
                 loss = ...
                 loss.backward()
+                step += adascale.gain()
+                update_lr(step)
```
mikerabbat, line 70: Would it make sense to use the same example code across Readme, rst, and pydoc?

fairscale/optim/adascale.py
```diff
@@ -38,27 +38,36 @@
 import torch
 from torch.autograd import Variable
 import torch.distributed as dist
+from torch.optim import Optimizer
 
 
-class AdaScale(object):
+class AdaScale(Optimizer):
```
mikerabbat, line 44: Not sure this is needed.

In the documentation example in the rst file, things are instantiated in the following order:
```
optimizer = [base optimizer class]
scheduler = ...
optimizer = AdaScale(optimizer)
```

As long as things are done in that order, shouldn't everything work ok without needing AdaScale to subclass Optimizer?

mikerabbat: Looks like things happen in a different order in `test_lr_scheduler`

mikerabbat (commented):
Few additional comments/suggestions

README.md
```diff
@@ -117,6 +117,31 @@ AdaScale can be used to wrap a SGD optimizer and to be used in DDP (Distributed
...
+optim = AdaScale(SGD(model.parameters(), lr=0.1))
+scheduler = LambdaLR(optim, ...)
+...
+# Note: the train loop should be with DDP or with gradient accumulation.
+last_epoch = 0
+step = 0
+done = False
+while not done:
+    for sample in dataset:
+        ...
```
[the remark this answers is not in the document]
min-xu-ai: Good catch! Will do.

docs/source/tutorials/adascale.rst
```diff
@@ -75,18 +79,30 @@ as follows.
...
                 outputs = model(data)
                 loss = loss_fn(outputs, target)
                 loss.backward()
+                step += optimizer.gain()
                 optimizer.step()
+                epoch = step // len(dataloader)
+                if last_epoch != epoch:
+                    scheduler.step()
+                    last_epoch = epoch
+                if epoch >= len(epochs):
```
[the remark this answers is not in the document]
min-xu-ai: Right! Sharp eyes! Will fix.

fairscale/optim/adascale.py
min-xu-ai: I was thinking that with a varieties of examples, maybe it helps the user more in different situations?

fairscale/optim/adascale.py
min-xu-ai: I see. I meant the order in the test but I both orders are possible. Which one do you think is cleaner? It seems that if we only support a specific order, it is harder for people to understand the error when they reversed the order? What's the potential downside of subclassing from Optimizer?

fairscale/optim/adascale.py
mikerabbat: Sure, makes sense, though it could also potentially cause confusion. I'd suggest if we want to show multiple examples, to group them together in the same place (example 1, example 2, ...) and clearly point out that there's more than one way to use this.

min-xu-ai: Good idea. How about the following?

* we keep the README.md example more "official" (the lr_schedule one)
* keep the tutorial example as is
* here, we have two examples, one for each style

fairscale/optim/adascale.py
mikerabbat: The main downside I could imagine is that we don't currently wrap or implement all methods in the Optimizer interface. If someone uses one of the methods we don't explicitly wrap or override, it wouldn't have the expected effect, since it will operate on the adascale instance rather than adascale._optimizer. I don't have a strong opinion, though. Is it strange to have a class subclass Optimizer, but then internally have an instance of Optimizer that is really providing the main implementation?

fairscale/optim/adascale.py
mikerabbat: Sounds great!

docs/requirements.txt
msbaines: This file is used by the readthedocs service. The impact of this is that the readthedocs service will now use latest PyTorch after every commit to generate our documentation. That might be OK since we now try to support latest PyTorch ASAP.

msbaines (approved):
LGTM. Thanks for writing docs and tests.

fairscale/optim/adascale.py
min-xu-ai: We discussed more over chat. We will keep the subclass for now so that we maintain to be a full wrapper.

mikerabbat (approved):
Looks good to me. Thanks for taking all the feedback and suggestions into account!

PR #235, opened 2020-12-06
[fix] more adascale gradient accumulation tests and smoothing factor fix

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


min-xu-ai:
Do not review yet. Still working on it.

min-xu-ai:
I was suspecting bug but turns out that is not the case. This PR adds more tests to ensure DDP and grad-accum works exactly the same. Thanks for help, @mikerabbat

min-xu-ai:
@mikerabbat, interesting that maybe there is an indeed different between ddp and grad_accum modes?

see this diff:

```
@@ -127,7 +127,7 @@ class AdaScale(object):
         # Set smoothing based on world_size rather than scale here, since world_size
         # determines the number of samples being averaged over at every update
         self._smoothing = (
-            max(1 - self._world_size / 1000, 0)
+            max(1 - (self._world_size * self._num_grads_to_accum) / 1000, 0)
             if smoothing is None
             else smoothing
         )
```

min-xu-ai:
let me convert this back to draft and add the fix and test for smoothing factor.

mikerabbat:
Good catch on using `world_size * self._num_grads_to_accum` for default `smoothing`, rather than `world_size` alone!

The AdaScale paper recommends setting the smoothing value in this way. The motivation to use smoothing is that the estimate of the gain ratio at any point in time is noisy, and smoothing over time helps eliminate some of the noise/fluctuations. However, as we aggregate over more gradient samples, the amount of noise decreases, so we need less smoothing (hence limit 0).

There are some experiments on CIFAR10 in the paper (Appendix C) and the conclusion is that `max(1 - scale/1000, 0)` is reasonable. Let's leave it like this for now, but in the future this could be something to revisit since we've definitely seen that the smoothing value can impact performance.

min-xu-ai:
@mikerabbat, this is ready again. Thanks for the review in advance. I added a test that covers the smooth factor and I tried to reintroduce the bug and the test caught it.

mikerabbat (approved):
Looks good to me!

PR #237, opened 2020-12-07
[fix] Workaround Classy doing nasty things with optimizers

### LINKED ISSUE #236: [ShardedGradScaler] Handle optimizers not being OSS instances
## 🚀 Feature
Some frameworks (Classy for instance) change the optimizers to enable different non-pytorch-compliant features. ShardedGradScaler asserts on that

## Motivation
Enable ShardedDDP + AMP on Classy-like frameworks

## Pitch
Remove the assert, replace by a one time warning

## Alternatives
Not doing anything

## Additional context
@mannatsingh

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #236

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/grad_scaler.py
+++ b/fairscale/optim/grad_scaler.py
@@ -32,16 +32,16 @@ class ShardedGradScaler(TorchGradScaler):
 
     def __init__(self) -> None:
         super().__init__()
-        self.warning = True
+        self.display_warning = True
 
     def unscale_(self, optimizer: Optimizer) -> None:
         # Could be a mistake, this scaler is supposed to work with ZeroRedundancyOptimizer only
-        if self.warning and not isinstance(optimizer, OSS):
+        if self.display_warning and not isinstance(optimizer, OSS):
             logging.warning(
                 "ShardedGradScaler is to be used in combination with a sharded optimizer, this could not be checked"
             )
 
-        self.warning = False  # Only warn once
+        self.display_warning = False  # Only warn once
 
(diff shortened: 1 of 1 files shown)


blefaudeux:
cc @prigoyal, needed for the ShardedDDP + AMP run that I shared late last week

fairscale/optim/grad_scaler.py
```diff
@@ -31,9 +32,16 @@ class ShardedGradScaler(TorchGradScaler):
 
     def __init__(self) -> None:
         super().__init__()
+        self.warning = True
```
min-xu-ai, line 35: change to "warned"?

min-xu-ai (approved):
fair enough!

fairscale/optim/grad_scaler.py
blefaudeux: it's "not warned" right now, renaming to "display_warning" if it's ok for you, to be super explicit ?

prigoyal:
curious question: why do we need this fix?

blefaudeux:
> curious question: why do we need this fix?

Classy wraps the optimizer in a dedicated class, which does not inherit from the actual optimizer you requested (see https://github.com/facebookresearch/ClassyVision/blob/0673482cb41683c72d0630b4765122cccef074b9/classy_vision/optim/classy_optimizer.py#L47, the attributes are passed on so that it pseudo-inherits but the type is different), so this assert would fail

PR #242, opened 2020-12-09
[fix] Check ShardedDDP / DDP parity + bugfix

### LINKED ISSUE #241: [ShardedDDP] Race condition, different results from DDP
## 🐛 Bug
When running a training and comparing the results to DDP ([for instance](https://github.com/PyTorchLightning/pytorch-lightning/blob/a102f7bd18f4da8a8e4bd35227fa849f050fa89e/benchmarks/test_sharded_parity.py#L139-L167)), a recent commit broke parity

## Command
See link above

## Expected behavior
We should get the same results as DDP, out of principle

blefaudeux:
…ssue that Sean spotted

# Before submitting

## What does this PR do?
Fixes #241.
- add a unit test to make sure that we don't break on that front in the future
- fix a gross counting error with respect to the number of requests which can be issued, which was the basis for a gatekeeper sync (which was then often wrong). Note that this is unrelated to the iGPT training problem a priori (unused parameter), which will still need to be properly tackled

## PR review
cc @seannaren

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -560,7 +560,6 @@ class OSS(Optimizer):
                 work_handle.callback()
 
         self.work_handles.clear()
-        print(f"{self.rank} - Work handles consumed")
 
     def _handle_trailing_buckets(self, flush_type: BucketFlush) -> None:
         """
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -74,7 +74,7 @@ class OSS(Optimizer):
         broadcast_buffer_size: int = 2 ** 17,
         **default: Any,
     ):
-        logging.warning("Disabling bucketing for now, error prone for some models")
+        # logging.warning("Disabling bucketing for now, error prone for some models")
         broadcast_buffer_size = 0
 
         # Hold all the model params in the root .param_groups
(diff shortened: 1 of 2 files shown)


fairscale/optim/oss.py
```diff
@@ -613,5 +615,9 @@ def _setup_bucket_strategy(self) -> None:
                 self.buckets[device][dst_rank].global_rank = self.global_rank
 
         # Determine the max work handles in flight:
-        # - all the direct reduce/broadcast + 1 bucket
-        self._max_work_handles = sum(not value for value in self.should_bucket_param.values()) + 1
+        # - all the direct reduce/broadcast
+        self._max_work_handles = sum(not value for value in self.should_bucket_param.values())
```
blefaudeux, line 619: this was wrong:
- if no bucketing, then the +1 was wrong (no more requests, just the direct ones)
- if bucketing, the +1 was mostly wrong (one more request per bucket, which means one per device per rank)

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -123,6 +115,84 @@ def run_test(backend, device, world_size=2):
...
+    run_test(backend=dist.Backend.GLOO, device=torch.device("cpu"), world_size=4)
+
+
+@skip_if_no_cuda
+@skip_if_single_gpu
+def test_step_on_gpu():
+    run_test(backend=dist.Backend.NCCL, device=torch.device("cuda"))
+
+
+def run_ddp_parity(rank, world_size, backend, temp_file_name):
```
blefaudeux, line 128: new unit test which checks that we have the same results as DDP, so that this does not happen again

min-xu-ai (approved):
Great! Thanks for the test!

PR #243, opened 2020-12-10
[doc] updating the pipe balance doc a bit

min-xu-ai:
- Added some notes as I was using those two functions.
- Also added a warning to pipeline.py when the partition output is not supported.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/pipe/balance/__init__.py
+++ b/fairscale/nn/pipe/balance/__init__.py
@@ -23,13 +23,20 @@ Usage::
 
     import torch
     from fairscale.nn import Pipe
-    from fairscale.nn.balance import balance_by_time
+    from fairscale.nn.pipe.balance import balance_by_time
 
     sample = torch.empty(128, 3, 224, 224)
     balance = balance_by_time(torch.cuda.device_count(), model, sample)
 
     pipe = Pipe(model, balance, chunks=8)
 
+.. note::
+    balance_by_time does not work with inplace ReLU because we exhausetively search
+    every partition boundary, which could hit an inplace ReLU.
+
+.. note::
+    If the model is larger than a single CUDA device memory, use "cpu"
+    in the balance_by_time function.
 """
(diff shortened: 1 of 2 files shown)


fairscale/nn/pipe/balance/__init__.py
```diff
@@ -23,13 +23,20 @@
...
 
     pipe = Pipe(model, balance, chunks=8)
 
+.. note::
+    balance_by_time does not work with inplace ReLU because we exhausetively search
+    every partition boundary, which could hit an inplace ReLU.
+
+.. note::
+    If the model is larger than a single CUDA device memory, use "cpu"
+    in the balance_by_time function.
```
min-xu-ai, line 39: Just a note, when I was using it on the 10B model, it crashed with a segfault. Obviously I had to use the cpu device. So I didn't get a useful split out of this function. In the end, my pipeline split has to be done by hand tuning anyway. :-(

fairscale/nn/pipe/pipeline.py
```diff
@@ -180,7 +180,10 @@ def function(
             part_id: int = j,
         ) -> TensorOrTensors:
             with use_skip_tracker(skip_tracker), record_function("chunk%d-part%d" % (chunk_id, part_id)):
-                return partition(input)
+                ret = partition(input)
+                if type(ret) is list:
```
msbaines, line 184: I'm not sure about this check. It will be run for every partition for every micro-batch.

msbaines (commented):
Nice! I'm not sure about the warning though.

fairscale/nn/pipe/pipeline.py
min-xu-ai: Yeah, was worrying about that too. How about I turn it into a comment?

msbaines (approved):

PR #245, opened 2020-12-12
[fix] Sharded restore bucketing

blefaudeux:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -23,6 +23,18 @@ from fairscale.optim import OSS
 from fairscale.optim.utils import Workhandle
 
 
+class SyncLayer(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx: Any, sddp: "ShardedDataParallel", sentinel: torch.Tensor, *inputs) -> Any:  # type: ignore
+        ctx.sddp = sddp
+        return inputs
+
+    @staticmethod
+    def backward(ctx, *grad_inputs: Any):  # type: ignore
+        ctx.sddp._gatekeeper()
+        return tuple((None, None, *(g for g in grad_inputs)))
+
+
 class ShardedDataParallel(nn.Module):
     """
     Wrap the model, and reduce the gradients to the right rank during the backward pass.
@@ -96,6 +108,8 @@ class ShardedDataParallel(nn.Module):
         self._param_iterator = chain(*[optim.should_bucket_param.keys() for optim in self.sharded_optimizers])
         self._grad_to_be_reduced = [True for _ in self._param_iterator]
         self._grad_accs: List[Callable] = []
+        self._sync_layer = SyncLayer()
+
         self._setup_backward_hooks()
 
         # Make sure that all ranks start with the same model
@@ -114,6 +128,11 @@ class ShardedDataParallel(nn.Module):
 
         # Reset all the grad reduce and bucket state flags
         self._grad_to_be_reduced = [True] * len(self._grad_to_be_reduced)
+
+        # Register a layer in the autograd which will fire the gatekeeper during the backward pass
+        sentinel = torch.zeros((1))  # This makes sure that this function is properly registered
+        sentinel.requires_grad = True
+        inputs = self._sync_layer.apply(self, sentinel, *inputs)
 
         # Normal FW on the base model
         return self.module(*inputs, **kwargs)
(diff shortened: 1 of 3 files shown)


blefaudeux:
cc @SeanNaren @mrshenli @albanD, if you could have a look ? 

The gist of it is that we want to automatically reduce all gradients during the backward pass, and some of them are bucketed for performance reasons. Bucketing strategy is decided beforehand, smallest parameters are packed to be reduced in one go.
Per-parameter backward hooks are dropped, but if not all parameters are actually used, then some of these hooks will not fire. If any of the missing parameters were supposed to be in a bucket, this bucket does not flush and all hell breaks loose. 

The intent here is to be able to run a "gatekeeper" function at the end of the backward pass, which both unrolls the futures from the reduce calls (until we get a better mechanism) and possibly empties the partially filled buckets. I tried quite a few options for that, and would like to check with you guys that this is reasonable (I know that there's a C++ API for 'end of backward callback', but to the best of my knowledge it's not exposed in python).

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -167,6 +186,16 @@ def _find_rank(self, param: Parameter) -> Tuple[OSS, int]:
         assert False, "This parameter is not present in an optimizer, this should not happen"
         return (None, -1)
 
+    def _gatekeeper(self) -> None:
+        # Make sure that all the asynchronous calls have concluded before moving on. Consume the futures
+        # and execute the delayed actions (release gradients, unroll the buckets)
+        for optimizer in self.sharded_optimizers:
+            # Empty in-flight futures
+            Variable._execution_engine.queue_callback(optimizer._consume_work_handles)
```
albanD, line 194: Why do you queue callbacks here? The backward should already be finished for the part that you are interested in and so you should be able to call these functions directly no?

fairscale/optim/utils.py
```diff
@@ -130,6 +130,31 @@ def recursive_copy_to_device(value: Any, non_blocking: bool, device: torch.devic
     return value
 
 
+def recursive_tensor_clone(value: Any) -> Any:
```
albanD, line 133: nit: this is unused right?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -115,6 +129,11 @@ def forward(self, *inputs: Any, **kwargs: Any) -> Any:
         # Reset all the grad reduce and bucket state flags
         self._grad_to_be_reduced = [True] * len(self._grad_to_be_reduced)
 
+        # Register a layer in the autograd which will fire the gatekeeper during the backward pass
+        sentinel = torch.zeros((1))  # This makes sure that this function is properly registered
+        sentinel.requires_grad = True
+        inputs = self._sync_layer.apply(self, sentinel, *inputs)
```
albanD, line 135: This looks ok to me.
A couple of questions though:
- Is this module supposed to allow re-entrant autograd (calling backward within the backward, which `utils.checkpoint` is doing)?
- Is there a reason not to use a similar logic as DDP to detect un-used parameters? I am not saying their strategy is better, I am just wondering if there are key differences between the two?

blefaudeux: Thanks for the feedback ! Really a draft, testing out how/if that works, sorry for the unused code lost in the middle.
- re-entrant: I did not think of that to be honest, I don't think that it would work well by default, but I would need to understand more abut the expected behaviour in that case to think about remediations. When a second backward is called in the middle of a first one,  is that to get a second derivative or to recompute gradients which have been dropped from the first pass (given the checkpoint utilisation, guess is second hypothesis)
- re: DDP logic to find unused parameters, it's something we discussed with @mrshenli, I think that the added cost for every FW is a shame though, so I'm trying to make sure that there's no other solution before going down that route

albanD: re-entrant is to compute gradients for an independent graph while your compute this backward. That is how checkpoint uses it indeed.

What would be the difference between this in practice vs the one in DDP?

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: just saw that now (after my other reply), sorry for the async. Yes I think you're right, it's some copy/pasta of an old codebase, it does not seem to make sense in that case, I agree.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: (not sure whether this branch will live on, but the reentrant question is interesting) This moves the gradient to specific ranks instead of all-reducing (and clears the gradients if possible after the reducing is done), that's the only real difference with DDP unfortunately (longer term plan is to ditch any somethingDDP code here and fold back to pytorch DDP)

blefaudeux:
some ideas made through, some of that would not be relevant anymore with tensor views

PR #251, opened 2020-12-16
[chore] update to torch v1.7.1

msbaines:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ At a high level, we want ML researchers to:
 
 # Testing
 
-We use circleci to test on PyTorch versions 1.5.1, 1.6.0 and 1.7.0 and CUDA version 10.1. Please create an [issue](https://github.com/facebookresearch/fairscale/issues) if you are having trouble with installation.
+We use circleci to test on PyTorch versions 1.5.1, 1.6.0 and 1.7.1 and CUDA version 10.1. Please create an [issue](https://github.com/facebookresearch/fairscale/issues) if you are having trouble with installation.
 
 ## Contributors
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,12 +62,12 @@ install_dep_160: &install_dep_160
         python -m torch.utils.collect_env
 
 
-install_dep_170: &install_dep_170
+install_dep_171: &install_dep_171
   - run:
       name: Install Dependencies
       command: |
         sudo apt-get install -y libopenmpi-dev
-        pip install --progress-bar off  torch==1.7.0+cu101 torchvision==0.8.1+cu101 -f https://download.pytorch.org/whl/torch_stable.html
+        pip install --progress-bar off torch==1.7.1+cu101 torchvision==0.8.2+cu101 -f https://download.pytorch.org/whl/torch_stable.html
         pip install --progress-bar off -r requirements-test.txt
         pip install --progress-bar off git+https://github.com/msbaines/torch_pg.git@c85c96f#egg=torch-pg
         python -c 'import torch; print("Torch version:", torch.__version__)'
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -155,7 +155,7 @@ At a high level, we want ML researchers to:
 
 # Testing
 
-We use circleci to test on PyTorch versions 1.5.1, 1.6.0 and 1.7.0 and CUDA version 10.1. Please create an [issue](https://github.com/facebookresearch/fairscale/issues) if you are having trouble with installation.
+We use circleci to test on PyTorch versions 1.5.1, 1.6.0 and 1.7.1 and CUDA version 10.1. Please create an [issue](https://github.com/facebookresearch/fairscale/issues) if you are having trouble with installation.
```
min-xu-ai, line 158: I use cuda 11 at home. Do circleCI servers support it?

min-xu-ai (approved):
awesome

min-xu-ai:
One test was flaky?

msbaines:
> One test was flaky?

No looks like the wrong version of torch is getting installed for some reason:

Installing collected packages: torch
  Attempting uninstall: torch
    Found existing installation: torch 1.7.1+cu101
    Uninstalling torch-1.7.1+cu101:
      Successfully uninstalled torch-1.7.1+cu101
Successfully installed torch-1.7.0

README.md
stas00: it doesn't matter what circleCI has - pytorch comes with its own cudatoolkit, see: https://pytorch.org/get-started/locally/

fairscale at the moment chooses to test w/ cuda-10.1, but they could test with cuda-11.0 (and 11.2 soon), as pt-1.7.1 w/ cuda-11.0 is available too.

blefaudeux:
Master is failing because of that actually, pip install -r requirements.txt seems to have recently defaulted to installing 1.7, and a version incompatible with cuda 10.1, which subsequently breaks the benchmarks.
See [the dep install](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1069/workflows/3e694deb-ebab-46bd-89a6-0a8547a4588e/jobs/4476/parallel-runs/0/steps/0-108) and [the benchmark run](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1069/workflows/3e694deb-ebab-46bd-89a6-0a8547a4588e/jobs/4476/parallel-runs/0/steps/0-111)

stas00:
What might help is to pass the specific dir with the only cuda-10.1 choices, by adding to `pip install`:

`-f https://download.pytorch.org/whl/cu101/torch_stable.html`

README.md
anj-s: +1 on testing with cuda 11 since that seems to the default pip install torch version (#267). Is there a reason we want to only test with cuda 10.1?

min-xu-ai: I am guess that we test 10.1 because that's the default of FAIR cluster.

msbaines: We are limited by the driver version:

nvidia-smi
Fri Jan 15 23:51:12 2021       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 418.87.00    Driver Version: 418.87.00    CUDA Version: 10.1     |

This Nvidia Linux kernel driver installed on the circleci image we are using only support up to Cuda 10.1:

https://docs.nvidia.com/deploy/cuda-compatibility/index.html

We can install Cuda 11.0 but it is incompatible with the driver the Linux kernel is using.

We may be able to work around this by building our own docker image.

min-xu-ai: resolving this since it is not needed for this PR.

min-xu-ai: Agreed. I think we may need to use a different image in the future for that. Definitely not meant to request this change here. I was ignorant about what is needed when I commented on this.

image

msbaines:
> should we assert 1.7.1 on line 74 below? also, on CPU only environment, this will not install cuda version if I remember correctly. In any case, this is minor. I can try to address them in my PR for py3.9. If you have any comments, I'd love to know too.

Good idea. I can add a check for the patch level in a future PR.

Should be fine to install coda version of torch on cpu-only vm. Doesn't cause any harm and avoids a cpu-only config.

PR #252, opened 2020-12-16
[refactor] Remove unused variables, add configuration objects and basic cleanup for pipe benchmarks.

anj-s:
# Before submitting

## What does this PR do?
Refactors existing benchmark test for fairscale.nn.Pipe. Specifically:
- Move model and datasets out to their own files so that we can support benchmarking multiple models. (This is needed in preparation for the transformer benchmark. I was thinking of having common models/ and data/ folders where we can import different architectures needed for benchmarking OSS, Pipe, Adscale etc.)
- Remove duplicated configurations and consolidate all model specific configs in a dict and use that as a SSOT.
- Add flags to specify synthetic data and use this in place of `new_data` boolean.

## PR review

## Did you have fun?


benchmarks/datasets.py
```diff
@@ -0,0 +1,30 @@
+import warnings
```
min-xu-ai, line 1: add copyright header?

benchmarks/datasets.py
```diff
@@ -0,0 +1,30 @@
...
+        batch_size = 20
+        eval_batch_size = 10
+        train_data = batchify(train_txt, batch_size, TEXT, device)
+        val_data = batchify(val_txt, eval_batch_size, TEXT, device)
+        test_data = batchify(test_txt, eval_batch_size, TEXT, device)
+
+        return ntokens, train_data, val_data, test_data
+
+
+def batchify(data, bsz, TEXT, device):
```
min-xu-ai, line 25: add some docstring for these two functions will help new readers who are not familiar with NLP.

min-xu-ai: All CAP TEXT argument name looks strange to me.

benchmarks/models.py
```diff
@@ -0,0 +1,77 @@
+import math
```
min-xu-ai, line 1: same, header

min-xu-ai: the filename models.py seems to be too generic. Can you rename to it something more specific or do we want to put all model related classes in a single file?

benchmarks/models.py
```diff
@@ -0,0 +1,77 @@
...
+
+import torch
+import torch.nn as nn
+
+
+class EmbeddingLayer(nn.Embedding):
```
min-xu-ai, line 7: some docstring to document why do we wrap those underlying classes here will be useful.

benchmarks/models.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, ntoken, ninp, initrange):
+        super().__init__(ntoken, ninp)
+        self.ninp = ninp
+        self.weight.data.uniform_(-initrange, initrange)
+
+    def forward(self, src):
+        return super().forward(src) * math.sqrt(self.ninp)
+
+
+class PositionalEncodingLayer(nn.Module):
```
min-xu-ai, line 17: is there a pytorch lib we can use or we need to write our own positional embedding?

min-xu-ai (commented):
Thanks for doing this! Very nice improvements. Can I ask you to use this opportunity to improve the document of the code? This is first set of questions and comments. I will come back and do more review.

benchmarks/models.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class EmbeddingLayer(nn.Embedding):
+    def __init__(self, ntoken, ninp, initrange):
+        super().__init__(ntoken, ninp)
+        self.ninp = ninp
+        self.weight.data.uniform_(-initrange, initrange)
+
+    def forward(self, src):
+        return super().forward(src) * math.sqrt(self.ninp)
```
min-xu-ai, line 17: minor: there is no need to sqrt ninp everytime in the forward since ninp is not changing.

benchmarks/models.py
min-xu-ai: ignore that. I didn't find standard lib for that. perhaps just add a docstring here like:

""" This implements the positional encoding for original Transformer paper. """

benchmarks/models.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__(ninp, nhead, nhid, droupout)
+        self.src_mask = None
+
+    def _generate_square_subsequent_mask(self, sz):
+        mask = (torch.triu(torch.ones(sz, sz)) == 1).transpose(0, 1)
+        mask = mask.float().masked_fill(mask == 0, float("-inf")).masked_fill(mask == 1, float(0.0))
+        return mask
+
+    def forward(self, src):
+        if self.src_mask is None or self.src_mask.size(0) != len(src):
```
min-xu-ai, line 52: we should document what's src.shape. len(src) is src.size(0), which is batch dimension, right? I thought the mask should be generated on the sequence dimension, not the batch dimension?

also, regenerating the mask whenever len(src) changes can be expensive if len(src) changes all the time. It seems better to generate it once for biggest possible value and reuse a smaller portion of the tensor depending on len(src)?

benchmarks/datasets.py
anj-s: Done.

benchmarks/pipe.py
```diff
@@ -5,17 +5,17 @@
 import math
 import os
 import time
-import warnings
 
 from benchmark_dataset import BenchmarkLMDataset, collate_sentences_lm
+import datasets
+import models
+from models import TransformerLMSequntial
```
min-xu-ai, line 12: it is strange to import some class one way and others the other way. perhaps just import models and use models.<foo> would be better?

benchmarks/models.py
anj-s: Added a header.

I want to add a models and data folder where we can add specific models and datasets sort of like a mxn combination. This is happening in a followup PR where I will allow the benchmark to specify different models and datasets to run with the same training loop.

benchmarks/pipe.py
```diff
@@ -276,8 +158,7 @@ def dump_cuda_tensors():
     pprint.pprint(torch.cuda.memory_stats())
 
 
-def train(lm_dataloader, model, criterion, optimizer, vocab_size, args):
-    model.train()
+def log_number_of_parameters(model):
     from functools import reduce
     import operator
```
min-xu-ai, line 163: move those imports to the top of the file?

benchmarks/pipe.py
```diff
@@ -288,42 +169,67 @@ def train(lm_dataloader, model, criterion, optimizer, vocab_size, args):
...
+
+def get_fake_dataloader():
+    thing = {"input": torch.zeros(args.batch_size)}
+
+    class FakeDataset:
+        def __getitem__(self, index):
+            return thing
+
+        def __len__(self):
+            return len(lm_dataloader)
```
min-xu-ai, line 214: where is lm_dataloader from? make it a parameter passed to this class? better, make the length a param?

benchmarks/models.py
anj-s: We can use the [nn.transformer](https://pytorch.org/docs/master/nn.html#transformer-layers) which provides the entire model or specific encoder/decoder layers. There are no specific positional encoding layers provided. The [pytorch tutorial] (https://pytorch.org/tutorials/beginner/transformer_tutorial.html) seemed to have it's own PositionalEncodingLayer as well.

benchmarks/pipe.py
```diff
@@ -497,64 +401,70 @@ def generate_balance(num_devices, num_layers):
...
         return {
             "model": model,
-            "criterion": criterion,
             "optimizer": optimizer,
             "data": data,
         }
 
 
-def bench_single_process(args):
+def create_benchmark_config(model_name):
```
min-xu-ai, line 425: docstring, esp. the return type should be documented.

benchmarks/pipe.py
```diff
@@ -566,19 +476,18 @@ def run_mp_worker(args, available_workers):
         input_device=torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu"),
         pipelined_backward=args.pipelined_backward,
         checkpoint=args.checkpoint,
-        # loss_fn=blob["criterion"],
+        # FIXME: Do we need to comment this out? loss_fn=benchmark_config["criterion"],
```
min-xu-ai, line 479: it seems that the test need to have an option to test either this line is commented out or not?

benchmarks/pipe.py
```diff
@@ -681,14 +590,18 @@ def bench_mpi(args):
...
+parser.add_argument("--use_synthetic_data", default=True, help="Uses synthetic data to run benchmarks.")
+parser.add_argument(
+    "--model_name", default="seq_pred", choices=["seq_pred", "transformer"], help="Model used to benchmark pipe."
+)
 parser.set_defaults(pipelined_backward=True)
 
 if __name__ == "__main__":
     args = parser.parse_args()
-    # bench_multi_process(args, all_at_once=True)
+    # FIXME: Add support for multiprocess benchmarking.
```
min-xu-ai, line 601: use TODO (name) so that we can know the context of this TODO in the future.

benchmarks/datasets.py
anj-s: Done. Removed the CAP TEXT args, not sure why they were in CAPs.

benchmarks/pipe.py
```diff
@@ -497,64 +401,70 @@ def generate_balance(num_devices, num_layers):
     return balance
 
 
-def make_model_and_data(args, device, new_data: bool = True):
+def make_model_and_data(args, new_data: bool = True, config=None):
```
min-xu-ai, line 404: docstring

benchmarks/pipe.py
```diff
@@ -335,20 +241,12 @@ def get_last_device(model):
             find_unused_parameters=False,
         )
 
+    # FIXME: Why do we need this? Can we use synthetic data instead?
```
min-xu-ai, line 244: I think this is needed for multiprocess pipe so that fake data is sent to the model replicas except the first and the last replicas. I think a comment here will be helpful. A TODO perhaps can be added to figure out how to avoid the need of this in the future in the API level.

min-xu-ai (commented):
more comments. thanks for the attention to details in the current diff. I can see that.

benchmarks/models.py
min-xu-ai: OK to leave it in as our own implementation. It is not very complex anyway. Good comments will be nice to have.

benchmarks/models.py
min-xu-ai: I think it is added to make initialization of the weights easier. I think you can just add that in the docstring here.

benchmarks/models.py
min-xu-ai: sounds good. rename later is totally fine.

min-xu-ai:
btw, I use this alias to do code formatting. it makes life easier:

```
$ type fs_test
fs_test is aliased to `black . && isort . && flake8 && mypy --ignore-missing-imports --scripts-are-modules --pretty . && echo ALL GOOD'
```

benchmarks/models.py
```diff
@@ -0,0 +1,80 @@
...
+            device = src.device
+            mask = self._generate_square_subsequent_mask(len(src)).to(device)
+            self.src_mask = mask
+
+        return super().forward(src, self.src_mask)
+
+
+class LinearLayer(nn.Linear):
```
min-xu-ai, line 60: docstring to say that we wrap it for initialization purpose.

benchmarks/pipe.py
```diff
@@ -288,42 +169,67 @@ def train(lm_dataloader, model, criterion, optimizer, vocab_size, args):
...
-        if isinstance(model, DDP):
-            model = model.module
 
-        if not torch.cuda.is_available():
-            return torch.device("cpu")
-        if model.devices:
-            return model.devices[-1]
-        else:
-            return torch.cuda.current_device()
+def get_last_device(model):
```
min-xu-ai, line 194: can merge with the last function into get_first_or_last_device to avoid some duplication. Only one line is different between the two anyway.

benchmarks/models.py
anj-s: Done, added the comment.

Is there a reason we prefer normalized weights in a given range as opposed to the default weight initialization? Should we be recommending users to follow this?

min-xu-ai: I don't know. I think it is unlike to matter much in this benchmarking test.

benchmarks/models.py
```diff
@@ -49,6 +51,7 @@ def _generate_square_subsequent_mask(self, sz):
         return mask
 
     def forward(self, src):
+        # TODO(anj-s): Fix the data format so that we have [seq_len, batch_size, embedding dim].
```
min-xu-ai, line 54: it is unclear from this comment what's the current format of src. did you run it and inspect the shape of src?

benchmarks/pipe.py
```diff
@@ -497,64 +387,74 @@ def generate_balance(num_devices, num_layers):
...
             "optimizer": optimizer,
             "data": data,
         }
 
 
-def bench_single_process(args):
+def create_benchmark_config(model_name):
+    """Return a dict with configurations required for benchmarking `model_name` model."""
+
+    if model_name == "seq_pred":
```
min-xu-ai, line 414: should have an "else" clause?

benchmarks/pipe.py
```diff
@@ -681,14 +580,18 @@ def bench_mpi(args):
 parser.add_argument(
     "--no-pipelined-backward", dest="pipelined_backward", action="store_false", help="Pipelined backward pass"
 )
+parser.add_argument("--use_synthetic_data", default=True, help="Uses synthetic data to run benchmarks.")
```
min-xu-ai, line 583: the help msg doesn't seem to match the actual semantics in the code?

benchmarks/models.py
anj-s: Done.

benchmarks/models.py
anj-s: Added a comment to investigate more. If we don't need the weight initialization then we can use the layer directly.

benchmarks/models.py
anj-s: Good point. I think this benchmark code follows this [tutoria](https://pytorch.org/tutorials/beginner/transformer_tutorial.html#load-and-batch-data) very closely where the data is supposed to be batched such that seq_len is the first dimension and batch_size is the second dimension. However I think there is an issue with the way we are batching now such that the first dimension is the batch size. Note, that this batching is correct if we use real data and not synthetic so there is some discrepancy between the two input pipelines.

I am going to refactor the input pipeline such that both the real and synthetic data pipelines use similar intermediate steps.

benchmarks/models.py
anj-s: Added an extra comment to clarify. Yes, I did run it to confirm that we have batch size in the first dimension for synthetic data.

benchmarks/pipe.py
anj-s: Yes, I think so. I did add that in a followup PR that I will send out.

benchmarks/pipe.py
anj-s: I was thinking of the eventual state :) I modified it to reflect that we will use synthetic data for only a sample training run.

benchmarks/pipe.py
anj-s: Done.

benchmarks/models.py
anj-s: Done.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
anj-s: I am not sure at the moment. I am going to tackle this when I enable the benchmark again.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
anj-s: Done.

anj-s:
> btw, I use this alias to do code formatting. it makes life easier:
> 
> ```
> $ type fs_test
> fs_test is aliased to `black . && isort . && flake8 && mypy --ignore-missing-imports --scripts-are-modules --pretty . && echo ALL GOOD'
> ```
Thanks Min! That is super useful. I spent some time figuring this out in my first PR. I ended up with the following alias command `isort <dir> && flake8 <dir> && black <dir>` which is missing some of the other good stuff you have :)

benchmarks/models.py
```diff
@@ -6,18 +6,21 @@
...
 
 
-# FIXME: Add docstring explaining why we wrap the Embedding layer.
 class EmbeddingLayer(nn.Embedding):
+    """Wrapped nn.Embedding layer to allow for weight initialization."""
+
     def __init__(self, ntoken, ninp, initrange):
         super().__init__(ntoken, ninp)
-        self.ninp = ninp
+        self.ninp = math.sqrt(ninp)
```
min-xu-ai, line 15: it is better to rename var to ninp_sqrt

min-xu-ai (approved):
LGTM! Good luck merging. One nitpicking comment above about var name. I think you may want to file an issue on the batch vs. seq dimension issue you have identified here.

blefaudeux:
Just to be sure I'm not missing something here @anj-s @min-xu-ai, who's merging this ? I'm guessing it should not stay up in the air for too long :)

min-xu-ai:
> Just to be sure I'm not missing something here @anj-s @min-xu-ai, who's merging this ? I'm guessing it should not stay up in the air for too long :)

I don't plan to merge this @anj-s's PR. I only merge others PR if (1) no pending changes needed & all approved and (2) it is small and (3) author don't have merge permission and (4) I need to follow-up the work.

blefaudeux:
> > Just to be sure I'm not missing something here @anj-s @min-xu-ai, who's merging this ? I'm guessing it should not stay up in the air for too long :)
> 
> I don't plan to merge this @anj-s's PR. I only merge others PR if (1) no pending changes needed & all approved and (2) it is small and (3) author don't have merge permission and (4) I need to follow-up the work.

I was not sure about the permissions actually, but fine by me then !

anj-s:
> > > Just to be sure I'm not missing something here @anj-s @min-xu-ai, who's merging this ? I'm guessing it should not stay up in the air for too long :)
> > 
> > 
> > I don't plan to merge this @anj-s's PR. I only merge others PR if (1) no pending changes needed & all approved and (2) it is small and (3) author don't have merge permission and (4) I need to follow-up the work.
> 
> I was not sure about the permissions actually, but fine by me then !

Thanks for checking @blefaudeux! Just merged the PR without issues. Thanks @min-xu-ai for the reviews.

PR #259, opened 2020-12-17
[OSS] Getting rid of the "should bucket" hash table, just use a list + robustness related fixes

### LINKED ISSUE #258: [OSS] Params without requires_grad can slip out of the bucketing table, and break
## 🐛 Bug
Root bug behind https://github.com/huggingface/transformers/issues/9156, which explains why a "new" parameter was showing up. The hash table is probably slow and not really needed any way, I should have stuck to a list as recommended by @msbaines long ago

## Expected behavior
Should never break 

cc @stas00

blefaudeux:
Properly handle all params, with or without requires_grad, HuggingFace has triggered a bug which was always there. Remove the hash table to store the strategy (a bit risky if the tensors are moved around) and replace with a simple list. Make sure that we update the strategy if a new parameter group is added

# Before submitting

## What does this PR do?
Fixes https://github.com/huggingface/transformers/issues/9156 for the multi-node case, ie: with bucketing involved, tested locally by commenting out the no-bucketing statement.
Fixes #258 
Makes sure that this case is caught by the unit test, I'm really surprised that this was not covered before, but I did check that master fails on this

## PR review

## Did you have fun?
Well, yes 🙃 the fp16 issue is not solved though

cc @stas00

### DIFF
--- a/tests/nn/data_parallel/test_sharded_ddp.py
+++ b/tests/nn/data_parallel/test_sharded_ddp.py
@@ -44,6 +44,7 @@ def run_one_step(rank, world_size, backend, device, temp_file_name):
         model.register_buffer("test_buffer", torch.ones((1)) * rank)
         model.to(device)
 
+        next(model.parameters()).requires_grad = False
         optimizer = OSS(params=model.parameters(), optim=torch.optim.SGD, lr=0.01, momentum=0.99)
         ddp_model = ShardedDataParallel(model, optimizer, broadcast_buffers=broadcast_buffers)
 
--- a/tests/nn/data_parallel/test_sharded_ddp.py
+++ b/tests/nn/data_parallel/test_sharded_ddp.py
@@ -7,6 +7,7 @@
 Testing OssDdp class.
 """
 
+from contextlib import suppress
 import copy
 import tempfile
 from typing import List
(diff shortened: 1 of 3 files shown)


stas00:
Thank you very much for the fix, @blefaudeux!

blefaudeux:
ping reviews, if you don't mind, I would love master to work for HuggingFace

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -44,6 +44,7 @@ def check(broadcast_buffers: bool, grad_accumulation: bool = False) -> None:
         model.register_buffer("test_buffer", torch.ones((1)) * rank)
         model.to(device)
 
+        next(model.parameters()).requires_grad = False
```
min-xu-ai, line 47: add a comment?

min-xu-ai (approved):
Looks sane to me. Again, it seems the changes are centered around having params that doesn't have require_grads. Perhaps more explicit test cases would help prevent regression?

tests/nn/data_parallel/test_sharded_ddp.py
blefaudeux: Yes, I was about to right that this is the unit test change which makes sure that this is caught in the future. Will do !

PR #260, opened 2020-12-17
[refactor] Modify train and benchmark functions to account for multiple models and datasets.

anj-s:
# Before submitting

## What does this PR do?
- Refactor train() to allow us to run different models
- Refactor benchmark_language_model to run different models with different datasets
- Rename functions and move constants to config dicts
- Modify benchmark function to only focus on train performance. The previous version hardcoded validation and testing results so in essence only train was being benchmarked. 

## PR review

## Did you have fun?


benchmarks/pipe.py
```diff
@@ -43,7 +44,14 @@ def init_random_seed(seed: int):
     numpy.random.seed(seed)
 
 
-def make_model(args, device, config):
+def get_model_and_optimizer(args, device, config):
```
min-xu-ai, line 47: docstring

benchmarks/pipe.py
```diff
@@ -43,7 +44,14 @@ def init_random_seed(seed: int):
...
+
+    if args.model_name == "seq_pred":
+        return get_seq_pred_model(args, device, config)
+
+
+def get_seq_pred_model(args, device, config):
```
min-xu-ai, line 53: docstring.

this model is usually called LM (language model). perhaps rename the string to "lm" or "language_model" and rename the function get_language_model().

Also, we can use python enum instead of string in the place of "seq_pred".

benchmarks/pipe.py
```diff
@@ -245,10 +253,12 @@ def train(data_config, model, benchmark_config, args):
     if pipe_group and pipe_group.rank() != 0 and pipe_group.rank() != (pipe_group.size() - 1):
         lm_dataloader = get_fake_dataloader()
 
+    total_words = 0
     for i, batch in enumerate(lm_dataloader):
 
         if args.max_batch and i > args.max_batch:
             break
+        total_words += np.prod(list(batch["input"].size()))
```
min-xu-ai, line 261: batch["input"].numel() returns the number of elements, which is that what this is computing I think.

For total_words, we usually compute it exclude the padding tokens. here, I don't know the data contains padding or not. either way, using total_tokens might be more accurate.

benchmarks/pipe.py
```diff
@@ -284,7 +294,7 @@ def train(data_config, model, benchmark_config, args):
         if pipe_group is None or pipe_group.rank() == pipe_group.size() - 1:
             total_loss += loss.item()
             log_interval = 1
-            word_counter += batch["ntokens"]
+            word_counter += np.prod(list(batch["input"].size()))
```
min-xu-ai, line 297: again, I think batch["input"].numel() should do.

total_words and word_counter are really similar. if they are different counters, I think we should rename them to make sure their names convey the difference.

benchmarks/pipe.py
```diff
@@ -297,10 +307,13 @@ def train(data_config, model, benchmark_config, args):
...
                 start_time = time.time()
 
+    return total_words, loss.item()
+
 
-def evaluate(eval_model, data_source, criterion, bptt, ntokens):
+def evaluate(eval_model, data_source, criterion, ntokens):
     eval_model.eval()
     total_loss = 0.0
+    bptt = 35  # Used for evaluation not training.
```
min-xu-ai, line 316: the comment seems to be obvious. It would be useful to explain why is it 35, how is that chosen if we know it.

benchmarks/pipe.py
```diff
@@ -322,58 +335,48 @@ def get_number_of_words(data):
     return data.size()[0] * data.size()[1]
 
 
+def verify_seq_pred_run(wps):
```
min-xu-ai, line 338: nice. it is good to have a verification function. Thanks!

benchmarks/pipe.py
```diff
@@ -322,58 +335,48 @@ def get_number_of_words(data):
...
+    # Assert that words per second is within 3 standard deviations of the average
+    # of six golden runs
+    assert wps > 36954.4 - (3 * 116.825)
+
+    print("Peak allocated bytes on cuda:0: {:1d}".format(torch.cuda.memory_stats(0)["allocated_bytes.all.peak"]))
```
min-xu-ai, line 343: it is better to turn this in a loop

benchmarks/pipe.py
```diff
@@ -322,58 +335,48 @@ def get_number_of_words(data):
...
+    print("Peak allocated bytes on cuda:1: {:1d}".format(torch.cuda.memory_stats(1)["allocated_bytes.all.peak"]))
+    print("Peak allocated bytes on cuda:2: {:1d}".format(torch.cuda.memory_stats(2)["allocated_bytes.all.peak"]))
+    print("Peak allocated bytes on cuda:3: {:1d}".format(torch.cuda.memory_stats(3)["allocated_bytes.all.peak"]))
+
+    # Assert that memory usage on each GPU is within 10% of golden run
+    # Right-hand-side is golden run bytes * 110%
+    assert torch.cuda.memory_stats(0)["allocated_bytes.all.peak"] < 4061909504 * 1.1
```
min-xu-ai, line 350: again, turn it into a loop like:

indexes = [...]
refs = [...]
for idx, ref in zip(indexes, refs):
   assert ..(idx)[..] < ref * 1.1

benchmarks/pipe.py
```diff
@@ -322,58 +335,48 @@ def get_number_of_words(data):
...
+    assert torch.cuda.memory_stats(1)["allocated_bytes.all.peak"] < 4050944 * 1.1
+    assert torch.cuda.memory_stats(2)["allocated_bytes.all.peak"] < 10427392 * 1.1
+    assert torch.cuda.memory_stats(3)["allocated_bytes.all.peak"] < 2031824896 * 1.1
+    print("No regression detected")
```
min-xu-ai, line 354: this seems redundant.

benchmarks/pipe.py
```diff
@@ -444,30 +467,30 @@ def benchmark_single_process(args):
     init_random_seed(0)
 
     benchmark_config = create_benchmark_config(args.model_name)
-    model_config = make_model_and_data(args, config=benchmark_config)
+    model_config = create_model_config(args, config=benchmark_config)
     model = model_config["model"]
 
     balance = generate_balance(min(num_devices, 4), len(model))
-    p = pipe.Pipe(
+    pipe_model = pipe.Pipe(
```
min-xu-ai, line 474: thanks for renaming!

benchmarks/pipe.py
```diff
@@ -444,30 +467,30 @@ def benchmark_single_process(args):
...
     )
     del model
     del model_config["model"]
 
     if args.use_synthetic_data:
-        train(model_config, p, benchmark_config, args)
+        train(model_config, pipe_model, benchmark_config, args)
     else:
-        benchmark_language_model(model_config, p, benchmark_config, args)
+        benchmark_language_model(model_config, pipe_model, benchmark_config, args)
```
min-xu-ai, line 483: How about add a comment like:

`# args.use_synthetic_data implies that we don't need to benchmark. Hence, we call train() directly without the benchmark_language_model wrapper.`

Since the functions are called the same way, it is better to reduce duplicate a bit:

func = train
if not args.use_synthetic_data:
    func = benchmark_language_model
func(...)

min-xu-ai (commented):
Pretty nice. Some comments.

min-xu-ai:
Please ping me to review this again once you are ready.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
anj-s: I want to have users to be able to benchmark synthetic data as well. The flow that I am thinking of is the following:
- introduce a dry_run flag where we don't compare values and instead just train the model
- use_synthetic_data will switch between using real vs synthetic data as the name suggests.

benchmark_single_process should look something like
model, optimizer = get_model_and_optimizer
dataloader = get_dataloader
benchmark_config = get_benchmark_config

benchmark_model(model, optimizer, dataloader, benchmark_config)

benchmarks/pipe.py
anj-s: I modified the name since we will have a second benchmark which is also a language model. The two names that I had in mind were "seq_pred" (existing model) and "seq_trans" (new model). Language model seemed too generic given the two benchmarks we want to add. Open to other naming suggestions if the previously mentioned ones don't feel right.

I am not convinced about if we should add an enum. Implementing a flag based enum will be a bit of boilerplate code. LMK if you feel strongly and I can add the implementation.

min-xu-ai: Usually, seq_pred is called LM, seq_trans is called NMT. Language modeling (not language model) is a special kind of task. So, I would recommend LM and NMT.

String is fine, since you have argparse's choice to select the right string anyway.

anj-s: Done.

benchmarks/pipe.py
anj-s: I am not aware of why we chose it. We don't benchmark evaluation right now. I've added a comment to followup though.

benchmarks/pipe.py
anj-s: Done.

benchmarks/pipe.py
min-xu-ai: sounds good!

benchmarks/pipe.py
```diff
@@ -577,7 +607,10 @@ def bench_mpi(args):
 )
 parser.add_argument("--use_synthetic_data", default=True, help="Uses synthetic data for a sample training run.")
 parser.add_argument(
-    "--model_name", default="seq_pred", choices=["seq_pred", "transformer"], help="Model used to benchmark pipe."
+    # TODO(anj-s): In the process of adding more models and hence the requirement for a flag.
+    "--model_name",
+    default="lm",
```
min-xu-ai, line 612: should this be a `choice` type for argparse?

min-xu-ai (approved):
Looks great! Just a minor comment that can be done in the next commit.

benchmarks/pipe.py
anj-s: Yes, I was planning to add the `choices` parameter when adding support for running a second model.

PR #263, opened 2020-12-17
Resolve issue where grad norms because inf with OSS / fp16

joshim5:
Discussed with @blefaudeux over chat. For large transformer models where grad norm is high (~5000 per device), the OSS clip norm code causes clip norm to become inf. This is because the logic is done in fp16 and 5000 ** 2 causes overflow in fp16.

The solution is to cast everything to fp32 before squaring/reducing. The return norm is also in fp32, as a result.


blefaudeux (approved):
LGTM, many thanks for the fix and the time spent on root finding @joshim5 !

PR #272, opened 2020-12-28
[chore] Move all unit tests dist init to being file based

blefaudeux:
# Before submitting

## What does this PR do?
Needed for T82199332 (FB internal), concurrent execution of the unit tests. Being file based, the torch distributed init does not collide anymore with existing instances. It also make it easier to upstream this code, since bigger projects would probably follow that rule also

## PR review

## Did you have fun?


fairscale/utils/testing.py
```diff
@@ -81,35 +82,34 @@ def torch_version() -> Tuple[int, ...]:
     return tuple(int(n) for n in numbering)
 
 
-def dist_init(rank: int, world_size: int, hostname: Optional[str] = None) -> None:
-    if hostname is None:
-        hostname = "localhost"
-    print(f"dist init r={rank}, world={world_size}, host={hostname}")
-    os.environ["MASTER_ADDR"] = hostname
-    os.environ["MASTER_PORT"] = "10638"
+def dist_init(rank: int, world_size: int, filename: str) -> None:
```
min-xu-ai, line 85: add a comment on why we are using file based init? Also, the limitation of file based init is that the test can only run within a node, but that should be fine since all of the tests are within a node, right?

min-xu-ai (approved):
very nice, I like it.

fairscale/utils/testing.py
blefaudeux: ah yes, two good points ! we're only testing within a node indeed, same for FB infra, I think it makes some sense for now

PR #276, opened 2020-12-30
[fix] Dead code removal for OSS

blefaudeux:
# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [ ] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

## What does this PR do?
Removes some dead code which was only useful for DDP, and must have some footprint for a big model (traverse the graph).
Opportunistically update the changelog, forgotten over many diffs

## PR review
Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #281, opened 2020-12-30
[fix] Regression testing oss+sharded_ddp only

blefaudeux:
# Before submitting

## What does this PR do?
The regression detection, which triggers following https://github.com/facebookresearch/fairscale/pull/276. The gist of it is that with the current config, oss+DDP is part of the regression test, while we're really interested in OSS+ShardedDDP (and the threshold is much lower for this one actually, so the current setting was not very good).

3 changes then:
- regression test oss+sharded_ddp only
- much tighter threshold (900MB vs 1100MB) which actually reflects this solution's real perf
- keep pytorch+oss/ddp+oss/sharded_dpp as free runs, I found it useful to be able to compare the 3 solutions side by side 

## PR review

Sorry for the noise around that today, hopefully more stable now

## Did you have fun?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,8 @@ run_oss_benchmark: &run_oss_benchmark
   - run:
       name: Run OSS Benchmark
       command: |
-        python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 660 --reference_memory 1120 --reference_loss 0.023
+        python benchmarks/oss.py --world_size 4 --epochs 2
+        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp --reference_speed 660 --reference_memory 930 --reference_loss 0.023
 
 run_oss_gloo: &run_oss_gloo
   - run:


min-xu-ai (approved):
sounds good to me. this fixes the master branch?

blefaudeux:
> sounds good to me. this fixes the master branch?

yes, it's something else than the unit test issue, it's tied to the dead code removal from this afternoon which penalized slightly the hybrid use case "zero + ddp". I think that this PR improves on the situation because it tightens a lot the requirements for the situation which really counts, oss + shardedDDP, so should be for the better

.circleci/config.yml
```diff
@@ -135,7 +135,8 @@ run_oss_benchmark: &run_oss_benchmark
   - run:
       name: Run OSS Benchmark
       command: |
-        python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 660 --reference_memory 1120 --reference_loss 0.023
+        python benchmarks/oss.py --world_size 4 --epochs 2
+        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp --reference_speed 660 --reference_memory 930 --reference_loss 0.023
```
blefaudeux, line 139: same loss, just a tighter reference memory, 930MB vs. 1120 for the same config. The idea is that we're testing OSS+ShardedDDP, which has a lower memory consumption actually, this was not tested properly. We were checking for the worst case, which degraded slightly this afternoon

.circleci/config.yml
```diff
@@ -135,7 +135,8 @@ run_oss_benchmark: &run_oss_benchmark
   - run:
       name: Run OSS Benchmark
       command: |
-        python benchmarks/oss.py --check_regression --world_size 4 --reference_speed 660 --reference_memory 1120 --reference_loss 0.023
+        python benchmarks/oss.py --world_size 4 --epochs 2
```
anj-s, line 138: Do we need to run this if we don't check for regressions?

PR #296, opened 2021-01-07
[perf][minor] ShardedDDP micro-optim

blefaudeux:
# Before submitting

## What does this PR do?
replaces a division by a in-place multiplication, with a pre-computed factor. Does not change a lot in practice, probably picked up by some compiler optimization, but my guess is that it's better to have it right in the first place. A follow up could be to check with Fairseq if their fused reduce/scale could be reused. 

## PR review

## Did you have fun?


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         # Communication related attributes
         self.process_group = process_group if process_group is not None else dist.group.WORLD
         self.world_size = dist.get_world_size(self.process_group)
+        self.world_scaling = 1.0 / self.world_size  # > 0
```
anj-s, line 72: nit: if self.world_size isn't used anywhere else (only used here AFAICT), should we inline the previous line here?

anj-s: nit: perhaps world_size_scaling instead of world_scaling?

anj-s (approved):

PR #299, opened 2021-01-08
[doc] ShardedDDP minor additions

blefaudeux:
# Before submitting

## What does this PR do?
Following a couple of discussions (for instance in https://github.com/pytorch/pytorch/pull/46750), adding better explanations and suggestions directly in ShardedDDP docstrings

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -23,8 +23,7 @@ from fairscale.optim.utils import Workhandle
 
 
 class ShardedDataParallel(nn.Module):
-    """
-    Wrap the model, and reduce the gradients to the right rank during the backward pass.
+    """ Wrap the model, and reduce the gradients to the right rank during the backward pass.
 
     - the partition is given by the sharded optimizer
     - wrap the base model with a model which knows where to reduce each gradient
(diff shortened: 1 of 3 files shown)


fairscale/optim/adascale.py
```diff
@@ -554,10 +554,11 @@ def set_num_gradients_to_accumulate(self, num_gradients_to_accumulate: int, upda
         `set_scale` needs to be called to update the scale as well.
 
         TODO (min): need a way of determine how much to increase the step size?
+
         TODO (min): have both `set_scale` and `set_num_gradients_to_accumulate`
-                    is hard to use and easy to make mistake. I think it is better
-                    to specific a specify a `base_scale`. But more discussion is
-                    needed here.
+        is hard to use and easy to make mistake. I think it is better
```
blefaudeux, line 559: sphinx was complaining about that, just a warning removal

fairscale/optim/oss.py
```diff
@@ -239,7 +239,6 @@ def clip_grad_norm(
 
         .. warning: This needs to be called on all ranks, since synchronization primitives will be used
 
-        .. warning: Model paralelism -groups other than world- are not yet supported
```
blefaudeux, line 242: not true anymore

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -46,6 +45,21 @@ class ShardedDataParallel(nn.Module):
             Synchronize the models in between the ranks when starting up. Not needed if each rank has the same seed,
             or the training restarts from a saved state
 
+
+    .. warning:
+        ShardedDDP implements gradient sharding, meaning that each rank only owns a unique shard of the model gradients
+        after the backward pass, in order to save memory and some communication bandwidth.
+
+    .. warning:
+        As a consequence of sharding, in case of gradient clipping, one has to use the `clip_grad_norm` exposed by
```
anj-s, line 54: If users miss this docstring, will they hit an error message? How can we communicate this and the below hanging behavior to users? my fear is that they may miss these warnings and end up with a hanging run which are hard to debug.

blefaudeux: it's hard to catch because this is typically done outside of OSS/ShardedDDP, what happens is that the grad scaler will check the gradients (on each rank), and will bail out if there are NaNs/inf. If it's not the case on all ranks, then some of them bail out of the step and others repeat the step, they're not in sync and the communication primitives (broadcast) hang and time out. 
I'm not sure how to improve on that, some frameworks (lightning and huggingface for instance) help with their fairscale integration because they handle all of that and the user does not have to know

anj-s (approved):

fairscale/nn/data_parallel/sharded_ddp.py
anj-s: Thanks for the context! We will have to instrument timeouts on collective operations to handle hanging issues such as these. Definitely tricky to handle all the cases though since we don't have visibility.

PR #297, opened 2021-01-08
[refactor][OSS] Removing ad-hoc object broadcast, use pytorch's

blefaudeux:
# Before submitting
- [ ] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [ ] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

## What does this PR do?
Follow up from, use a torch helper to broadcast objects instead of a ad-hoc implementation https://github.com/pytorch/pytorch/pull/46750#pullrequestreview-563929958

## PR review
Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.
Thanks  @pritamdamania87 for the pointer !

## Did you have fun?
Make sure you had fun coding 🙃

PR #300, opened 2021-01-08
[perf][OSS] tensor views for bucketing

blefaudeux:
# Before submitting

## What does this PR do?
Speedups by a couple of %, removes a lot of code, saves some memory (buckets are free)

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -53,6 +53,8 @@ class OSS(Optimizer):
             torch.distributed group (default: group.WORLD)
         broadcast_buffer_size (int):
             the max size of the buffer used to batch the small parameter tensors, in number of elements (default 16M).
+            this will not impact the long term memory consumption, but the peak memory can be impacted by the moment
+            when the buffers are allocated and the bucketed params have not yet been relocated to them.
     """
 
     #: The optimizer used for a given shard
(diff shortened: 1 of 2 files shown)


blefaudeux:
@msbaines I'm not sure how that works for torch1.5 and 1.6, I don't think that we're actually testing it nowadays (since the pip default change)

blefaudeux:
> @msbaines I'm not sure how that works for torch1.5 and 1.6, I don't think that we're actually testing it nowadays (since the pip default change)

I've just tested, this part (tensor views) is fine, but a previous PR (https://github.com/facebookresearch/fairscale/pull/297) was not in fact pytorch 1.5 compatible (cc @pritamdamania87), and CI did not catch that..

fairscale/optim/oss.py
```diff
@@ -52,7 +52,7 @@ class OSS(Optimizer):
         group (group):
             torch.distributed group (default: group.WORLD)
         broadcast_buffer_size (int):
-            the size of the buffer used to batch the small parameter tensors (default 128k).
+            the max size of the buffer used to batch the small parameter tensors, in number of elements (default 16M).
```
min-xu-ai, line 55: talk a bit about the impact on per-GPU memory usage here?

min-xu-ai (approved):
Really like the simplification and performance/memory gain!

fairscale/optim/oss.py
blefaudeux: the beauty of this PR is that the impact on GPU usage is now.. 0 (zero) :) the small parameters are just a view of a bigger buffer, so the bucketing takes no extra space (on top of taking no extra cpu cycle).
I just realized that I should resize it to match the last bucketed parameter though, right now the remainder is unused and do increase the gpu memory a tiny bit

blefaudeux: hmm, the test jobs do show a slightly larger memory consumption, looking into that

blefaudeux: ok, makes sense given the initial extra memory allocation, before the parameters have been folded to the buffer, I've updated the docstring to reflect that

PR #303, opened 2021-01-09
[test] cover python 3.7 to 3.9 on CPU

min-xu-ai:
- covering common python versions on CPU

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -282,8 +282,10 @@ jobs:
       - <<: *run_black
       - <<: *run_mypy
       - <<: *run_flake8
-      - <<: *run_unittests
-      - <<: *run_mpi_unittests
+      # FIXME: py39 still not stable for us, example:
+      # https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1349/workflows/534aae41-e01d-404e-bfc1-fdc58566c39c/jobs/5952
+      # - <<: *run_unittests
+      # - <<: *run_mpi_unittests
       - <<: *run_doc_build
 
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,9 +10,19 @@ version: 2
 # -------------------------------------------------------------------------------------
 # Environments to run the jobs in
 # -------------------------------------------------------------------------------------
-cpu: &cpu
+cpu_py37: &cpu_py37
   docker:
     - image: circleci/python:3.7
+  resource_class: medium
+
+cpu_py38: &cpu_py38
+  docker:
+    - image: circleci/python:3.8
+  resource_class: medium
+
+cpu_py39: &cpu_py39
+  docker:
+    - image: circleci/python:3.9
   resource_class: medium
 
 gpu: &gpu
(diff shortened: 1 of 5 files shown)


min-xu-ai:
python 3.9 seems to have trouble running the unit tests. The problem seems to be random segv after a subprocess finishes. The tests seems to run fine but process exit logic seems to be broken. I think for now we can at least make sure linting all works on py3.9 and re-enable the tests when py3.9 is more stable.

blefaudeux:
I've another branch up to restore torch 1.5 and 1.6 testing (surprise, a lot of things were broken since we were not testing.. ). Would you mind waiting for that to be done before rebasing this one  @min-xu-ai  ? cc @msbaines

.circleci/config.yml
```diff
@@ -70,6 +80,17 @@ install_dep_17_cpu: &install_dep_17_cpu
         python -c 'import torch; print("Torch version:", torch.__version__)'
         python -m torch.utils.collect_env
 
+install_dep_17_cpu_cu11: &install_dep_17_cpu_cu11
+  - run:
+      name: Install Dependencies
+      command: |
+        sudo apt-get install -y libopenmpi-dev
+        pip install --progress-bar off torch==1.7.1+cu110 -f https://download.pytorch.org/whl/torch_stable.html
```
blefaudeux, line 88: I'm not seeing where cuda 11 is actually requested on the agent, bar that this will run on cpu only I think

blefaudeux: given the other PR (thanks for waiting and sorry for the merge conflicts !) I think you'll probably get something good with just installing requirements if you want 1.7 on cpu, no need to install a specific version, at least that's my guess

min-xu-ai: No problem. Thanks for fixing the other issue. That will make this PR even better. I will handle the conflicts.

Yeah for cpu, we can just install the version without cuda. I will test it out.

msbaines: To support CUDA 11.0, we would need to create a docker image with the correct kernel driver.

min-xu-ai:
Hi @msbaines @blefaudeux, I fixed up this PR. We now have two more things:

- we cover py 3.7 to 3.9, in terms of static checking for py39. Py39 is not yet stable to run all unit tests.
- we also cover doc build so that doc problems can be detected earlier in the PRs.

What do you guys think?

.circleci/config.yml
```diff
@@ -135,20 +165,30 @@ run_oss_gloo: &run_oss_gloo
...
-    name: Run OSS with Torch AMP
-    command: |
-      python benchmarks/oss.py --amp --epochs 3 --optim_type oss_sharded_ddp
+   - run:
+       name: Run OSS with Torch AMP
+       command: |
+         python benchmarks/oss.py --amp --epochs 3 --optim_type oss_sharded_ddp
+
+
+run_doc_build: &run_doc_build
```
blefaudeux, line 175: nice !

requirements-test.txt
```diff
@@ -1,15 +1,16 @@
+# Tools for static checking.
 black == 19.10b0
 flake8 == 3.7.9
 isort == 5.6.4
-mpi4py == 3.0.3
-mypy == 0.770
+mypy == 0.790
```
blefaudeux, line 5: does that change anything, practically ?

blefaudeux (approved):
lgtm, thanks for this work Min !

requirements-test.txt
min-xu-ai: only mypy version is updated so that py39 can work. no impact on py37 and py38.

min-xu-ai: mpi4py is moved into a group only.

PR #304, opened 2021-01-11
[feat] Add AMPnet implementation in experimental dir

sidgoyal78:
# Before submitting

## What does this PR do?
This PR adds the implementation of AMPnet (https://arxiv.org/abs/1705.09786). We replicate the MNIST results from the paper and add a benchmark on identity prediction LM task using transformers.

## PR review

## Did you have fun?


tests/nn/pipe_process/test_pipe.py
```diff
@@ -25,6 +25,8 @@
 import pytest
 import torch
 from torch import nn
+from torch.optim.optimizer import Optimizer
```
min-xu-ai, line 28: does it make sense to have the new tests in a new file?

experimental/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+__version__ = "0.1.4"
```
min-xu-ai, line 6: is this needed?

Can you add some high level doc on:

1. what's this variant of pipeline
2. how to use it

Thanks for moving to the experimental dir. Really appreciate it. I think adding a bit doc and I think it is ready to go. @msbaines @blefaudeux, @anj-s what do you guys think?

tests/nn/pipe_process/test_pipe.py
sidgoyal78: Yeah, good point. I will move new tests under experimental.

experimental/__init__.py
sidgoyal78: Yes, I will add a markdown file with results. 
We're still in the process of getting some more results. But let me add the current set.

sidgoyal78: > is this needed?
I don't think so. Thanks for pointing it out.

tests/nn/pipe_process/test_pipe.py
sidgoyal78: @min-xu-ai : Do you suggest adding these 2 tests in a new file in the same location? Something like tests/nn/pipe_process/test_ampnet_pipe.py

Or do you think we should create a new directory tests under experimental? Something like experimental/tests/nn/pipe_process/test_ampnet_pipe.py

sidgoyal78: As of the recent-most commit, I followed the second option. And we can run those tests using `pytest experimental/tests`

But happy to do it the other way.

min-xu-ai: sounds good to me. thanks a lot!

experimental/__init__.py
sidgoyal78: @min-xu-ai I added a docstring in the class AMPnetPipe. I also added instructions on how to run the benchmark script, experimental_ampnet.py

sidgoyal78:
@msbaines and @min-xu-ai : Could you please take another look? I refactored the code further, added documentation.

msbaines (approved):

PR #306, opened 2021-01-11
[chore][ci] restore 1.5 & 1.6 tests and compatibility

blefaudeux:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


.circleci/config.yml
```diff
@@ -328,7 +325,7 @@ jobs:
           keys:
             - cache-key-benchmarks-{{ checksum "setup.py"}}-{{ checksum "requirements-test.txt"}}
 
-      - <<: *install_dep_17_gpu
+      - <<: *install_dep_17
```
min-xu-ai, line 328: Great catch! I think we need a cmd to check for the expected version and if mismatch failing it. I can add it in my PR if you don't have time adding it here.

min-xu-ai (commented):
whenever this PR is ready for ready, please let me know. looks great so far.

.circleci/config.yml
blefaudeux: I think that some ideas from @anj-s would be a natural match, if we collect the results of the runs somewhere then catching the non-runs would be kind of easy, we would just see (or assert) that there are no results really. Right now the CI is "fire and forget", as long as the job does not crash it's green, even if it actually did nothing :/

fairscale/nn/pipe/pipe.py
```diff
@@ -19,12 +19,12 @@
 
 """The Pipe interface."""
 from collections import OrderedDict
+from dataclasses import dataclass, field
```
blefaudeux, line 22: testing with different conda environments to make sure that I caught all the issues, somehow isort behaves differently it seems and pre-commit agrees, sorry about that

.circleci/config.yml
```diff
@@ -53,37 +53,34 @@ install_dep_16: &install_dep_16
...
 
-install_dep_17_cpu: &install_dep_17_cpu
+
+install_dep_17: &install_dep_17
   - run:
       name: Install Dependencies
       command: |
         sudo apt-get install -y libopenmpi-dev
-        pip install --progress-bar off torch==1.7.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html
         pip install --progress-bar off -r requirements-test.txt
```
blefaudeux, line 68: that's what I was trying to explain @msbaines and @min-xu-ai , 1.5 and 1.6 were never actually running because of that. This line updated torch because of the torchvision dependency, and installed a version not compatible with CUDA on the agents (11). Thus we were initially not testing cuda, and not testing 1.5 and 1.6. 

This was a hotfix to at least make sure that CUDA was running with 1.7 https://github.com/facebookresearch/fairscale/pull/267, and I thought at the time that https://github.com/facebookresearch/fairscale/pull/251 would handle the rest

blefaudeux:
hmm torch 1.5 tests now pass but hang out on exit.. fixing that

fairscale/optim/utils.py
```diff
@@ -44,3 +46,31 @@ def recursive_copy_to_device(value: Any, non_blocking: bool, device: torch.devic
         return device_val
 
     return value
+
+
+def broadcast_object(
+    obj: Any, src_rank: int, group: object = dist.group.WORLD, dist_device: torch.device = torch.device("cpu")
```
min-xu-ai, line 52: add a comment on why is this function needed here?

min-xu-ai (approved):
This is super awesome!

fairscale/optim/utils.py
blefaudeux: ah yes, I added comments where I was pulling that in, it's not in the old torch.. adding another comment here

msbaines:
Thanks much for fixing!

fairscale/utils/testing.py
```diff
@@ -230,10 +243,10 @@ def replacement(*args: Any, **kwargs: Any) -> None:
             if "OMPI_COMM_WORLD_RANK" in os.environ:
                 os.environ["RANK"] = os.environ["OMPI_COMM_WORLD_RANK"]
                 os.environ["WORLD_SIZE"] = os.environ["OMPI_COMM_WORLD_SIZE"]
-                os.environ["MASTER_ADDR"] = "localhost"
-                os.environ["MASTER_PORT"] = "10638"
-                torch.distributed.init_process_group("mpi")
+                _, filename = tempfile.mkstemp()
```
blefaudeux, line 246: @anj-s could that be the faulty part ? looks wrong to me actually now, each rank will get a different file to sync on

anj-s: Looks like this might be the issue. The file that workers sync on needs to be on a shared file system [doc](http://man.hubwiz.com/docset/PyTorch.docset/Contents/Resources/Documents/distributed.html#shared-file-system-initialization) Given the timeout issue which points to workers not connecting to each other, this would make sense.

CircleCI provides a way to share metadata which we can perhaps use for this. I'll try changing this for the current test.

Looking at the previous version of this code it looks like ENV vars were used. It would still require the rank 0 worker IP and port for all processes to connect across workers. If this is on a single machine which I think is the case, can we move to using ENV vars again? Any reason why we prefer a file based approach?

anj-s: Another thing that the file based change does not answer is the flakiness. I would have expected this to always fail and for all torch versions.

blefaudeux: the issue with env variables is that all instances of these tests use the same port, so they often collide with another if the test infra is concurrent (which FB infra is). I would really prefer to stick to a file, it's just that it needs to be shared across the ranks (like for c10 and rpc init), instead of being done here.

blefaudeux: > Another thing that the file based change does not answer is the flakiness. I would have expected this to always fail and for all torch versions.

yep, this I don't know

blefaudeux: proposed fix in https://github.com/facebookresearch/fairscale/pull/316

PR #310, opened 2021-01-13
[feat][OSS] elastic and pytorch compatible checkpoints

### LINKED ISSUE #164: OSS: Make the checkpoints partition-agnostic
## 🚀 Feature
Change the consolidated state dict so that it becomes partition-independent

## Motivation
This would make it possible to change the number of hosts when restarting a job

## Pitch
state_dict() and load_state_dict() need to flatten/shard everything out, instead of storing data per rank

## Alternatives
Current status, same number of ranks before and after

## Additional context
Capturing elements of a discussion with the DeepSpeed MSFT team

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #164, and make the saved state pytorch-compliant (no extra keyword). The number of ranks can change before and after the checkpoints, it will automatically adapt by repartitioning at load. Adding a new unit test which checks reproducibility (cc @joshim5)

## PR review
@joshim5 @stas00 @SeanNaren this breaks compatibility with old checkpoints, is that a big issue ? I could add some scaffolding to move old checkpoints to the new form.

@mannatsingh I know you mentioned that a long time ago, finally there. Not sure how you would rate the complexity of doing this (masking the sharding when rebuilding a pytorch-compatible state dict), but it's now out of the box with this PR

## Did you have fun?


fairscale/optim/oss.py
```diff
@@ -464,14 +469,14 @@ def rank_local_state_dict(rank: int, state_dict: dict) -> dict:
         param_groups = state_dict["param_groups"][state_dict["partition"][rank][0] : state_dict["partition"][rank][1]]
         return {"state": state_dict["state"][rank], "param_groups": param_groups}
 
-    def load_local_state_dict(self, state_dict: dict) -> None:
```
blefaudeux, line 467: I'm not sure whether this is really needed, it was mostly convenient before since the saved state was a list of sharded states, now that it's not the case I figured that was not needed anymore

joshim5:
Incompatibility with old Fairscale checkpoints is fine with me -- but if we could support checkpoints from any arbitrary ADAM optimizer, that would be a huge win! It doesn't even have to be in this class - a separate translation script could work if that's easier.

blefaudeux:
converting to draft, I think the tensor reindexing may be subtly wrong

blefaudeux:
> Incompatibility with old Fairscale checkpoints is fine with me -- but if we could support checkpoints from any arbitrary ADAM optimizer, that would be a huge win! It doesn't even have to be in this class - a separate translation script could work if that's easier.

should be doable with the same PR, I'll have a look at that

anj-s:
@blefaudeux Let me know when this is ready for review.

stas00:
Thank you for the ping, @blefaudeux!

@sgugger did the fairscale integration, so we should probably check with him. (I only wrote the docs and added some basic tests)

I had a quick look and I don't see we are doing anything special for fairscale - checkpoint-wise, just straight pytorch. On the other hand we don't really have any tests at the moment that test checkpointing w/ fairscale (or deepspeed for that matter) so it's possible that it has never worked properly in the first place and you're doing us a favor with this fix. Yay!

blefaudeux:
> Thank you for the ping, @blefaudeux!
> 
> @sgugger did the fairscale integration, so we should probably check with him. (I only wrote the docs and added some basic tests)
> 
> I had a quick look and I don't see we are doing anything special for fairscale - checkpoint-wise, just straight pytorch. On the other hand we don't really have any tests at the moment that test checkpointing w/ fairscale (or deepspeed for that matter) so it's possible that it has never worked properly in the first place and you're doing us a favor with this fix. Yay!

thanks for the feedback ! To be clear it works right now, this is not really a fix but a change of things to (a) align with pytorch (a pytorch checkpoint should load) and (b) make it possible to change the number of ranks in between checkpoints (save with 8 ranks and reload with 16 for instance)

sgugger:
I don't think the compatibility issue is a big matter on our side: for us checkpoints are used to resume training if it was interrupted for some reason, so I don't expect users will have updated their envs in-between.

blefaudeux:
> I don't think the compatibility issue is a big matter on our side: for us checkpoints are used to resume training if it was interrupted for some reason, so I don't expect users will have updated their envs in-between.

alright, perfect, and thanks for dropping by ! I'll remember to keep you in the loop @sgugger, but no big change expected on OSS after this one

blefaudeux:
the optimizer state contents changed in torch1.6+, I'm not handling 1.5 well right now (this change was new to me). Updating that and then putting the PR up for review

blefaudeux:
Ready for review, figuring out something not too ugly which would work for both torch 1.5 (some map indices are id(params)) and above (just plain geometric indices, 0, 1, 2) was interesting, I think that it's mostly ok now.

tests/optim/test_oss.py
```diff
@@ -754,8 +802,29 @@ def closure_sharded(input_tensor=input_tensor):
...
 
+        # Check that the checkpoints are compatible
+        # - get states
+        ddp_state_dict = ddp_optimizer.state_dict()
+        sharded_optimizer.consolidate_state_dict()
+        sharded_optim_state_dict = sharded_optimizer.state_dict()
+        sharded_optim_state_dict = sync_object_ranks(sharded_optim_state_dict, 0, device)
+
+        # - cross load the states
+        ddp_optimizer.load_state_dict(sharded_optim_state_dict)  # mixup on purpose !
```
blefaudeux, line 821: one side effect from this "elastic" change: the checkpoints are provably pytorch compatible

joshim5: this is amazing! I may add a test later on to check this works with model parallelism too

fairscale/optim/oss.py
```diff
@@ -334,12 +321,97 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        """Restore the global parameter groups as well as the shard.
+
+        Arguments:
+            state_dict (dict): optimizer state. Should be an object returned
+                from a call to :meth:`state_dict`
+        """
+
+        super().load_state_dict(state_dict)
```
joshim5, line 383: This is happening on CPU, right? For large models, we might get CPU OOM - e.g. if there are 8 processes (for 8 GPUs). One possible improvement (maybe you are doing this already) is to load on one rank only and then broadcast to the other ranks. Here's an example of how we do this in Fairseq: https://github.com/fairinternal/fairseq-py/blob/master/fairseq/trainer.py#L328

blefaudeux: good question !
- this call should not change the localization as far as I know, so if the state dict that it gets is already on different gpus for instance then the params will stay there. 
- I'll have a second look at the optim state here, I think you're right in that this will copy too much of a state on each rank -unused-, I can prune the unused states for this rank first

joshim5: If we load a pytorch checkpoint that hasn't been sharded, then it's not guaranteed to fit on a single GPU. How about we load on CPU, shard, then move to GPU?
Awesome, pruning the unused states could save a lot of memory for large DDP rank

blefaudeux: > How about we load on CPU, shard, then move to GPU?
when and who does that, on which gpu ? I don't think that there's the needed information at this level, or at least I'm not sure how that will look like. Here people can use `load_state_dict` with a state on CPU, that's completely fine, but I would not make that call within the optimizer, or at least I think that would break the logic for many people

> Awesome, pruning the unused states could save a lot of memory for large DDP rank
done, except for torch1.5 of course, which is different.. incoming fix

fairscale/optim/oss.py
```diff
@@ -334,12 +321,110 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+            logging.warning("Please call `consolidate_state_dict()` beforehand if you meant to save the global state")
+            return self.optim.state_dict()
+
+        # Unify the shard states and the state that pytorch would expect, given the model.
+        # Indexation needs several redirections, since each shard only knows a limited scope of the model
+        # - get the pytorch compliant parameter indexing
+        state_dict = super().state_dict()
+
+        # - get an id map which links the parameter id to the index in the reference state
+        global_id_map = [{id(p): i for pg in self.param_groups for i, p in enumerate(pg["params"])}]
```
msbaines, line 352: This looks wrong. You are creating a single dictionary wrapped in a list. Please add a test for the case where there are multiple param groups. I think it should be:
```suggestion
        global_id_map = [{id(p): i  for i, p in enumerate(pg["params"])} for pg in self.param_groups]
```

msbaines (changes requested):

fairscale/optim/oss.py
```diff
@@ -304,19 +304,6 @@ def clip_grad_norm(
         return total_norm
 
     # State dict interfaces
-    def local_state_dict(self) -> dict:
```
msbaines, line 307: We are removing a public API so could potentially break user code. Did anyone ever use this API?

blefaudeux: I asked, not on the Lightning and HuggingFace side

fairscale/optim/oss.py
blefaudeux: makes sense, I had a quick thought about that but did not follow through. I'll fix and add a test

fairscale/optim/oss.py
SeanNaren: If we were worried, you could just add a deprecation warning for the next release and call the underlying function?

fairscale/optim/oss.py
```diff
@@ -640,9 +628,11 @@ def _setup_bucket_strategy(self) -> None:
                         # This parameter becomes a view of the bucket
                         offset_next = offset + param.numel()
 
-                        self.buckets[device][dst_rank][offset:offset_next].copy_(param.data.flatten())
-                        param.data = self.buckets[device][dst_rank][offset:offset_next].view_as(param.data)
-
+                        try:
```
SeanNaren, line 631: I assume random try/catch is for debugging purposes since this is a draft :P when does this crash? any ideas?

SeanNaren:
Late to the party but yes, totally fine for Lightning :) this is a nice addition, great job as always ben!

fairscale/optim/oss.py
blefaudeux: yes that's a mistake, sorry about that ! closing that today hopefully, different states to handle with pytorch 1.5 and cleanup needed

fairscale/optim/oss.py
blefaudeux: good point, I'll do that

blefaudeux:
I'm really sorry for the complexification here, but the pytorch 1.5 compatibility is a hard requirement. Optimizer states changed in between (1.5)[https://pytorch.org/docs/1.5.1/_modules/torch/optim/optimizer.html#Optimizer.state_dict] and (1.6)[https://pytorch.org/docs/1.6.0/_modules/torch/optim/optimizer.html#Optimizer.state_dict], in that the packed state are indexed with a simple range for the later, while the previous versions indexed using the runtime id(param) (which makes little sense to me, since this is ephemeral from the get go). The trick as far as I understood is that the params were always strictly ordered (hence the "OrderedDict" type), which means that upon reloading one could rely on the order of the parameters as exposed by the param_groups.

This PR handles both cases (1.5 and following), and to the best of my knowledge makes checkpoints completely compatible with vanilla PyTorch, in both directions. @joshim5 could you please check that I'm not breaking anything on your side ? Happy to correct anything, and I would love careful eyes on that one

joshim5:
Yes, let's add a new unit test for it into Fairseq so we can ensure it works going forward. I can probably find time for that tomorrow. Let me know if you're working against a timeline

blefaudeux:
> Yes, let's add a new unit test for it into Fairseq so we can ensure it works going forward. I can probably find time for that tomorrow. Let me know if you're working against a timeline

fine by me, thanks Joshua !

joshim5:
@blefaudeux I took a quick look - doesn't seem to be working out of the box. Maybe I need to use the sharded backend. See stack trace below. Unfortuatnely I don't have anymore time to dig into this but I can look early next week.

Here's the test I ran on 2 GPUs:
```python train.py /checkpoint/bioseq_nonsecure/model-parallel-data/tiny_sample_valid_ur50-bin  --task masked_lm --criterion masked_lm --arch roberta_large --sample-break-mode eos --tokens-per-sample 1023 --max-positions 1024 --max-tokens 1024 --skip-invalid-size-inputs-valid-test --shorten-method random_crop --shorten-data-split-list train --dataset-impl mmap --activation-fn gelu --optimizer adam --encoder-normalize-before  --update-freq 8 --log-interval 10  --encoder-ffn-embed-dim 3072 --encoder-layers 2 --encoder-attention-heads 12 --encoder-embed-dim 768 --max-tokens 1024 --save-dir /checkpoint/jmeier/sharded-oss-test/1/ --save-interval 1 --dataset-impl fasta --log-format json```

Ran for a few epochs then did `mv checkpoint2.pt checkpoint_last.pt` and ran again with `--zero-sharding os`

```
-- Process 1 terminated with the following error:
Traceback (most recent call last):
  File "/private/home/jmeier/.conda/envs/fairseq-20210102-py38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 19, in _wrap
    fn(i, *args)
  File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/distributed_utils.py", line 160, in distributed_main
    main(args, **kwargs)
  File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq_cli/train.py", line 112, in main
    extra_state, epoch_itr = checkpoint_utils.load_checkpoint(
  File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/checkpoint_utils.py", line 168, in load_checkpoint
    extra_state = trainer.load_checkpoint(
  File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/trainer.py", line 347, in load_checkpoint
    self.optimizer.load_state_dict(last_optim_state, optimizer_overrides)
  File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/optim/fairseq_optimizer.py", line 93, in load_state_dict
    self.optimizer.load_state_dict(state_dict)
  File "/private/home/jmeier/fairscale/fairscale/optim/oss.py", line 394, in load_state_dict
    for i_param, key in enumerate(state_dict["state"].keys()):
TypeError: string indices must be integers
```

**Comment by blefaudeux** [comment]
> @blefaudeux I took a quick look - doesn't seem to be working out of the box. Maybe I need to use the sharded backend. See stack trace below. Unfortuatnely I don't have anymore time to dig into this but I can look early next week.
> 
> Here's the test I ran on 2 GPUs:
> `python train.py /checkpoint/bioseq_nonsecure/model-parallel-data/tiny_sample_valid_ur50-bin --task masked_lm --criterion masked_lm --arch roberta_large --sample-break-mode eos --tokens-per-sample 1023 --max-positions 1024 --max-tokens 1024 --skip-invalid-size-inputs-valid-test --shorten-method random_crop --shorten-data-split-list train --dataset-impl mmap --activation-fn gelu --optimizer adam --encoder-normalize-before --update-freq 8 --log-interval 10 --encoder-ffn-embed-dim 3072 --encoder-layers 2 --encoder-attention-heads 12 --encoder-embed-dim 768 --max-tokens 1024 --save-dir /checkpoint/jmeier/sharded-oss-test/1/ --save-interval 1 --dataset-impl fasta --log-format json`
> 
> Ran for a few epochs then did `mv checkpoint2.pt checkpoint_last.pt` and ran again with `--zero-sharding os`
> 
> ```
> -- Process 1 terminated with the following error:
> Traceback (most recent call last):
>   File "/private/home/jmeier/.conda/envs/fairseq-20210102-py38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 19, in _wrap
>     fn(i, *args)
>   File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/distributed_utils.py", line 160, in distributed_main
>     main(args, **kwargs)
>   File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq_cli/train.py", line 112, in main
>     extra_state, epoch_itr = checkpoint_utils.load_checkpoint(
>   File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/checkpoint_utils.py", line 168, in load_checkpoint
>     extra_state = trainer.load_checkpoint(
>   File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/trainer.py", line 347, in load_checkpoint
>     self.optimizer.load_state_dict(last_optim_state, optimizer_overrides)
>   File "/private/home/jmeier/proteinseq_unsupervised_contacts/3rdparty/fairseq-src/fairseq/optim/fairseq_optimizer.py", line 93, in load_state_dict
>     self.optimizer.load_state_dict(state_dict)
>   File "/private/home/jmeier/fairscale/fairscale/optim/oss.py", line 394, in load_state_dict
>     for i_param, key in enumerate(state_dict["state"].keys()):
> TypeError: string indices must be integers
> ```

the message tells me that at this point "state_dict" is a string (like, filepath, not actual file object), is that possible ?  else could you confirm that the state without zero is ddp/pytorch ? because the only compatibility I claim (unit tested) is with pytorch, if it's something fairseq specific then there's not much I could do. I would guess that the issue is in the file load though

**Comment by blefaudeux** [comment]
ping reviewers and testers

File: fairscale/optim/oss.py [hunk 7]
```diff
@@ -319,12 +306,129 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+
+                for local_param_index in local_pg["params"]:
+                    # Update the state, if any
+                    if local_param_index in s["state"].keys():
+                        unordered_state[local_index_to_param_id[local_param_index]] = s["state"][local_param_index]
+
+        # - save the states in the expected order, meaning the one from the param_groups
+        for i, param in enumerate(chain(*(g["params"] for g in self.param_groups))):
+            if id(param) in unordered_state.keys():
+                state_dict["state"][i] = unordered_state[id(param)]
```
msbaines, line 374: Why can't we just add the parameter directly here to the returned state_dict and avoid the later loops?
```suggestion
                for local_param, global_param in zip(local_pg["params"], global_pg["params"]):
                    global_id = global_id_map[id(global_param)]
                    state_dict["state"][global_id] = local_parm
```

blefaudeux: kind of a mind breaker (the indices all flip) but I think that can work indeed, cleaner, thanks for the suggestion

blefaudeux: just to explain a little, the suggestion does not work out of the box because the `local_pg["params"]` is just a list of indices, there are no real params in there, and we need to pull the corresponding state from the smaller optimizer, but I think that the idea works in that I can reassign directly here instead of looping twice indeed

blefaudeux: [done, not exact same as explained above but similar enough, one less loop]

fairscale/optim/oss.py
```diff
@@ -310,12 +297,116 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+            # Check that this rank owns this param, if not remove from the state
+            param = global_id_map[i_param]
+            if self.param_to_rank[param] != self.rank:
+                state_dict["state"][key] = None
+
+        super().load_state_dict(state_dict)
+
+        # Set the sharded optimizer state.
+        # Keep the original type (not respected by PyTorch which casts to the model type)
+        for k, (_, v) in enumerate(state_dict["state"].items()):
```
msbaines, line 386: I think this loop can be combined with the loop above and would result in simpler code.

fairscale/optim/oss.py
```diff
@@ -310,12 +297,116 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+                        self.optim.state[param] = recursive_copy_to_device(v, non_blocking=True, device=param.device)
+
+        # Update the param_group keys
+        for fpg, pg in zip(state_dict["param_groups"], self.param_groups):
+            for key in fpg.keys():
+                if key != "params":
+                    pg[key] = fpg[key]
+
+        # Sync with the optimizer param groups
+        self._sync_param_groups(local_to_global=False)
```
msbaines, line 402: Might be simpler if there was just a function which synced between a src and dst param group
```suggestion
        self._sync_param_groups(self.param_groups, state_dict["param_groups"])
        self._sync_param_groups(self.optim.param_groups, self.param_groups)
```

fairscale/optim/oss.py
```diff
@@ -310,12 +297,116 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+        Arguments:
+            state_dict (dict): optimizer state. Should be an object returned
+                from a call to :meth:`state_dict`
+        """
+
+        # NOTE: PyTorch 1.5 does not index linearly but with the id(params) at saving time
+        # we work around that here by using the fact that the params are ordered as in the param_groups
+
+        # Param index to param map
+        global_id_map = {i: p for i, p in enumerate(chain(*(g["params"] for g in self.param_groups)))}
```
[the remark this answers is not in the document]
blefaudeux: what if the user adds param_groups ? then it needs to be updated. It's a one liner, but it can be moved somewhere else , of course

fairscale/optim/oss.py
```diff
@@ -310,12 +297,116 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+
+        # Prune the state_dict from the states which this rank does not own, then normal base load
+        for i_param, key in enumerate(state_dict["state"].keys()):
+            # Check that this rank owns this param, if not remove from the state
+            param = global_id_map[i_param]
+            if self.param_to_rank[param] != self.rank:
+                state_dict["state"][key] = None
```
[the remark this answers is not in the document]
blefaudeux: could you elaborate on why it's better ? GC ?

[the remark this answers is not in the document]
blefaudeux: Well, you cannot change the dictionary size while iterating on it, so that's not an option anyway

fairscale/optim/oss.py
blefaudeux: good point, looks a lot nicer indeed

fairscale/optim/oss.py
blefaudeux: done in a new revision, I'm not sure whether it's simpler to understand but it does work and it's more concise indeed

fairscale/optim/oss.py
blefaudeux: (the first line is flipped though)

fairscale/optim/oss.py
```diff
@@ -310,12 +297,116 @@ def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
...
+
+        # NOTE: PyTorch 1.5 does not index linearly but with the id(params) at saving time
+        # we work around that here by using the fact that the params are ordered as in the param_groups
+
+        # Param index to param map
+        global_id_map = {i: p for i, p in enumerate(chain(*(g["params"] for g in self.param_groups)))}
+
+        # Prune the state_dict from the states which this rank does not own, then normal base load
+        for i_param, key in enumerate(state_dict["state"].keys()):
+            # Check that this rank owns this param, if not remove from the state
```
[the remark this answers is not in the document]
blefaudeux: me trying to explain what's happening :)

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [next rel] - TBD
+### Added
+- Pytorch compatibility for OSS checkpoints
+- Elastic checkpoints for OSS, world size can vary in between save and loads
+- Tensor views for OSS bucketing, reduced CPU use
```
blefaudeux, line 11: this one is from an earlier PR, forgotten

fairscale/optim/oss.py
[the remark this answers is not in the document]
msbaines: I think adding param_groups after __init__ is undefined behavior. I don't think this is code is robust to param_groups being added for other reasons.

In addition to the code duplication, iterating over all params seems like it would be expensive for a large model so it might be worth the savings to do it just once.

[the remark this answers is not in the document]
blefaudeux: There's a optim.add_param_group() method, I'm a bit surprised that this would lead to undefined behavior, I'm not sure that we're talking about the same thing. Else sure I can cache this somewhere else, hard to guess what people prefer (it will add an indirection), but sure, I will do that.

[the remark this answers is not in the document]
blefaudeux: https://pytorch.org/docs/stable/optim.html#torch.optim.Optimizer.add_param_group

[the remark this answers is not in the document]
blefaudeux: updated

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [next rel] - TBD
+### Added
```
anj-s, line 8: Shall we add the PR number for each of these updates? Looks like we've been doing that in the past.

blefaudeux: yes, good point, will do !

blefaudeux:
ping reviewers, happy to change anything if it means that this can land somehow, hard to keep old PRs alive

blefaudeux:
This fixes the linting that https://github.com/facebookresearch/fairscale/pull/342 is supposed to have broken, even if not touching the files

blefaudeux:
> This fixes the linting that #342 is supposed to have broken, even if not touching the files

solved with #348

msbaines (commented):
Can you add a test for multiple param_groups? There was a bug related to that in an earlier iteration of this change. Other than that, LGTM.

msbaines (approved):
Please add a test for multiple param_groups before submitting.

tests/optim/test_oss.py
```diff
@@ -619,48 +678,36 @@ def run_state_dict_distributed(rank, world_size, tempfile_name):
     sharded_optimizer2 = optim.OSS(model_oss2.parameters(), lr=0.1, momentum=0.99)
     sharded_optimizer2.add_param_group({"params": head_oss2.parameters()})
```
blefaudeux, line 679: @msbaines this uses multiple param groups (through the  `.add_param_groups` interface that I mentioned another time)

tests/optim/test_oss.py
```diff
@@ -726,11 +766,17 @@ def check_optimizer_equivalence(optimizer: Type[torch.optim.Optimizer]):
         model.register_buffer("test_buffer", torch.ones((1)) * rank)
         model.to(device)
 
-        sharded_optimizer = optim.OSS(params=model.parameters(), optim=optimizer, lr=1e-3)
+        optimizer_settings = {"lr": 1e-3}
```
blefaudeux, line 769: @msbaines I can add another group in this ddp parity test, let me know if you had something else in mind

blefaudeux:
> Please add a test for multiple param_groups before submitting.

the ddp_parity test takes 2 groups now, with different learning rates, let me know if you had something else in mind

msbaines:
global_id_map was incorrect in an earlier iteration for multiple param_groups but no tests were failing. Would be nice to have a test that verifies that new load_state_dict and state_dict logic work when there are multiple param_groups.

> > Please add a test for multiple param_groups before submitting.
> 
> the ddp_parity test takes 2 groups now, with different learning rates, let me know if you had something else in mind

global_id_map was incorrect in an earlier iteration of the patch for the case of multiple param_groups but no tests were failing. Would be nice to have a test that verifies that new load_state_dict and state_dict logic work when there are multiple param_groups. Not sure if the ddp_parity test handles that.

tests/optim/test_oss.py
```diff
@@ -719,19 +759,51 @@ def run_ddp_parity(rank, world_size, backend, temp_file_name):
...
+            {"params": trunk.parameters(), "lr": 1e-5},
+            {"params": head.parameters(), "lr": 1e-4},
+        ]
+
+        optimizer_settings = {}
+        if isinstance(optim, torch.optim.SGD):
+            optimizer_settings["momentum"] = 0.9
+
+        sharded_optimizer = optim.OSS(
+            params=oss_trainable_params,
```
blefaudeux, line 787: @msbaines

tests/optim/test_oss.py
blefaudeux: @msbaines the states are saved on both pytorch and oss sides (defined with multiple groups) and cross loaded

blefaudeux:
> global_id_map was incorrect in an earlier iteration for multiple param_groups but no tests were failing. Would be nice to have a test that verifies that new load_state_dict and state_dict logic work when there are multiple param_groups.
> 
> > > Please add a test for multiple param_groups before submitting.
> > 
> > 
> > the ddp_parity test takes 2 groups now, with different learning rates, let me know if you had something else in mind
> 
> global_id_map was incorrect in an earlier iteration of the patch for the case of multiple param_groups but no tests were failing. Would be nice to have a test that verifies that new load_state_dict and state_dict logic work when there are multiple param_groups. Not sure if the ddp_parity test handles that.

yes it does, it checks that:
- after a couple of steps you get the exact same in both cases
- if you save dict on both sides, then cross load (ddp loads the oss state dict, and vice versa) nothing breaks
- if you iterate after that, you still have the same results on both sides

There's also another test which checks that on OSS alone stepping, then rolling back (load an old state dict) and stepping again gives the same result, this one has one param group only, always possible to add more of course

PR #317, opened 2021-01-21
Add FlattenParamsWrapper

myleott:
Adds a generic Module wrapper that flattens the Module's parameters into a single contiguous Parameter. This can be useful in many places both as a convenience and to improve speed.

### DIFF
--- a/fairscale/nn/misc/flatten_params_wrapper.py
+++ b/fairscale/nn/misc/flatten_params_wrapper.py
@@ -1,0 +1,155 @@
+# Copyright (c) Tongzhou Wang
+# Licensed under the MIT License.
+
+from collections import namedtuple
+from contextlib import contextmanager
+from typing import List, Optional
+
+import torch
+import torch.nn as nn
+
+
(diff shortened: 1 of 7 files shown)


fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) Tongzhou Wang
+# Licensed under the MIT License.
```
msbaines, line 2: You'll need to update the NOTICE file and the LICENSE and REFERENCES section of README.md

msbaines (commented):
LGTM. Thanks! Just need to update NOTICE and README.md.

fairscale/nn/misc/flatten_params_wrapper.py
myleott: Done!

myleott:
Ah, looks like some lint issues causing CI to break. Will submit a fix.

Separately, can we make CI run automatically on PRs?

PR #320, opened 2021-01-21
[fix] Lint flattenparams

blefaudeux:
# Before submitting

## What does this PR do?
Fixes a broken master. Unfortunately both overload and OrderedDict support is dependent on python versions and mostly broken, fixing that in the least disruptive way I believe (I did try other options)

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/misc/flatten_params_wrapper.py
+++ b/fairscale/nn/misc/flatten_params_wrapper.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 from contextlib import contextmanager
-from typing import Any, Dict, Generator, List, Optional, OrderedDict, Tuple
+from typing import Any, Dict, Generator, List, Optional, Tuple
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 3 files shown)


anj-s (approved):

anj-s:
Will this fix master?

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -137,9 +127,7 @@ def __getattr__(self, name: str) -> Any:
         except AttributeError:
             return getattr(self.module, name)  # fallback to wrapped module
 
-    def state_dict(
-        self, prefix: str = "", keep_vars: bool = False,
-    ) -> OrderedDict[str, Tensor]:
+    def state_dict(self, prefix: str = "", keep_vars: bool = False) -> "OrderedDict[str, Tensor]":  # type: ignore
```
blefaudeux, line 130: mypy overloading is supposed to be handled like that : declare the interfaces first, then the implementations ((see)[https://mypy.readthedocs.io/en/stable/more_types.html?highlight=overload#function-overloading]). For some reason I could not get it to agree with the stub, +OrderedDict have a different type depending on the python versions (only properly defined after 3.7.5 it seems). All in all I thought that # type: ignore was the most reasonable path

PR #326, opened 2021-01-25
[refactor] pipe: separate out Single and MultiProcess pipe

msbaines:
The two implemntations share very little code. The code is much
more readable now that all the ifs are gone.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


sidgoyal78 (commented):
Thanks for the refactor, Mandeep. This is going be very helpful for readability :)
I just noticed that some of the tests are failing. Are you working on fixing those? Wanted to make sure before looking at the changes more closely.

examples/tutorial_pipe_multiprocess.py
```diff
@@ -27,10 +27,10 @@ def run(rank, world_size):
 
     device = torch.device("cuda", RANK) if DEVICE == "cuda" else torch.device("cpu")
 
-    model = fairscale.nn.Pipe(
+    model = MultiProcessPipe(
```
anj-s, line 30: Should we specify this separately if it is not an API? Also it might be confusing to understand MultiProcessPipe with the style argument below.

fairscale/nn/pipe/mppipe.py
```diff
@@ -0,0 +1,660 @@
...
+        TypeError:
+            the module is not a :class:`nn.Sequential <torch.nn.Sequential>`.
+        ValueError:
+            invalid arguments, or wrong balance
+        IndexError:
+            the number of devices is fewer than the number of partitions.
+
+    """
+
+    MultiProcess: PipelineStyle = PipelineStyle.MultiProcess
```
anj-s, line 391: Can we add a comment about what these two style options do? Can we say that the default style is MultiProcess and AsyncSchedule is a special case?

anj-s:
Thank you for the refactor! This definitely helps understand nn.Pipe better. 

nit: Can we rename files to be multiprocess_pipe or mp_pipe?

msbaines:
> Thanks for the refactor, Mandeep. This is going be very helpful for readability and future extensions :)
> I just noticed that some of the tests are failing. Are you working on fixing those? Wanted to make sure before looking at the changes more closely.

Yes working on fixing.

fairscale/nn/pipe/mppipe.py
msbaines: Cool if I address this in a future change? Still thinking about how to handle Async.

tests/nn/pipe_process/test_pipe.py
```diff
@@ -31,9 +31,11 @@
     get_pipeline_parallel_group,
     initialize_model_parallel,
 )
-from fairscale.nn.pipe import LazyModule, Pipe
+from fairscale.nn.pipe import LazyModule, MultiProcessPipe
 from fairscale.utils.testing import get_worker_map, set_random_seed, torch_spawn, torch_version
 
+Pipe = MultiProcessPipe
```
sidgoyal78, line 37: Quick question: Would be it confusing if we just import MultiProcessPipe as Pipe ( i mean `from fairscale.nn.pipe import MultiProcessPipe as Pipe`)

I see you have are using in couple of tests.

Not sure if it will be good for readability, but just putting it out there :)

benchmarks/pipe.py
```diff
@@ -25,7 +25,7 @@
 from fairscale.nn import Pipe
```
sidgoyal78, line 25: In this script we are using model.group check in the `log_number_of_parameters` function (which i saw is a reason for a test failure). We'll have to modify that to check for Pipelinestyle instead of group.

sidgoyal78 (commented):
I just put in a few minor comments. Thanks for the PR, overall looks great.
Before merging, there's one more change that has to be done to pass the `benchmark_single_process` test.

examples/tutorial_pipe_multiprocess.py
msbaines: I'd like to limit the scope of this change to just separating Pipe from MultiProcessPipe. I plan to focus on cleaning up MultiProcessPipe in subsequent changes. In this change I've limited myself to renaming and dead-code removal.

sidgoyal78 (approved):
LGTM, thanks.

anj-s (approved):

PR #327, opened 2021-01-26
[ShardedDDP] Bucketing reduce calls, tensor views

blefaudeux:
# Before submitting

## What does this PR do?
Speed up for high latency environments, bucket the calls in the reduce phase using tensor views (no extra copies). Testing on FBCluster yielded double digits speed ups for 4 nodes+ jobs

## PR review
cc @vreis @mannatsingh 

## Did you have fun?


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -69,6 +73,7 @@ def __init__(
         process_group: Any = None,
         broadcast_buffers: bool = True,
         sync_models_at_startup: bool = True,
+        reduce_buffer_size: int = 2 ** 23,
```
ananthsub, line 76: n00b question: is there a good rule of thumb for how to tune this parameter?

blefaudeux: not that I know really, it would depend on the speed/latency of the interconnect in between the nodes, and the amount of memory that you're ready to loose (this kills the gradient sharding basically)

blefaudeux: I tested this value and it provides a very significant speedup vs. previous status, but I did not have time to do a sweep honestly

blefaudeux:
ping reviewers, pretty please

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -113,6 +118,24 @@ def __init__(
         # - keep track of the grads which have already been reduced
         self._reduced_grads: Dict[OSS, int] = {}
         self._reduced_grads_max = {o: len(o.param_to_rank.values()) for o in self.sharded_optimizers}
+
+        # - setup buckets and tensor views
+        model_size = sum([p.numel() for p in self.module.parameters()])
+        if dist.get_world_size(self.process_group) <= 8:
```
min-xu-ai, line 124: 8 -> torch.cuda.device_count()?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -212,23 +240,43 @@ def reduce(*_: Any) -> None:
...
+                        )
                     )
-                )
-                self._reduced_grads[optimizer] += 1
+                    self._reduced_grads[optimizer] += 1
+                else:
+                    bucket = self.buckets[optimizer][param.device][dst_rank]
+                    bucket.params_checked_in += 1
+
+                    if bucket.full():
```
min-xu-ai, line 266: will there be a chance the bucket is not full and reduce is called? if so, why? Do we have to deal with models that has unused params like find_unused_params in ddp?

fairscale/optim/utils.py
```diff
@@ -76,3 +76,25 @@ def broadcast_object(
...
+    """
+
+    def __init__(self, buffer: torch.Tensor) -> None:
+        # The actual flat tensor
+        self.buffer = buffer
+        self.max_size = buffer.numel()
+
+        # Current status for this buffer
+        self.params_checked_in = 0
+        self.max_params_checked_in = 0
```
min-xu-ai, line 93: add a comment on who sets this value for real? It is not clear to me here.

tests/optim/test_oss.py
```diff
@@ -262,6 +262,9 @@ def run_test_zero_grad(rank, world_size, tempfile_name):
 
 def test_zero_grad():
     world_size = 2
+    if torch.cuda.is_available() and torch.cuda.device_count() < world_size:
+        world_size = min(world_size, torch.cuda.device_count())
```
min-xu-ai, line 266: can world_size be 1 allowed here?

min-xu-ai (approved):
looks good. some questions and comments.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: I think that replicating the DDP feature to find unused parameters would be a lot of work without any visible upside, it requires tracing the whole FW pass and it's already doable with DDP. The gist of the issue around that here is that there's no signal at the python level for when the BW pass is done, so no real way that I know of to catch that, and I've tried (cc @mrshenli, but this has been discussed extensively already).
If there's no unused parameter, then all the buckets will be flushed, I can add an assert in the next FW so that if it was not the case it can recommend to check for unused params ?

tests/optim/test_oss.py
blefaudeux: this comes from another branch actually, it was just to check CI. Doing a mega-refactor of this ugly piece of code (my bad) right now

fairscale/nn/data_parallel/sharded_ddp.py
min-xu-ai: understood. adding a comment would be good. can revisit later for sure

blefaudeux: adding an assert right now, it's a very good call I think, results would be wrong if this is missed

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: in that case it was about the number of ranks, but I'm not sure that it's that good of a call actually, could also be removed and be kept entirely in the user scope.

blefaudeux: removed, I think it would need more testing, better to remove a branch I suppose

PR #329, opened 2021-01-27
[fix] OSS: removing the torch broadcast util altogether, broken on 1.7.1

blefaudeux:
# Before submitting

## What does this PR do?
Unbeknownst to me this was broken in 1.7, fixed in 1.8 but not 1.7.1 (see https://github.com/pytorch/pytorch/pull/46897, the device is not respected). Thanks @SeanNaren for the heads up
May not be the only explanation, but definitely broken and part of https://github.com/PyTorchLightning/pytorch-lightning/issues/5646

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/pipe/pipe.py
+++ b/fairscale/nn/pipe/pipe.py
@@ -202,7 +202,7 @@ class Pipe(Module):
             ``'except_last'``, or ``'never'`` (default: ``'except_last'``)
         deferred_batch_norm (bool):
             whether to use deferred BatchNorm moving statistics (default:
-            :data:`False`, see :ref:`Deferred Batch Normalization` for more
+            :data:`False`, see :class:`Deferred Batch Normalization <DeferredBatchNorm>` for more
             details)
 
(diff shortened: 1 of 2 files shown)


min-xu-ai (approved):
LGTM. One idea is to add a unit test that assert this incorrect behavior in 1.7.1 and once we upgrade and if it is fixed, the unit test would fail and tell us to use the upstream code. But that's completely optional.

blefaudeux:
> LGTM. One idea is to add a unit test that assert this incorrect behavior in 1.7.1 and once we upgrade and if it is fixed, the unit test would fail and tell us to use the upstream code. But that's completely optional.

ah I just saw that, yes I agree but I'm not sure that it will ever be fixed in 1.7 actually, so it means that we'll be able to flip in quite some time (since we support the last 3 versions). In the meantime the upstreamed OSS is using the torch util, I guess that at that time we'll deprecate the fairscale implementation of OSS and only use PyTorch's ?

PR #333, opened 2021-01-27
[fix] CPU OSS tests

blefaudeux:
# Before submitting

## What does this PR do?
Typo, the world size was broken on CPU / no cuda available, not visible before somehow

## PR review

## Did you have fun?

### DIFF
--- a/tests/optim/test_oss.py
+++ b/tests/optim/test_oss.py
@@ -237,7 +237,7 @@ def run_test_add_param_group(rank, world_size, tempfile_name):
 
 def test_add_param_group():
     world_size = 4
-    if not torch.cuda.is_available() or torch.cuda.device_count() < world_size:
+    if torch.cuda.is_available() and torch.cuda.device_count() < world_size:
         world_size = min(world_size, torch.cuda.device_count())
 
     temp_file_name = tempfile.mkstemp()[1]
@@ -262,6 +262,9 @@ def run_test_zero_grad(rank, world_size, tempfile_name):
 
 def test_zero_grad():
     world_size = 2
+    if torch.cuda.is_available() and torch.cuda.device_count() < world_size:
+        world_size = min(world_size, torch.cuda.device_count())
+
     temp_file_name = tempfile.mkstemp()[1]
     mp.spawn(run_test_zero_grad, args=(world_size, temp_file_name), nprocs=world_size, join=True)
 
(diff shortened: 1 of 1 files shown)


tests/optim/test_oss.py
```diff
@@ -237,7 +237,7 @@ def some_trainable():
 
 def test_add_param_group():
     world_size = 4
-    if not torch.cuda.is_available() or torch.cuda.device_count() < world_size:
+    if torch.cuda.is_available() and torch.cuda.device_count() < world_size:
         world_size = min(world_size, torch.cuda.device_count())
```
blefaudeux, line 241: this was essentially setting world_size == 0 if cuda was not available

tests/optim/test_oss.py
```diff
@@ -262,6 +262,9 @@ def run_test_zero_grad(rank, world_size, tempfile_name):
 
 def test_zero_grad():
     world_size = 2
```
blefaudeux, line 264: unrelated but opportunistic, this was broken if run with a single cuda GPU

tests/optim/test_oss.py
```diff
@@ -474,7 +477,11 @@ def closure():
                     dist.gather(p, receptacle, dst=0, group=process_group)
                     if rank == 0:
                         for sync_p in receptacle[1:]:
-                            assert torch.all(torch.eq(receptacle[0], sync_p)), "Models differ in between ranks"
+                            assert torch.all(
```
min-xu-ai, line 480: This is non-deterministic failure seems hard to catch. Perhaps you can run this test in a loop for 1000 times just to make sure it is not hitting locally? Otherwise, looks great.

min-xu-ai (approved):

tests/optim/test_oss.py
blefaudeux: no so it was deterministic, it's just that one needed to have the right environment (without cuda), which I didn't have, so to begin with I couldn't repro. Else it's an old test, always been reliable, I've just changed this line so that it displays the values when something is wrong (to check whether it's a tolerance thing or if the values are not related altogether)

min-xu-ai: I see. I was surprised that it failed with py37 not 38.

min-xu-ai: and wasn't failing before.

min-xu-ai: interesting, similar but different place assert:

https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1409/workflows/68a4e48f-1e55-4783-af06-107f70c3690b/jobs/6326

PR #335, opened 2021-01-28
[cleanup] multiprocess_pipe: dead-code removal and simplification

msbaines:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/pipe/async_schedule.py
+++ b/fairscale/nn/pipe/async_schedule.py
@@ -18,7 +18,7 @@ from fairscale.nn.model_parallel import get_pipeline_parallel_ranks
 from .messages import Transport
 from .microbatch import Batch
 from .skip.tracker import SkipTrackerThroughPotals
-from .types import EVENT_LOOP_QUEUE, PipelineStyle, PipeMessage, Tensors
+from .types import EVENT_LOOP_QUEUE, PipeMessage, Tensors
 
 
 @dataclass(frozen=True)
@@ -194,13 +194,7 @@ class AsyncEventLoop:
         from .multiprocess_pipeline import create_task
 
         task = create_task(
-            PipelineStyle.AsyncSchedule,
-            self.checkpoint_stop,
-            batch.index,
-            self.group.rank(),
-            batch,
-            partition.module,
-            skip_trackers,
+            self.checkpoint_stop, batch.index, self.group.rank(), batch, partition.module, skip_trackers,
         )
         result = task.compute()
         task.finalize(result)
--- a/fairscale/nn/pipe/async_schedule.py
+++ b/fairscale/nn/pipe/async_schedule.py
@@ -17,6 +17,7 @@ from fairscale.nn.model_parallel import get_pipeline_parallel_ranks
 
 from .messages import Transport
 from .microbatch import Batch
+from .multiprocess_pipeline import create_task
 from .skip.tracker import SkipTrackerThroughPotals
 from .types import EVENT_LOOP_QUEUE, PipeMessage, Tensors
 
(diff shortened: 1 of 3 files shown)


fairscale/nn/pipe/multiprocess_pipe.py
```diff
@@ -463,7 +461,6 @@ def __init__(
 
         self.group = group
         self.worker_map = worker_map
-        self.input_device = input_device
```
sidgoyal78, line 466: We are using this in ampnet
https://github.com/facebookresearch/fairscale/blob/master/experimental/nn/ampnet_pipe/pipe.py#L60

fairscale/nn/pipe/async_schedule.py
```diff
@@ -194,13 +194,7 @@ def run_invocation(
...
 
         task = create_task(
-            PipelineStyle.AsyncSchedule,
-            self.checkpoint_stop,
-            batch.index,
-            self.group.rank(),
-            batch,
-            partition.module,
-            skip_trackers,
+            self.checkpoint_stop, batch.index, self.group.rank(), batch, partition.module, skip_trackers,
```
sidgoyal78, line 197: I think this is needed in the multiprocess_pipeline.py function in the run() function to distinguish the even-loop and non-event-loop cases.

sidgoyal78 (changes requested):
Thanks for the PR, Mandeep. I have put a few inline comments:

min-xu-ai (approved):
nice. looks shorter and cleaner

fairscale/nn/pipe/async_schedule.py
```diff
@@ -194,13 +194,7 @@ def run_invocation(
         from .multiprocess_pipeline import create_task
```
anj-s, line 194: can we move this to the top of the file?

fairscale/nn/pipe/multiprocess_pipe.py
```diff
@@ -622,7 +613,7 @@ def back_helper(self, output: List[microbatch.Batch]) -> None:
             raise ValueError("back_helper should only be called on non-final stages")
 
         if self.pipeline:
-            self.pipeline.back_helper(list(reversed(output)))
+            self.pipeline.back_helper(output)
```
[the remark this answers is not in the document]
msbaines: I am moving the reversed from here to multiprocess_pipeline. The code feels clearer this way. I prefer passing the unmodified output and then reversing when I need to. It was not obvious in multiprocess_pipeline that the output was reversed.

fairscale/nn/pipe/async_schedule.py
msbaines: Yes. But it is not used inside create_task.

fairscale/nn/pipe/multiprocess_pipe.py
msbaines: Ah. Good catch. Fixed.

fairscale/nn/pipe/multiprocess_pipeline.py
```diff
@@ -325,7 +319,6 @@ def compute(
         """Runs tasks with synchronization to copy streams."""
```
anj-s, line 320: If we do a quick return here for any style that is not MultiProcess, can we get rid of the two if statements below?

anj-s:
Would it be possible to add more comments to the code as you do the cleanup? It would really help understand this piece of code better. Thanks!

fairscale/nn/pipe/multiprocess_pipeline.py
msbaines: Great catch! Fixed.

fairscale/nn/pipe/async_schedule.py
msbaines: This is needed here to avoid a cyclic dependency. Added a TODO to remove the cycle.

sidgoyal78 (approved):
Thanks for the changes. Looks good.

PR #339, opened 2021-01-28
[test]: test with py39 + torch 1.8 nightly

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/adascale.py
+++ b/fairscale/optim/adascale.py
@@ -289,7 +289,7 @@ class AdaScale(Optimizer):
         if pg_idx is not None:
             return self._state["grad_sqr_avg"][pg_idx]
         else:
-            return np.sum(self._state["grad_sqr_avg"])
+            return float(np.sum(self._state["grad_sqr_avg"]))
 
     def _grad_var_avg(self, pg_idx: Optional[int] = None) -> float:
         """
@@ -307,7 +307,7 @@ class AdaScale(Optimizer):
         if pg_idx is not None:
             return self._state["grad_var_avg"][pg_idx]
         else:
-            return np.sum(self._state["grad_var_avg"])
+            return float(np.sum(self._state["grad_var_avg"]))
 
     def gain(self, pg_idx: Optional[int] = None) -> float:
         """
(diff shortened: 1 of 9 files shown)


min-xu-ai:
This should make our test coverage better with py39. The torch-pg dependency can be updated later.

blefaudeux:
we could remove torchvision from requirements-test Min, and only install it on the agent which runs the benchmark, we only run it once anyway ?

min-xu-ai:
> 
> 
> we could remove torchvision from requirements-test Min, and only install it on the agent which runs the benchmark, we only run it once anyway ?

Great idea. I just did that. I separated the requirements for benchmarks out. See this commit.

https://github.com/facebookresearch/fairscale/pull/339/commits/56563b6987ed09471bf07621b410a1b36f32226f

requirements-benchmarks.txt
```diff
@@ -0,0 +1,7 @@
+# Bring in everything that tests depends on.
```
blefaudeux, line 1: _like_

fairscale/optim/adascale.py
```diff
@@ -514,8 +514,9 @@ def add_param_group(self, pg: Dict) -> None:
         # Extend the states.
         for name in self._state.keys():
             assert name.startswith("grad_sqr_avg") or name.startswith("grad_var_avg"), name
-            if isinstance(self._state[name], int):
-                # This is the "_count" variable.
+            if name.endswith("_count"):
```
blefaudeux, line 517: hmm, is that expected ? looks like a logic change

blefaudeux (approved):
Lokos like there are a few adascale changes which slipped in, checking here that you're on top of that ? Else LGTM !

fairscale/optim/adascale.py
min-xu-ai: right, it is induced by mypy checking. newer numpy version actually enabled mypy to find out that adascale code here had a mixed typing, so it flagged it. I fixed it by change the typing to not be mixed. So, it is a typing change. No logic changes.

PR #346, opened 2021-01-31
[refactor] pipe: simplify balance and module checks

msbaines:
Also, remove run-time type checks. mypy is already checking types.
However, a user could still pass the wrong types and not use
type-checking.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/pipe/multiprocess_pipe.py
+++ b/fairscale/nn/pipe/multiprocess_pipe.py
@@ -53,85 +53,22 @@ else:
     NamedModules = OrderedDict
 
 
-def recommend_auto_balance(message: str) -> str:
-    """Expands a message with recommendation to :mod:`torchpipe.balance`."""
-    return f"""{message}
-
-If your model is still under development, its optimal balance would change
-frequently. In this case, we highly recommend 'fairscale.nn.pipe.balance' for
-naive automatic balancing:
-
-  from fairscale.nn import Pipe
-  from fairscale.nn.pipe.balance import balance_by_time
-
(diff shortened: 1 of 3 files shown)


fairscale/nn/pipe/multiprocess_pipe.py
```diff
@@ -53,85 +53,22 @@
...
-
-    if filter_unique:
-        module_len = len(set(map(id, module)))
-    else:
-        module_len = len(module)
-
-    if module_len != sum(balance):
-        raise BalanceError(
+def check_balance(module: Union[nn.Sequential, List[LazyModule]], balance: List[int]) -> None:
+    if len(module) != sum(balance):
```
sidgoyal78, line 62: I'm curious why this isn't `len(set(map(id, module))) != len(module)`, similar to the condition in `verify_module()`?

sidgoyal78 (commented):
Thanks for the refactor. I have put an inline question.

anj-s (approved):

fairscale/nn/pipe/multiprocess_pipe.py
msbaines: One verifies that there are no duplicates in the list of layers. While the other verifies that the size of the list of layers matches the size of balance.

sidgoyal78: Yeah, but from the code that you removed (at one place), the condition that I mentioned was getting called?

PR #347, opened 2021-02-01
[feat] Add AdaScaleWrapper

### LINKED ISSUE #302: [test] ensure adascale works with zero
## 🚀 Feature

ensure adascale works with zero.

## Motivation

we should try to make sure all libraries here work with each other.

## Pitch

avoid unexpected pitfalls when users try to combine them.

## Alternatives

document it if it is hard to make work.

## Additional context

this is mainly to add a unit test.

min-xu-ai:
- This enables a different API for wrapping an optimizer with AdaScale.
- This also enables AdaScale to be wrapped by OSS.
- However, OSS wrapping AdaScale results in different optimization,
  which future research will be needed to study its effects.

testing: add unit tests.

# Before submitting

## What does this PR do?
Fixes #302.

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/adascale.py
+++ b/fairscale/optim/adascale.py
@@ -32,13 +32,18 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 import functools
-from typing import Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type
 
 import numpy as np
 import torch
 from torch.autograd import Variable
 import torch.distributed as dist
-from torch.optim import Optimizer
+from torch.optim import SGD, Optimizer
+
+if TYPE_CHECKING:  # pragma: no cover
+    from torch.optim.optimizer import _params_t
+else:
+    _params_t = Any
 
 
 class AdaScale(Optimizer):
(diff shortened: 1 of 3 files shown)


min-xu-ai:
@blefaudeux, what do you think about this? I will add a test with shard_ddp next if this looks OK. For the next time, small shard_ddp change might be needed to detect that AdaScale wrapping OSS is also allowed in shard_ddp.

min-xu-ai:
ping reviewers

msbaines (commented):
Why is a separate wrapper class necessary? Why can't we just change AdaScale directly to work as a wrapper?

min-xu-ai:
> 
> 
> Why is a separate wrapper class necessary? Why can't we just change AdaScale directly to work as a wrapper?

Very good question. The current AdaScale API takes an instantiated optimizer object. That won't work with OSS, which expecting to wrap a optimizer that takes list of parameters. The wrapper allows both way of the initialization to be possible. This allow AdaScale to be wrapped by OSS. However, numerically and ML algorithm-wise, OSS wrap AdaScale is different from AdaScale's original idea, which mean that requires more research. But this wrapper allows such research to be done in the future. (i.e. study the effect of OSS wrapping AdaScale).

min-xu-ai:
> > Why is a separate wrapper class necessary? Why can't we just change AdaScale directly to work as a wrapper?

I should add that AdaScale(OSS) already works. OSS(AdaScale) is what this is trying to address.

fairscale/optim/adascale.py
```diff
@@ -582,3 +587,47 @@ def set_num_gradients_to_accumulate(self, num_gradients_to_accumulate: int, upda
...
+
+
+class AdaScaleWrapper(AdaScale):
+    """
+    A thin wrapper for AdaScale so that the constructor resembles a
+    standard optimizer. This allows it to work with other Optimizer
+    Wrappers, like `OSS`.
+
+    .. warn::
+        OSS(AdaScaleWrapper) (i.e. OSS wrapping AdaScale) resulting in each
```
anj-s, line 599: What is the benefit of having this work? Why not always require Adascale(OSS)?

tests/optim/test_oss_adascale.py
```diff
@@ -98,6 +104,14 @@ def test_sequential(oss):
         "expected_mean_weight": 52.92657661437988,
     }
 
+    if oss == "oss-wrapper":
+        # When OSS wraps AdaScale, the training is neumerically different
```
anj-s, line 108: typo:numerically

fairscale/optim/adascale.py
```diff
@@ -582,3 +587,47 @@ def set_num_gradients_to_accumulate(self, num_gradients_to_accumulate: int, upda
             # When effective world size is large enough, smoothing is probably
             # not needed, so the smoothing factor is 0.
             self._smoothing = max(1 - self._world_size * self._num_grads_to_accum / 1000, 0)
+
+
+class AdaScaleWrapper(AdaScale):
```
anj-s, line 592: should this be in an experimental folder since we are still validating this approach?

fairscale/optim/adascale.py
min-xu-ai: This allows future study for this combination. Even with only AdaScale(OSS), this wrapper allows a different way of initialization. In some cases, that might be easier. what do you think?

fairscale/optim/adascale.py
min-xu-ai: only OSS(AdaScale) is experimental. The wrapper can totally be used for AdaScale(OSS), just a different way of initialization.

anj-s (approved):

tests/optim/test_oss_adascale.py
min-xu-ai: fixed.

mikerabbat:
> > > Why is a separate wrapper class necessary? Why can't we just change AdaScale directly to work as a wrapper?
> 
> I should add that `AdaScale(OSS)` already works. `OSS(AdaScale)` is what this is trying to address.

Just curious, if `AdaScale(OSS)` already works, do we also need to support `OSS(AdaScale)`?

min-xu-ai:
> 
> 
> > > > Why is a separate wrapper class necessary? Why can't we just change AdaScale directly to work as a wrapper?
> > 
> > 
> > I should add that `AdaScale(OSS)` already works. `OSS(AdaScale)` is what this is trying to address.
> 
> Just curious, if `AdaScale(OSS)` already works, do we also need to support `OSS(AdaScale)`?

We don't have to. It is marked as experimental. In case we need to do some research in that direction in the future.

The wrapper is to allow some flexibility in terms of wrapping an underlying optimizer, just in case some trainer loop needs that form to be used. Does it sound OK to you?

mikerabbat:
Yes, of course, sounds good! Just wasn't sure if there was already a compelling use case :-)

min-xu-ai:
> 
> 
> Yes, of course, sounds good! Just wasn't sure if there was already a compelling use case :-)
totally. a very good question! Thanks for reviewing.

PR #350, opened 2021-02-01
[fix] ShardedDDP - cpu testfix - remove Gloo/CPU

blefaudeux:
# Before submitting

## What does this PR do?
Fixes intermittent issues with newer python versions, I could not reproduce locally. I don't think that cpu use is super important really so not loosing much

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/utils/testing.py
+++ b/fairscale/utils/testing.py
@@ -31,6 +31,7 @@ import logging
 import multiprocessing
 import os
 import random
+import sys
 import tempfile
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
@@ -52,6 +53,8 @@ skip_if_no_cuda = pytest.mark.skipif(
 skip_if_single_gpu = pytest.mark.skipif(
     not torch.cuda.is_available() or torch.cuda.device_count() < 2, reason="multiple GPUs required"
 )
+
+skip_if_py38 = pytest.mark.skipif(sys.version_info.major == 3 and sys.version_info.minor == 8)
 
 _, filename_mpi = tempfile.mkstemp()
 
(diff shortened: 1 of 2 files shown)


min-xu-ai:
is it only failing on py38? Is so, perhaps skip it only on py38? CPU test can still be useful for testing thing when multiple GPUs are not available. What do you think?

blefaudeux:
> is it only failing on py38? Is so, perhaps skip it only on py38? CPU test can still be useful for testing thing when multiple GPUs are not available. What do you think?

it seems to be only failing on py3.8, only for cpu only installs for some reason (the same test was run, also on cpu, on gpu enabled hosts and never had any problem, see for instance https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1504/workflows/9859ca20-6552-44e2-befa-642eb07f4a8e. It only randomly fails on the new cpu-only tests, and I can't even repro it locally, I guess it's tied to the specific torch that we install). I'm sure it can be investigated more but it takes a lot of time, I'm not sure that it's worth it right now. Besides, how would you only skip this test on py38 ? the job is not really visible from the python codebase

min-xu-ai:
skipping based on version should be simple. I had it at one point but never merged. I did the following:

add skip_if_py38 in testing.py:

it is something like:

import sys

if sys.version_info.major ==3 and sys.version_info.minor == 8: pytest.skip()

then, just use @skip_if_py38 to annotate the test.

min-xu-ai:
another way is to skip the assert model differ between ranks when it is running with a specific py version. I think that should minimal the overhead of us debugging it but still keep some portion of the code covered in different versions.

blefaudeux:
> sys.version_info.major ==3 and sys.version_info.minor == 8

something similar done, I would not have guessed that it was a preferred solution but that's completely fine by me

min-xu-ai (approved):
awesome!

PR #352, opened 2021-02-02
[feat] OSS Benchmark - regression test + background testing new optims

blefaudeux:
# Before submitting

## What does this PR do?
Fixes the regression not being run anymore, and adding a background test for the new multi-tensor optims (just checking that it runs and have an idea of the perf difference for now)

## PR review
cc @izdeby

## Did you have fun?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ run_oss_benchmark: &run_oss_benchmark
       name: Run OSS Benchmark
       command: |
         python benchmarks/oss.py --world_size 4 --epochs 2
-        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp --reference_speed 660 --reference_memory 1000 --reference_loss 0.026
+        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp
 
 run_oss_gloo: &run_oss_gloo
   - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ run_oss_benchmark: &run_oss_benchmark
       name: Run OSS Benchmark
       command: |
         python benchmarks/oss.py --world_size 4 --epochs 2
-        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp --reference_speed 660 --reference_memory 930 --reference_loss 0.023
+        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp --reference_speed 660 --reference_memory 1000 --reference_loss 0.026
 
 run_oss_gloo: &run_oss_gloo
   - run:
@@ -187,6 +187,12 @@ run_oss_amp: &run_oss_amp
        name: Run OSS with Torch AMP
        command: |
          python benchmarks/oss.py --amp --epochs 3 --optim_type oss_sharded_ddp
+
+run_oss_for_each: &run_oss_for_each
+   - run:
+       name: Run OSS with Torch AMP and ForEach optmizer
+       command: |
+         python benchmarks/oss.py --amp --epochs 3 --optim_type oss_sharded_ddp --multi_tensor_optim
 
 
 run_doc_build: &run_doc_build
@@ -450,6 +456,7 @@ jobs:
 
       - <<: *run_oss_amp
 
+      - <<: *run_oss_for_each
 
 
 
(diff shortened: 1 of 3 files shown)


benchmarks/golden_configs/oss_mnist.py
```diff
@@ -4,9 +4,9 @@
 def get_golden_real_stats():
 
     return {
-        "reference_speed": 1430,
-        "reference_memory": 1220,
-        "reference_loss": 0.006,
+        "reference_speed": 660,
```
blefaudeux, line 7: the numbers left must have been on a devfair, the number right are the ones we used on circleci previously

.circleci/config.yml
```diff
@@ -173,7 +173,7 @@ run_oss_benchmark: &run_oss_benchmark
       name: Run OSS Benchmark
       command: |
         python benchmarks/oss.py --world_size 4 --epochs 2
-        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp --reference_speed 660 --reference_memory 930 --reference_loss 0.023
+        python benchmarks/oss.py --check_regression --world_size 4 --optim_type oss_sharded_ddp
```
blefaudeux, line 176: the flags were not used actually, the "golden config" is being used

benchmarks/oss.py
```diff
@@ -273,9 +276,6 @@ def run_closure(closure, scaler, optimizer):
     parser.add_argument("--epochs", action="store", default=10, type=int)
     parser.add_argument("--batch_size", action="store", default=256, type=int)
     parser.add_argument("--check_regression", action="store_true", default=False)
-    parser.add_argument("--reference_speed", action="store", default=1430, type=float)
```
blefaudeux, line 276: not used anymore anyway, not changed by this PR

blefaudeux:
The numbers are slightly different for now with the new optims, being fixed with https://github.com/pytorch/pytorch/pull/48223. End plan is even faster and complete 1:1 bit parity, so I think that we could start recommending them

msbaines (approved):

benchmarks/oss.py
```diff
@@ -118,6 +117,10 @@ def train(
 ):
     logging.basicConfig(level=logging.INFO if not args.debug else logging.DEBUG)
 
+    use_multi_tensor = args.multi_tensor_optim and hasattr(torch.optim, "_multi_tensor")
+    OPTIM = torch.optim._multi_tensor.RMSprop if use_multi_tensor else torch.optim.RMSprop  # type: ignore  # attr is  checked but mypy misses that
```
anj-s, line 121: Have we tested this with OSS? I couldn't find an example.

.circleci/config.yml
```diff
@@ -188,6 +188,12 @@ run_oss_amp: &run_oss_amp
        command: |
          python benchmarks/oss.py --amp --epochs 3 --optim_type oss_sharded_ddp
 
+run_oss_for_each: &run_oss_for_each
+   - run:
+       name: Run OSS with Torch AMP and ForEach optmizer
+       command: |
+         python benchmarks/oss.py --amp --epochs 3 --optim_type oss_sharded_ddp --multi_tensor_optim
```
anj-s, line 195: Why do we want to add a regression benchmark? Would tests suffice?

blefaudeux: in that case it's not a regression, it just checks that it runs ?

benchmarks/oss.py
blefaudeux: no, sorry if I was not clear, the whole point of this PR is to test that :)

anj-s: I guess I was confused by the CL description which said adding tests for new optims but I guess you meant regression benchmarks. 
 If you want to test the new optims, you could do a golden weight check which is a different test than the regression benchmark you are adding. I am not sure if we need to add a regression benchmark (mainly because it adds to the time taken to run all checks on GH). Given I don't have context about this use case, I'll leave you to decide :)

.circleci/config.yml
anj-s: Replied below: You can do a golden weight comparison check which makes sure that the results match with the baseline optimizer (if that is what you want)

blefaudeux: no so there are two things in this PR, not completely related:
- restore the regression test for ShardedDDP which was broken
- start a test job with these new optimizers. Right now they are not 1:1 correct, it's being changed as I write so it should soon be the case, I thought that having a test job on the topic could be useful to check parity and catch early any eventual issues

blefaudeux:
@anj-s ok to land ? I can change something if you'd like

anj-s (approved):

PR #356, opened 2021-02-03
[chore] disheartening switch off of a OSS cpu test

blefaudeux:
# Before submitting

## What does this PR do?
This test runs on CPU, and typically never fails, on agents which are cuda enabled (it does use Gloo and device==cpu). The newer jobs which don't have cuda randomly fail on this very same test with python3.9, even if to the best of my knowledge nothing should change. This test remains active for gpu enabled agents, even if 

## PR review

## Did you have fun?

### DIFF
--- a/tests/optim/test_oss.py
+++ b/tests/optim/test_oss.py
@@ -22,7 +22,7 @@ import torch.multiprocessing as mp
 from torch.nn.parallel import DistributedDataParallel as DDP
 
 import fairscale.optim as optim
-from fairscale.utils.testing import skip_if_no_cuda, skip_if_single_gpu
+from fairscale.utils.testing import skip_if_no_cuda, skip_if_py39, skip_if_single_gpu
 
 BACKEND = dist.Backend.NCCL if torch.cuda.is_available() else dist.Backend.GLOO  # type: ignore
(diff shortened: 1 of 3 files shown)


tests/optim/test_oss.py
```diff
@@ -564,6 +564,7 @@ def closure():
     dist.destroy_process_group(process_group)
 
 
+@skip_if_py39_no_cuda
 def test_multiple_groups():
```
blefaudeux, line 568: the failure case is specifically on agents which do not have cuda, even if this should not matter to this specific test (gloo and cpu).

blefaudeux:
see https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1533/workflows/9fbae0c7-2586-4688-86fa-306e1aaf21ce/jobs/7202 (and the rest of the history, one step up) to have an idea of the flakiness of this test

min-xu-ai:
thanks for including the details. it is sad, but sure.

min-xu-ai (approved):

blefaudeux:
> thanks for including the details. it is sad, but sure.

I think it would be nice to debug eventually, but we're a bit short handed for that, and it's not really on the critical path to mission, so compromising..

PR #371, opened 2021-02-06
[refactor] OSS only use flat buffers

blefaudeux:
# Before submitting
- [x]  Did you make sure to update the docs?

## What does this PR do?
Just a natural extension of the params being tensor views of a bigger tensor (#300), no need to keep the non-bucketed params, I feel a little dumb but it's just much simpler and just as fast to have one single buffer per device and per rank (or faster for multi node, less latency). 

So this PR just does that, all the trainable params which belong to the same device/rank link are packed into a single tensor, like before but not bounded. There's no tradeoff really, the original parameter memory is released.

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -538,7 +538,6 @@ class OSS(Optimizer):
         last_work_handle = None  # Work handles are consumed within this scope, no callback
 
         for device in self.per_device_params.keys():
-            # Bucket and issue all the async calls
             for src_rank, bucket in enumerate(self.buckets[device]):
                 global_src_rank = self.get_global_rank(self.group, src_rank)
                 last_work_handle = dist.broadcast(tensor=bucket, src=global_src_rank, group=self.group, async_op=True)
@@ -568,12 +567,10 @@ class OSS(Optimizer):
         """Make all params which are on the same device and tied to the same rank views of a single buffer
         """
 
-        # Devise the bucketing strategy
         for device, per_rank_params in self.per_device_params.items():
             self.buckets[device] = []
 
             for dst_rank, params in enumerate(per_rank_params):
-
                 buffer_size = sum(map(lambda x: x.numel(), filter(lambda x: x.requires_grad, params)))
                 self.buckets[device].append(torch.zeros(buffer_size, dtype=params[0].dtype, device=device))
                 offset = 0
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -3,19 +3,19 @@
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
 
-from collections import OrderedDict, deque
+from collections import OrderedDict
 import copy
 from itertools import chain
 import logging
 from math import inf
-from typing import TYPE_CHECKING, Any, Callable, Deque, Dict, List, Optional, Type, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union
 
 import torch
 import torch.distributed as dist
 from torch.nn import Parameter
 from torch.optim import SGD, Optimizer
 
-from .utils import Workhandle, broadcast_object, recursive_copy_to_device
+from .utils import broadcast_object, recursive_copy_to_device
 
 __all__ = ["OSS"]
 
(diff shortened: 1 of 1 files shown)


fairscale/optim/oss.py
```diff
@@ -51,9 +51,7 @@ class OSS(Optimizer):
         group (group):
             torch.distributed group (default: group.WORLD)
         broadcast_buffer_size (int):
-            the max size of the buffer used to batch the small parameter tensors, in number of elements (default 16M).
-            this will not impact the long term memory consumption, but the peak memory can be impacted by the moment
-            when the buffers are allocated and the bucketed params have not yet been relocated to them.
+            (deprecated) used to cap the size of the broadcast buffers, not being used anymore.
```
min-xu-ai, line 54: not being used -> not being used because broadcast buffer not long exists?

fairscale/optim/oss.py
```diff
@@ -583,51 +563,23 @@ def _try_consume_work_handle(self) -> None:
...
 
-    def _setup_bucket_strategy(self) -> None:
-        """Tag parameters to either bucket them or broadcast/reduce them directly. The parameters are ordered
-        (smallest first), the bucket will hold the smallest elements, the remaining ones will be directly sent
-        over the wire.
-
-        Generating the partition once and for all allows us to save some time at runtime, and to know when all the
-        network requests have been issued.
+    def _setup_flat_buffers(self) -> None:
+        """Make all params which are on the same device and tied to the same rank views of a single buffer
```
min-xu-ai, line 567: mention that whether this is called only during init time or not?

min-xu-ai (approved):
seems like a really nice simplification and reduces the memory use.

fairscale/optim/oss.py
```diff
@@ -583,51 +563,23 @@ def _try_consume_work_handle(self) -> None:
...
 
-                for param in params:
-                    # Criteria to decide whether this parameter is to be bucketed or not:
-                    # - enough room in the bucket
-                    if param.requires_grad and (offset + param.numel()) < self.bucket_size:
-                        self.should_bucket_param.append(True)
+            for dst_rank, params in enumerate(per_rank_params):
+                if len(params) > 0:
+                    buffer_size = sum(map(lambda x: x.numel(), filter(lambda x: x.requires_grad, params)))
+                    self.buckets[device].append(torch.zeros(buffer_size, dtype=params[0].dtype, device=device))
```
msbaines, line 576:
```suggestion
                    self.buckets[device].append(torch.empty(buffer_size, dtype=params[0].dtype, device=device))
```

fairscale/optim/oss.py
```diff
@@ -583,51 +563,23 @@ def _try_consume_work_handle(self) -> None:
...
+            self.buckets[device] = []
 
-                for param in params:
-                    # Criteria to decide whether this parameter is to be bucketed or not:
-                    # - enough room in the bucket
-                    if param.requires_grad and (offset + param.numel()) < self.bucket_size:
-                        self.should_bucket_param.append(True)
+            for dst_rank, params in enumerate(per_rank_params):
+                if len(params) > 0:
+                    buffer_size = sum(map(lambda x: x.numel(), filter(lambda x: x.requires_grad, params)))
```
msbaines, line 575: You are running the same filter twice. Cache the result and use it down below?

msbaines (approved):
Nice! Some minor nits (non-blocking).

fairscale/optim/oss.py
blefaudeux: basically all the parameters tied to a given recipient are now one single tensor (+tensor views so that the FW pass does not see that), this setting is not respected anymore. I should have done that from the moment I realized that tensor views were an option (I thought initially that it was only on recent pytorch, turns out it was already there in 1.5), but at the time I got a bit stuck in the mental model for the reduce, where the buckets are more of a compromise right now (it's more complicated or not possible to release them -opposed to gradient sharding-, and they hold the whole BW back instead of overlapping coms/compute). 
In this case (optimizer step + broadcast the results) there's no reason not to batch *all* the parameters in one go (limited by device/ranks where they go), not that I can see at least, so this option is not useful anymore because the buffers are not capped (no duplicated memory, the original param storage is released), I let it there for backward compatibility only

fairscale/optim/oss.py
```diff
@@ -101,12 +104,17 @@ def __init__(
...
         self._device = list(self.per_device_params.keys())[0]
+        self.work_handles: Deque[Workhandle] = deque()
         self.buckets: Dict[torch.device, List[torch.Tensor]] = {}
-        self.buffer_max_size = broadcast_buffer_size
+        self._setup_flat_buffers()
 
-        self.should_bucket_param: List[bool] = []
-        self.work_handles: Deque[Workhandle] = deque()
-        self._setup_bucket_strategy()
+    def refresh_trainability(self) -> None:
```
blefaudeux, line 111: Open question @msbaines and @min-xu-ai , I just added these two lines (and a warning in the class docstring), related to https://github.com/facebookresearch/fairscale/pull/369, usecase which popped up for VISSL: freeze/unfreeze some parameters on the fly.
- what do you think of this option (if you changed the graph, call this refresh function, opt-in basically) ? still ok for you ? Not landing before a reaction here :)
- how does Pipe work with that, do you reevalue the graph all the time or is it frozen, or a solution like this one ?

blefaudeux: removing that, better to keep it for another PR if need be (see discussion on #369)

PR #376, opened 2021-02-10
Add fairscale.nn.misc.checkpoint_activations

myleott:
A friendlier wrapper for performing activation checkpointing.

Compared to the PyTorch version, this version:
- wraps an nn.Module, so that all subsequent calls will use checkpointing
- handles keyword arguments in the forward
- handles non-Tensor outputs from the forward
- supports offloading activations to CPU

Usage:
```python
checkpointed_module = checkpoint_wrapper(my_module, offload_to_cpu=True)
a, b = checkpointed_module(x, y=3, z=torch.Tensor([1]))
```

### DIFF
--- a/fairscale/nn/misc/checkpoint_activations.py
+++ b/fairscale/nn/misc/checkpoint_activations.py
@@ -1,0 +1,158 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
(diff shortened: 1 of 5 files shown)


blefaudeux:
It's really great, thanks @myleott ! I think we could add some doc around that to make sure it's exposed and known, I can do that in the coming days if you'd like

min-xu-ai (approved):
This looks great! Second Ben's comments too.

PR #383, opened 2021-02-12
[fix] oss dict load fix

### LINKED ISSUE #380: [OSS] When loading a state, the insertion order for the params may not match the params_groups
## 🐛 Bug
(FB Only f250383189) Guess is that the issue comes from the parameters not being inserted in the right order in the sharded optimizer state. [See](https://github.com/facebookresearch/fairscale/blob/master/fairscale/optim/oss.py#L403)

Would need fixing + repro in a unit test

blefaudeux:
# Before submitting

## What does this PR do?
Fix for an issue that Pyspeech was seeing, it looks like I broke the logic in the last updates of  #310
The unit test did not catch that because the model was symmetrical enough, tryign to find a better test 
The bug detection and fix suggestion is from Weiyi Zheng @zhengwy888

Fixes #380

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -395,21 +395,40 @@ class OSS(Optimizer):
         for i_param, (key, value) in enumerate(state_dict["state"].items()):
             param = self.index_to_param[i_param]
 
-            # Populate the sharded optimizer state on the fly
+            # If this rank does not own this param, eagerly wipe the state
             if self.param_to_rank[param] != self.rank:
                 state_dict["state"][key] = None
 
-            if key in self.index_to_param:
+        # Populate the base param_groups and state dict
+        super().load_state_dict(state_dict)
+
+        # If this rank does not own this param, eagerly wipe the state
+        delete_keys = []
+        for i_param, key in enumerate(state_dict["state"].keys()):
+            param = self.index_to_param[i_param]
+            if self.param_to_rank[param] != self.rank:
+                delete_keys.append(key)
+
+        for key in delete_keys:
+            del state_dict["state"][key]
+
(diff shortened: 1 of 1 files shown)


fairscale/optim/oss.py
```diff
@@ -391,16 +391,16 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
-        for i_param, (key, value) in enumerate(state_dict["state"].items()):
-            param = self.index_to_param[i_param]
+        for key, value in state_dict["state"].items():
+            param = self.index_to_param[pytorch15_index_redirect[key]]
 
             # Populate the sharded optimizer state on the fly
             if self.param_to_rank[param] != self.rank:
                 state_dict["state"][key] = None
 
-            if key in self.index_to_param:
```
blefaudeux, line 402: this test was definitely wrong, I should have seen that..

fairscale/optim/oss.py
```diff
@@ -391,16 +391,16 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
 
         # NOTE: PyTorch 1.5 does not index linearly but with the id(params) at saving time
         # we work around that here by using the fact that the params are ordered as in the param_groups
+        pytorch15_index_redirect = {k: i for i, k in enumerate(state_dict["state"].keys())}
```
blefaudeux, line 394: this is useless post pytorch 1.5

blefaudeux:
context: Planning for a follow up PR to find a unit test which exposes the previous bug (guess is that the model used was too symmetrical), but trying to fix master ASAP in the meantime

min-xu-ai (approved):
nice!

PR #386, opened 2021-02-13
[fix] OSS dict load/save fix - better fix than 383 and unit test

### LINKED ISSUE #380: [OSS] When loading a state, the insertion order for the params may not match the params_groups

blefaudeux:
# Before submitting

- [x] Did you write any new necessary tests? Catch the case which was broken before, and fix it

## What does this PR do?
Fixes #380. Better take than #383 because fixing another issue which was not caught (383 was not enough), and reproducing the issue in an updated unit test so that this does not happen again. Thanks again @zhengwy888

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -402,12 +402,7 @@ class OSS(Optimizer):
             if self.param_to_rank[param] != self.rank:
                 state_dict["state"][key] = None
             else:
-                # Only add this state to the sharded optimizer if it owns this param
-                for pg in self.optim.param_groups:
-                    if id(param) in [id(p) for p in pg["params"]]:
-                        self.optim.state[param] = recursive_copy_to_device(
-                            value, non_blocking=True, device=param.device
-                        )
+                self.optim.state[param] = recursive_copy_to_device(value, non_blocking=True, device=param.device)
 
         super().load_state_dict(state_dict)
 
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -395,21 +395,40 @@ class OSS(Optimizer):
         for i_param, (key, value) in enumerate(state_dict["state"].items()):
             param = self.index_to_param[i_param]
 
-            # Populate the sharded optimizer state on the fly
+            # If this rank does not own this param, eagerly wipe the state
             if self.param_to_rank[param] != self.rank:
                 state_dict["state"][key] = None
 
-            if key in self.index_to_param:
+        # Populate the base param_groups and state dict
+        super().load_state_dict(state_dict)
+
+        # If this rank does not own this param, eagerly wipe the state
+        delete_keys = []
+        for i_param, key in enumerate(state_dict["state"].keys()):
+            param = self.index_to_param[i_param]
+            if self.param_to_rank[param] != self.rank:
+                delete_keys.append(key)
+
+        for key in delete_keys:
+            del state_dict["state"][key]
+
+        # Remove the params that the shard optim does not own, and re-index
+        # TODO: Ben
+        for state_pg, pg, spg in zip(state_dict["param_groups"], self.param_groups, self.optim.param_groups):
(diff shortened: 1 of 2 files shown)


tests/optim/test_oss.py
```diff
@@ -859,10 +861,16 @@ def closure_sharded(input_tensor=input_tensor):
         sharded_optim_state_dict = sync_object_ranks(sharded_optim_state_dict, RECIPIENT_RANK, device)
 
         # - cross load the states
+        # run one step and check that the models are still the same
```
blefaudeux, line 864: this modified unit test does break on the old version

fairscale/optim/oss.py
```diff
@@ -379,6 +379,8 @@ def state_dict(self) -> Dict[str, Any]:
                         global_id = self.param_to_index[local_index_to_param_id[local_param_index]]
                         state_dict["state"][global_id] = s["state"][local_param_index]
 
+        # Make sure that the parameters are sorted in the state, as expected
+        state_dict["state"] = dict(sorted(state_dict["state"].items()))
```
blefaudeux, line 383: the state dict returned was sorted properly under the "param_groups" key, but not under the "state" field. Pytorch just uses the ordering from the "param_groups" key, and I was just testing the loading OSS-> Pytorch and vice versa, so t his was not caught

fairscale/optim/oss.py
```diff
@@ -391,23 +393,16 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
             # Populate the sharded optimizer state on the fly
             if self.param_to_rank[param] != self.rank:
                 state_dict["state"][key] = None
-
-            if key in self.index_to_param:
-                param = self.index_to_param[i_param]
-
-                # Only add this state to the sharded optimizer if it owns this param
-                for pg in self.optim.param_groups:
-                    if id(param) in [id(p) for p in pg["params"]]:
```
blefaudeux, line 407: this second check could mask an issue, we just checked above that this rank owns this param, so this is not needed (and potentially risky)

tests/optim/test_oss.py
```diff
@@ -832,8 +834,8 @@ def closure_sharded(input_tensor=input_tensor):
             loss_sharded_optim = cast(torch.Tensor, sharded_optimizer.step(closure=closure_sharded))
 
             assert torch.allclose(
-                loss_ddp, loss_sharded_optim
-            ), f"Losses differ in between Pytorch optim and OSS\nworld size {world_size}"
+                loss_ddp, loss_sharded_optim, rtol=1e-3
```
blefaudeux, line 837: the rtol change is only needed on pytorch 1.5 unfortunately, without that on a two gpu machine the difference becomes 
26.404895782470703 vs 26.404342651367188 (which I assume is due to a different casting and not structurally wrong) and this asserts

blefaudeux:
sorry @min-xu-ai for the revert of the previous one, I just thought this was cleaner and there was one fix left in the cold

tests/optim/test_oss.py
min-xu-ai: I think it is worth documenting the reason for 1e-3.

min-xu-ai (approved):
yeah, this seems to be much nicer.

fairscale/optim/oss.py
zhengwy888: I didn't know python's dictionary is ordered, so I just looked it up. turns out this has been enabled since python 3.5, good to know! https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6

Issue #398, opened 2021-02-18
[CI] Unit test ShardedDDP with buckets


blefaudeux:
## 🚀 Feature
Unit test properly ShardedDDP with and without buckets being activated 

## Motivation
It seems that I shipped a broken version for shardedddp wrt buckets (should now be fixed), this situation was not properly tested given the fact that for small world sizes the buckets are not used, and we test on small world sizes.

## Pitch
- expose buckets being used or not explicitly
- test both cases

## Alternatives
Shipping broken things

Issue #397, opened 2021-02-18
ShardedDataParallel  doesn't work with multiple nodes


kamo-naoyuki:
## 🐛 Bug

ShardedDataParallel successfully works with 8gpus x 1nodes, while got the following error with 8gpus x 2 nodes 

```
AssertionError: A bucket failed to be sent, probably unused parameters.Either remove the unused parameter or de-activate ShardedDDP buckets -set reduce_buffer_size to 0-
```

However, obviously there are not unused parameters. Note that torch.nn.DistributedDataParallel can work with same environment.

## To Reproduce

```python
import os
import sys 

import torch
import torch.distributed as dist
import torch.multiprocessing as mp
from fairscale.optim.oss import OSS 
from fairscale.nn.data_parallel import ShardedDataParallel as ShardedDDP


def train(rank, local_rank, world_size, init_method):
    print("DDP init", world_size, rank, local_rank, init_method, file=sys.stderr)
    dist.init_process_group(backend='nccl', init_method=init_method, rank=rank, world_size=world_size)
    torch.cuda.set_device(local_rank)

    model = torch.nn.Linear(3, 3).cuda()
    base_optimizer = torch.optim.Adam
    base_optimizer_arguments = {}

    optimizer = OSS(params=model.parameters(), optim=base_optimizer, **base_optimizer_arguments)
    model = ShardedDDP(model, optimizer)

    print("Train", file=sys.stderr)
    model.train()
    model.zero_grad()
    outputs = model(torch.randn(2,3).cuda())
    loss = outputs.sum()
    loss.backward()
    optimizer.step()
    print("finish", file=sys.stderr)
```

I used shared file system initialization for init_method.

```
init_method="file:..."
```

## Environment

Note that I tested tcp and infiniband connection both.

```
PyTorch version: 1.7.1
Is debug build: False
CUDA used to build PyTorch: 11.0
ROCM used to build PyTorch: N/A

OS: CentOS Linux 7 (Core) (x86_64)
GCC version: (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
Clang version: Could not collect
CMake version: version 2.8.12.2

Python version: 3.8 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: 
GPU 0: Tesla V100-SXM2-32GB
GPU 1: Tesla V100-SXM2-32GB
GPU 2: Tesla V100-SXM2-32GB
[...]

Issue #395, opened 2021-02-18
Why assuming single node environment if world_size<=8 for ShardedDataParallel?


kamo-naoyuki:
## ❓ Questions and Help

Thank you for sharing this project. Really useful.

I tried ShardedDataParallel with 1gpus x 2nodes, i.e. world_size=2, but I met the following.

https://github.com/facebookresearch/fairscale/blob/47042917536dfe3026543ccebd789ef3b1441ba1/fairscale/nn/data_parallel/sharded_ddp.py#L156-L157

Why do you assume single node environment if world_size<=8? Is it impossible if world_size<=8 with multiple nodes?

blefaudeux:
Hi Naoyuki, good question !
In that case, "nodes" means server, meaning that you typically have up to 8 GPUs on a single node (for instance nvidia DGX), it does not mean world size (world size = gpu per node x nodes). You can have 8 GPUs over two nodes, and in that case this heuristic will not be optimal, I was planning to expose it in an upcoming PR actually, in the FB ecosystem it's not really typical.

Now, why do we take this into account to begin with ? The communications inside a node are typically very fast, nvlink or pcie, while having to go outside of a node is typically slow. The buckets here are used to group small parameters together, they diminish the latency cost but they have a memory cost (could be improved in the future). So it's a tradeoff, and an initial safe bet was to assume that 8 gpus or less was a single node, so that the buckets would not help.

In your case if you're testing with 2 gpus I'm guessing that they are on a single machine, so you probably don't need buckets to efficiently reduce the parameters :) I hope that helps

kamo-naoyuki:
Thank you for quick.  Sorry, my question is unclear.

I mean node is sever, so I want to say 1gpus x 2machine, really.

Not always each machine has 8gpus, though I agree it is popular setup. I think this action is not necessary because user can set reduce_buffer_size=0 by themselves if it is a single node. Why automatically change this value?

blefaudeux:
> Thank you for quick. Sorry, my question is unclear.
> 
> I mean node is sever, so I want to say 1gpus x 2machine, really.
> 
> Not always each machine has 8gpus, though I agree it is popular setup. I think this action is not necessary because user can set reduce_buffer_size=0 by themselves if it is a single node. Why automatically change this value?

most people do not touch this value, so the idea was to try to have sane defaults. I'll put up a PR today to expose this toggle properly

PR #400, opened 2021-02-18
[feature] Unit test with and without buckets for all ShardedDDP unit tests

blefaudeux:
# Before submitting

## What does this PR do?
Too many papercuts (or full fledged bugs) slipped through recently, and I realized that the unit tests were not covering all combinations. The module can flip from train to test, be under a no_sync context, be using AMP, trainability can change on a per-parameter basis, and buckets can be used or not.

This PR adds buckets yes/no to all tests but one, including the ddp_parity test which covers most of the matrix above. It fixes a bug which went unseen on that field (not a silent bug though, as it would assert), when leaving a no_sync context and using buckets. It fixes another bug which went unseen in handling several groups while using buckets. 

Fixes https://github.com/facebookresearch/fairscale/issues/398
Makes sure that https://github.com/facebookresearch/fairscale/issues/397 does not happen again

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -51,9 +51,9 @@ class ShardedDataParallel(nn.Module):
             Synchronize the models in between the ranks when starting up. Not needed if each rank has the same seed,
             or the training restarts from a saved state
         reduce_buffer_size (int):
-            The max size of the buffer used to batch the small parameter tensors, in number of elements (default 8M).
+            The max size of the buffer used to batch the small parameter tensors, in number of elements (default 0 - unused).
             this will impact the long term memory consumption, because these buckets correspond to parameters which will not be sharded.
-            Set to 0 to remove all bucketing.
+            Set to 0 to remove all bucketing, 1M to 8M is usually reasonable.
         auto_refresh_trainable (bool):
             (default: True) Check whether the parameters trainability (`requires_grad`) has changed and update both ShardedDDP
             and OSS automatically if this is the case. If set to False, `refresh_trainable()` needs to be called anytime
@@ -98,7 +98,7 @@ class ShardedDataParallel(nn.Module):
         process_group: Any = None,
         broadcast_buffers: bool = True,
         sync_models_at_startup: bool = True,
-        reduce_buffer_size: int = 2 ** 23,
+        reduce_buffer_size: int = 0,
         auto_refresh_trainable: bool = True,
     ):
         super().__init__()
@@ -111,6 +111,7 @@ class ShardedDataParallel(nn.Module):
         # Handle a no_sync() context which prevents the gradient synchronization,
         # accumulate in place
         self.should_accumulate_grads = False
+        self.accumulate_grads_flipped = False
 
         # Communication related attributes
         self.process_group = process_group if process_group is not None else dist.group.WORLD
@@ -153,10 +154,6 @@ class ShardedDataParallel(nn.Module):
 
         # - setup buckets and tensor views
         model_size = sum([p.numel() for p in self.module.parameters()])
-        if dist.get_world_size(self.process_group) <= 8:
-            logging.info("Assuming single node environment. De-activating ShardedDDP buckets")
-            reduce_buffer_size = 0
-
         self.buffer_max_size = min(reduce_buffer_size, model_size)
         logging.info(
             "ShardedDDP bucket size: {:.2f}M parameters, model size {:.2f}M parameters".format(
@@ -206,6 +203,9 @@ class ShardedDataParallel(nn.Module):
         # Reset all the grad reduce and bucket state flags
         self._clear_counters()
 
+        # Failsafe, catch buckets which have not been sent, will not catch all cases
+        self._attach_failsafe_hook_to_inputs(*inputs)
+
         # Normal FW on the base model
         return self.module(*inputs, **kwargs)
 
@@ -230,6 +230,11 @@ class ShardedDataParallel(nn.Module):
         .. note::
             This method modifies the module in-place.
 
+        .. warning:
+            Device changes are not supported, and this will raise an exception. The issue in that case is not
+            really ShardedDDP, but OSS which will not be aware of the device change, and whose buffers will be
+            in a broken state.
+
         Arguments:
             device (:class:`torch.device`): the desired device of the parameters and buffers in this module.
             dtype (:class:`torch.dtype`): the desired floating point type of the floating point parameters and buffers.
@@ -237,14 +242,18 @@ class ShardedDataParallel(nn.Module):
 
         Returns:
             Module: self.
-
         """
 
-        for device in self.buckets.keys():
-            for bucket in self.buckets[device]:
+        assert device in self.buckets.keys(), "Changing devices is not supported, because this would break OSSs state"
+        assert (
+            len(self.buckets.keys()) == 1
+        ), "Several devices specified to begin with, incompatible with setting a single device here"
+
+        for _device in self.buckets.keys():
+            for bucket in self.buckets[_device]:
                 bucket.buffer.to(device=device, dtype=dtype, non_blocking=non_blocking)
 
-        self.module.to(device)
+        self.module.to(device=device, dtype=dtype, non_blocking=non_blocking)
 
     def refresh_trainable(self) -> None:
         """ If the module trainability has changed, update all the assumptions """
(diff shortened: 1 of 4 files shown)


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -153,10 +154,6 @@ def __init__(
 
         # - setup buckets and tensor views
         model_size = sum([p.numel() for p in self.module.parameters()])
-        if dist.get_world_size(self.process_group) <= 8:
```
blefaudeux, line 156: this proved polemical (see https://github.com/facebookresearch/fairscale/issues/395, among others), and a lot of people at FB also did not understand why things behaved differently on one or more nodes, so I guess that it's a good signal

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -98,7 +98,7 @@ def __init__(
         process_group: Any = None,
         broadcast_buffers: bool = True,
         sync_models_at_startup: bool = True,
-        reduce_buffer_size: int = 2 ** 23,
+        reduce_buffer_size: int = 0,
```
blefaudeux, line 101: failsafe default (does not care about parameters not being used)

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -230,21 +230,30 @@ def to(  # type: ignore
...
 
         Returns:
             Module: self.
-
         """
 
-        for device in self.buckets.keys():
-            for bucket in self.buckets[device]:
+        assert device in self.buckets.keys(), "Changing devices is not supported, because this would break OSSs state"
+        assert (
```
blefaudeux, line 248: changing the tests to make sure that all the combinations were covered (accumulation/amp/manual reduce/buckets) exposed the fact that this was broken, but on the OSS side. OSS remaps the parameters as views of tensors, but it does not see the .to() calls (it's not exposed as a module), so when the module is moved to another device OSS is in a broken state. Asserting here so that this does not go unseen

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -339,6 +349,7 @@ def no_sync(self) -> Generator:
         old_should_accumulate_grads = self.should_accumulate_grads
         self.should_accumulate_grads = True
         yield
+        self.accumulate_grads_flipped = self.should_accumulate_grads != old_should_accumulate_grads
```
blefaudeux, line 352: again, testing each and every combination exposed a broken case when we just went out of the no_sync context (the grads are not empty, but that's expected), which this fixes

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -604,3 +598,46 @@ def _try_consume_work_handle(self) -> None:
             work_handle = self._work_handles.popleft()
             if work_handle.callback is not None:
                 work_handle.callback()
+
+    def _attach_failsafe_hook_to_inputs(self, *inputs: Any) -> Any:
```
blefaudeux, line 602: this is a failsafe solution for when people are using reduce buffers, but not all parameters are being used. It should typically work in vision (inputs are floats) but not in NLP (inputs are tokens), in that case if parameters require grads but are not used ShardedDDP will assert as before

blefaudeux: removed, it turns out that one cannot assume that this will be called last

blefaudeux:
fixing the multiple_groups unit test and should be good to go, hopefully a good safety net moving forward

tests/nn/data_parallel/test_sharded_ddp.py
```diff
@@ -109,10 +111,15 @@ def closure():
             same_params = device == torch.device("cpu") or grad_accumulation
             check_same_model_params(same_params=same_params)
 
-    check(broadcast_buffers=False)
-    check(broadcast_buffers=True)
-    check(broadcast_buffers=False, grad_accumulation=True)
-    check(broadcast_buffers=True, grad_accumulation=True)
+    for broadcast_buffers in [False, True]:
```
min-xu-ai, line 114: I see several places here that you loop over what appears to be different test cases. Can they use "parameterized" annotator to be different tests? loop over the cases have several downsides:

- test can timeout
- test failure not clear which case failed
- not able to run tests in parallel

Thanks for adding this extensive tests though! The test is getting a bit too big. Is there a way to perhaps split the test files into multiple files?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -470,33 +488,11 @@ def _setup_backward_hooks(self) -> None:
             p_tmp = param.expand_as(param)
             assert p_tmp.grad_fn is not None
             grad_acc = p_tmp.grad_fn.next_functions[0][0]
-            dst_rank = OSS.get_global_rank(self.process_group, self._trainable_param_to_rank[param])
+            dst_rank = self._trainable_param_to_rank[param]
```
anj-s, line 491: I thought this was a recent change to support multiple groups?

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -394,10 +407,12 @@ def cleanup() -> None:
                             param.grad = None
 
                     # Async reduce for this buffer, log the future
+                    dst_global_rank = OSS.get_global_rank(self.process_group, dst_rank)
+
                     self._work_handles.append(
                         Workhandle(
                             handle=dist.reduce(
-                                tensor=param.grad.data, dst=dst_rank, group=self.process_group, async_op=True
+                                tensor=param.grad.data, dst=dst_global_rank, group=self.process_group, async_op=True
```
blefaudeux, line 415: this was broken when using multiple ranks

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: yes, it's kept, but the buckets were not supported (this was also used to index the buckets)

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -394,10 +407,12 @@ def cleanup() -> None:
                             param.grad = None
 
                     # Async reduce for this buffer, log the future
+                    dst_global_rank = OSS.get_global_rank(self.process_group, dst_rank)
```
blefaudeux, line 410: @anj-s it's now there, so that the local rank can be used to index the buckets, but the reduce goes to the right place. My previous PR was buggy for the buckets, but they were not tested.. this fixes both, test in place + fix

tests/nn/data_parallel/test_sharded_ddp.py
blefaudeux: >> The test is getting a bit too big. Is there a way to perhaps split the test files into multiple files?
Well, it's always possible but then one would need to check two files instead of one ? I tried to refactor a little to save some lines, possible to do more, but all these tests are for the same class so it kind of make sense that they're in the same file, to me at least ? We've a ton of boilerplate code with pytest / multiprocess, it's clearly not really meant for that, but since pytest it is there's not that much I can do on that front

blefaudeux:
I'm refactoring the unit tests a little right now to save some lines, update incoming but it will not change the logic

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -230,21 +227,30 @@ def to(  # type: ignore
         .. note::
             This method modifies the module in-place.
 
+        .. warning:
+            Device changes are not supported, and this will raise an exception. The issue in that case is not
```
min-xu-ai, line 231: nice warning. if possible, perhaps an assert to catch that?

min-xu-ai (approved):
really nice! Hopefully you looped many times of the test locally on a machine with 8 GPUs so that intermittent failures are caught. It would be nice to know them earlier rather than later.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: there is :) check the .to() method

blefaudeux:
> really nice! Hopefully you looped many times of the test locally on a machine with 8 GPUs so that intermittent failures are caught. It would be nice to know them earlier rather than later.

good point, I'll test right now, pytest actually has a nice interface for that

blefaudeux:
> > really nice! Hopefully you looped many times of the test locally on a machine with 8 GPUs so that intermittent failures are caught. It would be nice to know them earlier rather than later.
> 
> good point, I'll test right now, pytest actually has a nice interface for that.
> edit: addendum / I'm not sure why this would be more flaky than other tests though

run 5 times on CI, except for typos when refactoring no flakiness https://app.circleci.com/pipelines/github/facebookresearch/fairscale?branch=sdp_bucket_unit_testing

PR #406, opened 2021-02-19
[fix][OSS] adding an assert for empty shards + corresponding unit test

### LINKED ISSUE #405: [fix] Assert with an understandable error message if there are not enough parameters in the model given the world size
## 🚀 Feature
Assert if shards are empty, explain what the issue is

## Motivation
Can happen in dummy tests and be confusing  **https://github.com/facebookresearch/fairscale/issues/397**

## Pitch
Trivial catch, helps people

## Alternatives
Bump into the same issue later

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #405 .

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -141,8 +141,10 @@ class OSS(Optimizer):
                     self._partition_parameters[rank].append(param_group_rank)
 
             assert min(sum(len(pg["params"]) for pg in partition) for partition in self._partition_parameters) > 0, (
-                "One or more empty shards detected, the world size is too big or the model too small."
-                + "Please reduce your world size if this is the model you would like to train"
+                "One or more empty shards detected, the world size is too big or the model too small.\n"
+                + "Please reduce your world size if this is the model you would like to train\n"
+                + f"Current world size: {self.world_size}\n"
+                + "Current number of parameters: {}".format(sum(len(pg["params"]) for pg in self.param_groups))
             )
 
         return self._partition_parameters
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -140,6 +140,11 @@ class OSS(Optimizer):
                     param_group_rank["params"] = params
                     self._partition_parameters[rank].append(param_group_rank)
 
+            assert min(sum(len(pg["params"]) for pg in partition) for partition in self._partition_parameters) > 0, (
+                "One or more empty shards detected, the world size is too big or the model too small."
+                + "Please reduce your world size if this is the model you would like to train"
+            )
+
         return self._partition_parameters
 
     @property
--- a/fairscale/utils/golden_testing_data.py
+++ b/fairscale/utils/golden_testing_data.py
@@ -8,12 +8,12 @@
 
 adascale_test_data = [
     # "input" value is a list of input tensors for micro-batch/rank 0 and micro-batch/rank 1.
-    {"input": [[1.0, 0], [0, 1.0]], "expected_gain": 2.0},
+    {"input": [[1.0, 0], [0, 1.0]], "expected_gain": 4.0 / 3},
     {"input": [[1.0, 1.0], [1.0, 1.0]], "expected_gain": 1.0000001249999846},
     {"input": [[-1.0, 1.0], [1.0, -1.0]], "expected_gain": 2.0},
-    {"input": [[1.0, 4.0], [5.0, 0.5]], "expected_gain": 1.5022222222222221},
-    {"input": [[-0.2, 3.0], [5.0, 0.5]], "expected_gain": 1.9433267229211089},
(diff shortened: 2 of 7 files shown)


fairscale/optim/oss.py
```diff
@@ -140,6 +140,11 @@ def partition_parameters(self) -> List[List[dict]]:
                     param_group_rank["params"] = params
                     self._partition_parameters[rank].append(param_group_rank)
 
+            assert min(sum(len(pg["params"]) for pg in partition) for partition in self._partition_parameters) > 0, (
+                "One or more empty shards detected, the world size is too big or the model too small."
```
anj-s, line 144: How about printing out the size of the model and world size for easy debugging?

tests/optim/test_oss.py
```diff
@@ -263,6 +262,25 @@ def test_zero_grad():
     mp.spawn(run_test_zero_grad, args=(world_size, temp_file_name), nprocs=world_size, join=True)
 
 
+def run_test_catch_empty_shardd(rank, world_size, tempfile_name):
+    dist_init(rank, world_size, tempfile_name, backend="gloo")
+    x = torch.rand(1)
+    m = torch.nn.Linear(1, 1)
+    try:
```
anj-s, line 269: Can we use assertRaises if you are trying to confirm that an error is raised?

fairscale/optim/oss.py
blefaudeux: sure, I'm guessing that typically people know the world size though since they're the ones who started the job, but I can do that.I'm also guessing that people would not change models just to use OSS, but I can update the message with both informations still

tests/optim/test_oss.py
blefaudeux: sure, all test frameworks have their own checks but can do. I think that one promise of pytest is to limit these actually

fairscale/utils/golden_testing_data.py
```diff
@@ -8,12 +8,12 @@
 
 adascale_test_data = [
     # "input" value is a list of input tensors for micro-batch/rank 0 and micro-batch/rank 1.
-    {"input": [[1.0, 0], [0, 1.0]], "expected_gain": 2.0},
+    {"input": [[1.0, 0], [0, 1.0]], "expected_gain": 4.0 / 3},
```
blefaudeux, line 11: @min-xu-ai @mikerabbat checking with you that this is ok. Since adascale is not changed by this PR, I assumed that the current state was correct

tests/optim/test_oss_adascale.py
```diff
@@ -37,7 +37,7 @@ def _test_basic_func(rank, world_size, tempfile_name, test_case, oss, model=None
     _dist_init(rank, world_size, tempfile_name, backend="nccl")
 
     if model is None:
-        model = Linear(2, 2, bias=False)
+        model = Linear(2, 2)
```
min-xu-ai, line 40: you need the bias or otherwise there isn't enough params?

tests/optim/test_oss.py
```diff
@@ -263,6 +262,21 @@ def test_zero_grad():
     mp.spawn(run_test_zero_grad, args=(world_size, temp_file_name), nprocs=world_size, join=True)
 
 
+def run_test_catch_empty_shardd(rank, world_size, tempfile_name):
+    dist_init(rank, world_size, tempfile_name, backend="gloo")
+    m = torch.nn.Linear(1, 1)
+    with pytest.raises(AssertionError):
```
min-xu-ai, line 268: nice test

fairscale/utils/golden_testing_data.py
min-xu-ai: why is 4.0/3 is the new value? maybe if you init the bias to 0 then the value here won't change? the original value of 2 is because we have two grads from two ranks that are completely independent. it must be the grad from the bias are point to the same direction now, hence 4 grads but 3 directions. So it is fine.

min-xu-ai (approved):
interesting change! I think changing the expected gains are fine. Perhaps a few comments would be fine.

min-xu-ai:
unfortunately the unit tests are still failing. Perhaps you need to update a few more places since the golden values are used in several places.

tests/optim/test_oss_adascale.py
blefaudeux: yes, exactly

fairscale/utils/golden_testing_data.py
blefaudeux: I tried setting the bias to zero, but the returned expected gain is still 1.3333

min-xu-ai: That's fine. I think the new value make sense. Keep it 4.0/3 would be good.

PR #411, opened 2021-02-22
[perf][ShardedDDP] fp16 gradient reduce

### LINKED ISSUE #402: [perf] PyTorch AMP reverts the grads to fp32, meaning that reduce calls are overweight
## 🚀 Feature
Either through torch amp or via a postfix .half() cast (not too nice), make sure that we don't reduce fp32 grads when they were actually computed on fp16..

## Motivation
The gradient size are double what they should be if torch AMP is being used, which can be a bottleneck depending on internode coms

## Pitch
free speed

## Alternatives
Not doing anything, this does not affect correctness

blefaudeux:
# Before submitting

## What does this PR do?
Fixes #402, make it possible to optionally enforce the gradients to be reduced as fp16
TODO: proper unit testing, benchmark perf on slow cluster

## PR review

## Did you have fun?


blefaudeux:
Splitting the ShardedDDP test files in two, as discussed with Min

tests/nn/data_parallel/test_features_sharded_ddp.py
```diff
@@ -0,0 +1,458 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Testing ShardedDDP
```
blefaudeux, line 7: no change when compared to before, just all the tests without DDP parity

tests/nn/data_parallel/test_pytorch_parity_sharded_ddp.py
```diff
@@ -30,10 +30,18 @@
     check_same_models_across_ranks,
     skip_if_less_than_four_gpu,
     skip_if_no_cuda,
-    skip_if_py38,
     skip_if_single_gpu,
 )
 
+"""
+Check that ShardedDDP gets the same results as DDP in a variety of scenarii
```
blefaudeux, line 37: only the DDP parity tests

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -58,7 +58,9 @@ class ShardedDataParallel(nn.Module):
             (default: True) Check whether the parameters trainability (`requires_grad`) has changed and update both ShardedDDP
             and OSS automatically if this is the case. If set to False, `refresh_trainable()` needs to be called anytime
             a parameter is frozen or unfrozen.
-
+        reduce_fp16 (bool):
+            cast the grads to fp16 before reducing. Not needed if the model is already fp16, but will probably improve performance
```
anj-s, line 62: Why does PyTorch convert the grads to FP32?

blefaudeux: It follows the model type, I discovered that recently but basically autocast walks back the graph and follows the types. It makes sense on a single gpu since the next step is the optimizer, but it's a bit unfortunate when distributed since the next step is reduction/all reduce. With ddp you can use a "fp16 compression hook" for similar reasons

min-xu-ai (approved):
nice!

fairscale/nn/data_parallel/sharded_ddp.py
anj-s: Thank you for the context! Can we add this to the PR notes or perhaps a comment here to say that this is equivalent to the fp16 compression hook? (maybe you did but I missed it?)

anj-s (approved):

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: done !

PR #412, opened 2021-02-22
[bug]: not all CUDA memory is freed when model is deleted

min-xu-ai:
# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [ ] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

## What does this PR do?
Fixes # (issue).

## PR review
Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #415, opened 2021-02-23
[test]: add peak mem in checkpoint test

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/misc/checkpoint_activations.py
+++ b/fairscale/nn/misc/checkpoint_activations.py
@@ -30,27 +30,30 @@ def checkpoint_wrapper(module: nn.Module, offload_to_cpu: bool = False) -> nn.Mo
         checkpointed_module = checkpoint_wrapper(my_module, offload_to_cpu=True)
         a, b = checkpointed_module(x, y=3, z=torch.Tensor([1]))
 
+    To understand the benefits of checkpointing and the `offload_to_cpu` flag,
+    let's divide activations into 2 types: inner activations and outer
+    activations w.r.t. the checkpointed modules. The inner ones are saved
+    by activation checkpointing, the outer ones are saved by offload_to_cpu.
+
+    In terms of GPU memory savings:
+
+        - When inner ones are large in size and outer ones are small,
+          checkpointing helps a lot, offload_to_cpu may help a little.
+        - When inner ones are small and outer ones are large,
+          checkpointing helps little, offload_to_cpu helps a lot.
+        - When both inner and outer are large, both help and the
+          benefit is additive.
+
+    The first and last most likely do not benefit from the `offload_to_cpu` flag.
+    That's because input to first layer is the input to the model and won't
+    be moved to CPU by this module. The input the last layer is immediately
+    used by the backward pass and own't resulting in memory saving.
+
     Args:
         module (nn.Module):
             The module to be wrapped
         offload_to_cpu (Optional, bool):
-            Whether to offload activations to CPU. This is non-trivial
-            to use. Enable it doesn't always cause lower peak GPU memory.
-            It can cause more GPU memory being used if used incorrectly.
-
-                - layer with the model input doesn't benefit from offloading
-                  since the input tensor is hold in GPU anyway, offloading
-                  can cause more memory being used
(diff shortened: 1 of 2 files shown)


fairscale/nn/misc/checkpoint_activations.py
```diff
@@ -32,9 +32,25 @@ def checkpoint_wrapper(module: nn.Module, offload_to_cpu: bool = False) -> nn.Mo
...
+
+                - layer with the model input doesn't benefit from offloading
+                  since the input tensor is hold in GPU anyway, offloading
+                  can cause more memory being used
+                - last layer of the model doesn't benefit since immediately
+                  we load the tensor back from CPU in the backward
+                - long model with a lot of checkpoint blocks and small
+                  inner (recomputed) activations is more likely to benefit
+                  from this. But bigger checkpoint blocks might be better
+                  in that case.
```
myleott, line 49: Let's chat more about this. Here's a good repro I shared with Sam that illustrates when/where offload_to_cpu is useful: https://gist.github.com/myleott/7d2ea5a8d27cb6443948e9fa173faf8f#gistcomment-3642158

myleott (changes requested):
Let's also follow up on discussion in #412, I think the current memory tests are not correct, leading to wrong conclusions.

min-xu-ai:
> Let's also follow up on discussion in #412, I think the current memory tests are not correct, leading to wrong conclusions.

We discussed. The test is correct, but for #412 we will follow up with a change for better wrapping technique.

fairscale/nn/misc/checkpoint_activations.py
```diff
@@ -30,27 +30,30 @@ def checkpoint_wrapper(module: nn.Module, offload_to_cpu: bool = False) -> nn.Mo
...
+    In terms of GPU memory savings:
+
+        - When inner ones are large in size and outer ones are small,
+          checkpointing helps a lot, offload_to_cpu may help a little.
+        - When inner ones are small and outer ones are large,
+          checkpointing helps little, offload_to_cpu helps a lot.
+        - When both inner and outer are large, both help and the
+          benefit is additive.
+
+    The first and last most likely do not benefit from the `offload_to_cpu` flag.
```
min-xu-ai, line 47: first and last layer, missed the word layer.

fairscale/nn/misc/checkpoint_activations.py
```diff
@@ -30,27 +30,30 @@ def checkpoint_wrapper(module: nn.Module, offload_to_cpu: bool = False) -> nn.Mo
...
+    That's because input to first layer is the input to the model and won't
+    be moved to CPU by this module. The input the last layer is immediately
+    used by the backward pass and own't resulting in memory saving.
```
min-xu-ai: typo. fixed in the next commit.

fairscale/nn/misc/checkpoint_activations.py
myleott: Looks great! A couple comments:
- at first I thought the "first and last" here referred to the bullets above, but I think you mean layers?
- reflecting on it, I think the first layer may be useful to checkpoint in some cases, shall we just say the last layer? 

Something like:
```
Note: The last layer is not likely to benefit (...)
```

myleott (approved):

fairscale/nn/misc/checkpoint_activations.py
min-xu-ai: Why is first layer benefiting from cpu_offloading?

Yeah, I fixed the typo. I meant first and last layers.

myleott: If the input is very large (e.g., images? video?), you might still want to move it to CPU since you'll need it to recompute the activations for the first layer in the backward pass

min-xu-ai: But that input is often referred outside by the training loop, so this module can't really help since the outside reference will keep that input in GPU memory, right?

min-xu-ai: I clarified it in my last commit.

PR #422, opened 2021-02-23
[perf][OSS] Overlap broadcasts and optim step

blefaudeux:
# Before submitting

## What does this PR do?
Slightly more complicated to grasp but easy speed bump: overlap the optimizer step and the shard broadcast. If the interconnect is slow and the optimizer step is sizeable, then speed benefit, else little to no change, but never negative.

Note that this only works if the optimizer does not own the closure, but it's also the case for the pytorch grad scaler that everyone uses so it's not a big limitation in practice

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -235,14 +235,26 @@ class OSS(Optimizer):
         # Sync oss param_groups attributes in case they've been updated by a scheduler.
         OSS._sync_param_groups(self.param_groups, self.optim.param_groups)
 
-        # Run the optimizer step on this shard only:
-        if closure is not None:
-            loss = self.optim.step(closure=closure, **kwargs)  # type: ignore
-        else:
-            loss = self.optim.step(**kwargs)
+        rank_updated = False
+        last_work_handle = None
 
-        # Sync all the updated shards in between the ranks
-        self._broadcast_params()
+        if closure is not None:
+            # Run the optimizer step on this shard only, cannot be interleaved with comms
+            loss = self.optim.step(closure=closure, **kwargs)  # type: ignore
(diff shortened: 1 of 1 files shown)


blefaudeux:
tested, 20% throughput improvement on a multi node job (arguably with a not too good interconnect)

fairscale/optim/oss.py
```diff
@@ -235,14 +235,26 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
...
             loss = self.optim.step(closure=closure, **kwargs)  # type: ignore
-        else:
-            loss = self.optim.step(**kwargs)
+            rank_updated = True
+
+        # Interleave optimizer steps and broadcasts, overlap comms and compute
+        for device in self.buckets.keys():
+            for src_rank, bucket in enumerate(self.buckets[device]):
+                if src_rank == self.rank and not rank_updated:
+                    loss = self.optim.step(**kwargs)
```
min-xu-ai, line 250: I am confused. This only needs to be called once to step all params, right? why is it in this loop?

blefaudeux: it will only be called once (hence the `rank_updated` variable), the idea is to fire coms and compute calls in parallel instead of in a sequence

blefaudeux: ah wait, I'm mistaken indeed.. let me put it back to draft

blefaudeux:
> tested, 20% throughput improvement on a multi node job (arguably with a not too good interconnect)


hmm so I guess that this is mostly run to run noise, sorry about that.. what I wrote initially does seem wrong

PR #429, opened 2021-02-24
[fix] Fix loading of nested FSDP wrapper with flatten_params=True

myleott:


fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -142,11 +142,48 @@ def flat_state_dict(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
     def load_state_dict(
         self, state_dict: Union[Dict[str, Tensor], "OrderedDict[str, Tensor]"], strict: bool = True
     ) -> NamedTuple:
-        if "flat_param" in state_dict:
-            return super().load_state_dict(state_dict, strict=strict)
-        else:
-            with self.unflatten_params():
-                return self.module.load_state_dict(state_dict, strict)
+        # Note: nested FlattenParamsWrapper instances are problematic because
```
min-xu-ai, line 145: wow, this seems crazy complicated.

not for this change for sure, but I was thinking that perhaps we can make a generic wrapper for nn.Module in the future and use it as a mixin for other wrappers. Like:

class WrapperMixin(nn.Module):
    code that deal with generic delegating, set/get attr, state dict, Len(), index, key, value, etc.

class FlattenParamWrapper(nn.Module, WrapperMixin):
    code that specific for this wrapper

We need to make sure python MRO (method resolution order) is correct in this case.

I don't know if you think this is worth it?

myleott: Yes, that's a great idea and I do think it's worth it. I really like the wrapper abstraction, it seems to work well for a bunch of distributed techniques.

Somewhat of an aside, but in fairseq we have a few other wrappers that are pretty nice and should probably move to fairscale too, e.g., the [`DistributedTimeoutWrapper`](https://github.com/pytorch/fairseq/blob/master/fairseq/distributed/distributed_timeout_wrapper.py) to kill hung jobs.

myleott: re: why this is so complicated, it's partly because PyTorch calls every child module's `state_dict` method, but doesn't do the same for `load_state_dict`. Thus the outer-most call is responsible for doing any necessary manipulation for all the children instances :/

min-xu-ai: How about you move the list of things that can be moved over in an issue so that we don't forget to move them. Should be really nice to move them and then use a mixin.

For the `load_state_dict`, worth to open an issue with pytorch?

min-xu-ai:
I think you need to run fs_test to format the files.

myleott:
will do this in https://github.com/facebookresearch/fairscale/pull/434

PR #432, opened 2021-02-24
[feature] Add support for OffloadModel to enable training large models on 1 GPU.

anj-s:
# Before submitting

- [ X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?

## What does this PR do?
Add experimental support for using the OffloadModel API which enables training large models on a single GPU. OffloadModel chunks the given model into a list of modules and copies a given chunk from CPU->GPU during the FW pass. After FW computation the chunk is copied back to the CPU. The process is repeated for the BW pass. The current implementations supports:
* Specifying number of slices that you want to chunk your model into.
* Support for activation checkpointing.
* Support for running multiple microbatches at one time to offset latency due to multiple param copies fro CPU<->GPU.

Caveats: 
* This initial implementation only supports nn.Sequential models.
* The throughput of the model is smaller than when running without Offload. We will be continue to work on improving performance and suggest configurations that will enable the highest throughput.

References: 
* This work is inspired by [ZeRO Offload](https://arxiv.org/abs/2101.06840) and [L2L](https://arxiv.org/abs/2002.05645). 
* The starting point of this PR heavily borrows from [PR](https://github.com/facebookresearch/fairscale/pull/279) by @blefaudeux .  

## PR review

## Did you have fun?


min-xu-ai:
quick question on the test file location. Should it be

 tests/nn/experimental/test_offload.py 
 
or

 tests/experimental/nn/test_offload.py?

I think we mirror the dirs. File names can be shorten, like we have test_fsdp*.py but all in the same mirrored dir. That seems like a good convention?

min-xu-ai:
also, see this comment: https://github.com/PyTorchLightning/pytorch-lightning/pull/6152#issuecomment-785473993

anj-s:
> quick question on the test file location. Should it be
> 
> tests/nn/experimental/test_offload.py
> 
> or
> 
> tests/experimental/nn/test_offload.py?
> 
> I think we mirror the dirs. File names can be shorten, like we have test_fsdp*.py but all in the same mirrored dir. That seems like a good convention?

I agree. I want it to be in experimental/ just like I moved tests for ampnet.

fairscale/experimental/nn/offload.py
```diff
@@ -0,0 +1,447 @@
...
+        offload_device (torch.device):
+            Device where the inactive model should reside.
+
+        num_slices (int):
+            Number of slices into which the model should be chunked.
+    """
+
+    def __init__(
+        self,
+        model_cpu: nn.Sequential,  # hard pre-requisite for now, easier model slicing
```
blefaudeux, line 375: discussing elsewhere, but I think that the FSDP way (wrap submodules) could apply here, and why not keeping the two options (either one monolithic nn.Sequential call, or a per-module wrap) open ? I think that it adds a lot of flexibility and could be good enough in practice

blefaudeux: practically speaking this means that https://pytorch.org/docs/stable/generated/torch.nn.Module.html?highlight=forward%20hook#torch.nn.Module.register_forward_pre_hook can be used, but the latency will be pretty terrible if used "naively" (wait for the FW wavefront to touch, pull in the module), so it's not really a silver bullet

fairscale/experimental/nn/offload.py
```diff
@@ -0,0 +1,447 @@
...
+        # Bool indicating if we want to checkpoint activation on the host.
+        self._checkpoint_activation = checkpoint_activation
+
+        # Number of microbatches to run per batch on the device
+        self._num_microbatches = num_microbatches
+
+    def forward(self, *inputs: Any, **_: Any) -> Any:
+        # At least one of the inputs needs to have `requires_grad` set.
+        # TODO(anj-s): Should we require users to set this or should we set it here?
+        for inp in inputs:
```
blefaudeux, line 421: I would add an exception here if none of the inputs could be tagged, so that it would explain a bit more what the issue is ?

blefaudeux:
Thanks for the great PR @anj-s, it's super comprehensive ! I think that we can try to make it more generic over time, it does not have to be perfect right now and it's a very solid basis I believe. Minor nits if you don't mind and curious to have @min-xu-ai eyes on that

benchmarks/experimental/offload.py
```diff
@@ -0,0 +1,412 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
```
min-xu-ai, line 1: add license lines?

fairscale/experimental/nn/offload.py
```diff
@@ -0,0 +1,447 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+
```
min-xu-ai, line 6: here you can add a top level docstring for the file if you like.

fairscale/experimental/nn/offload.py
```diff
@@ -0,0 +1,447 @@
...
+            self.model_shard.to(self.offload_device, non_blocking=non_blocking)
+
+    def backward_drop(self, non_blocking: bool = True) -> None:
+        with torch.cuda.stream(self.cuda_stream):
+            self.model_shard.to(self.offload_device, non_blocking=non_blocking)
+
+
+class ActivationCheckpointing(torch.autograd.Function):
+    """
+     This Function enables us to override the forward and backward pass of the nn.Module.
```
min-xu-ai, line 134: the comment here doesn't seem to describe the class? In particular, it doesn't mention activation at all but the class name is activation.

fairscale/experimental/nn/offload.py
```diff
@@ -0,0 +1,447 @@
...
+
+class OffloadModel(nn.Module):
+    """Wrapper used offload parts of a model to the CPU.
+
+    The model is sharded such that parts of the model are copied
+    from CPU->GPU for the FW pass and back right after. The reverse
+    copy is carried out for the backward pass.
+
+    Args:
+        module (~torch.nn.Sequential): Module to be offloaded.
```
min-xu-ai, line 361: the doc seems to be out of sync with the code.

min-xu-ai (approved):
Sorry for late to the party. I agree with Ben that this gives us a good start. Lots of interesting things we can do potentially with this.

benchmarks/experimental/offload.py
anj-s: Done.

fairscale/experimental/nn/offload.py
anj-s: I am not sure about that yet. Let me punt on that for now.

PR #433, opened 2021-02-25
[feat]: add summon_full_params context mgr

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -384,17 +384,18 @@ class FullyShardedDataParallel(nn.Module):
         .. warning:: This needs to be called on all ranks, since synchronization
             primitives will be used.
         """
-        torch.cuda.synchronize()
-        self._lazy_init()
-        self._rebuild_full_params()
-        self._all_buffers_to(dtype=torch.float32)  # Buffers dtype stays consistent with parameters.
-        state_dict = self.module.state_dict(*args, **kwargs)
-        # We don't free the params after generating the state dict, since
-        # freeing is done in-place (via the Storage) and would corrupt the
-        # returned state dict. However, we need to maintain the invariant that
-        # p.data corresponds to the FP32 param shard, so we do that here.
-        self._use_fp32_param_shard()
-        self._all_buffers_to(dtype=self.compute_dtype)
+        with self.summon_full_params():
+            # Buffers dtype stays consistent with parameters.
+            self._all_buffers_to(dtype=torch.float32)
+
+            state_dict = self.module.state_dict(*args, **kwargs)
+            # We copy the state_dict since full param will be free after
+            # we exit the summon_full_params() context.
+            # TODO (Min): if the state is huge, the copy can OOM.
+            state_dict = copy.deepcopy(state_dict)
+
+            # In case we are in mixed precision, restore buffers back to fp16.
+            self._all_buffers_to(dtype=self.compute_dtype)
         return state_dict
 
     # TODO (Min): figuring out how to do typing for this overloaded function.
@@ -420,11 +421,8 @@ class FullyShardedDataParallel(nn.Module):
         .. warning:: This needs to be called on all ranks, since synchronization
             primitives will be used.
         """
-        torch.cuda.synchronize()
-        self._lazy_init()
-        self._rebuild_full_params()
-        output = self.module.load_state_dict(state_dict, strict)
-        self._free_full_params()
+        with self.summon_full_params():
+            output = self.module.load_state_dict(state_dict, strict)
         return output
 
     def load_local_state_dict(
@@ -462,11 +460,15 @@ class FullyShardedDataParallel(nn.Module):
     def summon_full_params(self) -> Generator:
         """
         A context manager to expose full params for the underlying model.
-        Can be useful after forward/backward for a model to get the params
+        Can be useful *after* forward/backward for a model to get the params
         for additional processing or checking.
 
         This can be used on inner FSDPs.
+
+        This can *not* be used within a forward or backward pass. Nor can forward
+        and backward be started from within this context.
         """
+        torch.cuda.synchronize()
         self._lazy_init()
         self.assert_state(TrainingState.IDLE)
         # Set the state so that we assert when trying to go into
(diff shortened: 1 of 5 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -438,6 +439,28 @@ def no_sync(self) -> Generator:
...
 
+    @contextlib.contextmanager
+    def summon_full_params(self) -> Generator:
+        """
+        A context manager to expose full params for the underlying model.
+        Can be useful after forward/backward for a model to get the params
+        for additional processing or checking.
+
+        This can be used on inner FSDPs.
+        """
```
myleott, line 450: perhaps add a note that this can't be used in the middle of a forward/backward pass, nor can a forward/backward pass be started from within the context manager

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -438,6 +439,28 @@ def no_sync(self) -> Generator:
             for m, old_flag in old_flags:
                 m.require_backward_grad_sync = old_flag
 
+    @contextlib.contextmanager
+    def summon_full_params(self) -> Generator:
```
myleott, line 443: can we use this to replace similar logic in state_dict and load_state_dict too?

myleott (approved):

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: good idea. will do.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: let me try it. If works, will just make the change and merge it in.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -383,17 +384,18 @@ def state_dict(self, *args, **kwargs):  # type: ignore
...
-        self._all_buffers_to(dtype=self.compute_dtype)
+        with self.summon_full_params():
+            # Buffers dtype stays consistent with parameters.
+            self._all_buffers_to(dtype=torch.float32)
+
+            state_dict = self.module.state_dict(*args, **kwargs)
+            # We copy the state_dict since full param will be free after
+            # we exit the summon_full_params() context.
+            # TODO (Min): if the state is huge, the copy can OOM.
+            state_dict = copy.deepcopy(state_dict)
```
min-xu-ai, line 395: @myleott @sshleifer let me know if this deepcopy is OK. It seems that for very huge model this would be bad, but it restores invariant of freeing full params after use. For huge models, state checkpointing probably needs more fundamental work anyway.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -815,8 +840,11 @@ def _wait_for_post_backward(self) -> None:
         if self.move_grads_to_cpu:
             # Wait for the non-blocking GPU -> CPU grad transfers to finish.
             torch.cuda.current_stream().synchronize()
-        # A backward pass is done.
-        self.training_state = TrainingState.IDLE
+        # A backward pass is done, update root and nested FSDP's flags.
+        for m in self.modules():  # includes self
```
min-xu-ai, line 844: Thanks for the suggestion. Found and fixed a bug on the TrainingState.

myleott: Nice catch!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
myleott: Ah, for very large models we will use `local_state_dict` which is not affected by this, so all good. There is some inefficiency here, but we can revisit if it becomes a problem.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: right now, the delayed reduce test actually catches this. I think there are about 4 tests affected by this. That's how I found it.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -419,11 +421,8 @@ def load_state_dict(
         .. warning:: This needs to be called on all ranks, since synchronization
             primitives will be used.
         """
-        torch.cuda.synchronize()
```
myleott, line 422: This needs to be added back, probably causing the test failures

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -383,17 +384,18 @@ def state_dict(self, *args, **kwargs):  # type: ignore
...
```
myleott, line 386: Same, add this back please

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -383,17 +384,18 @@ def state_dict(self, *args, **kwargs):  # type: ignore
...
-        self._lazy_init()
```
myleott, line 387: We still need these two lines to properly wait for the previous optimizer step to finish

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -419,11 +421,8 @@ def load_state_dict(
...
```
myleott, line 423: same, we need these two lines

min-xu-ai:
@myleott I see some comments about adding back cuda.synchronize/lazy_init calls in my email. But I couldn't find the comments on github UI for some reason. Let me know if I need to address more issues. I think those calls are not missed and I have found all the tests passing now.

I will merge it in now. Happy to address more comments separately for sure.

myleott:
Ah, please disregard the comments (I had deleted them). I thought the test failures were related to something else, but then realized I was mistaken

Issue #430, opened 2021-02-24
[FSDP] Can we access parameter views when using flatten_parameters=True?


SeanNaren:
## ❓ Questions and Help

This should explain the case:

```python
import torch
from fairscale.nn.data_parallel import FullyShardedDataParallel
import os

os.environ['MASTER_ADDR'] = 'localhost'
os.environ['MASTER_PORT'] = '1337'
torch.distributed.init_process_group("gloo", rank=0, world_size=1)


class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(5, 5)


model = FullyShardedDataParallel(Model(), flatten_parameters=False)

# prints parameter list of module
print([p for p in model.module.layer.parameters()])

model = FullyShardedDataParallel(Model(), flatten_parameters=True)

# prints []
print([p for p in model.module.layer.parameters()])

# Throws an error: ValueError: optimizer got an empty parameter list
optimizer = torch.optim.SGD(model.layer.parameters(), lr=1e-3)

```

When `flatten_parameters=True` we remove the parameter as we have migrated it to a contiguous buffer, but this means when we call `.parameters()` on specific modules (in the case we only want to wrap certain parts of the model with optimizers) this can not be done.

Any remedy to this problem? We were experimenting with the possibility of using views to replace this functionality however this doesn't return a parameter I think. Alternatively, we could tell the users if they run into issues like above, to turn off `flatten_parameters`.

myleott:
Hmm, this is a bit tricky. For the first part, this works (the params will get re-flattened after the context manager exits):
```
with model.unflatten_params():
    print([p for p in model.module.layer.parameters()])
```

The optimizer use case is harder though. The optimizer won't be happy with the re-flattening that happens when the context manager exits:
```python
class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(5, 5)

    def forward(self, x):
        return self.layer(x)

(...)

print("param norm (before)", torch.norm(torch.stack([p.norm() for p in model.parameters()])))

with model.unflatten_params():
    optimizer = torch.optim.SGD(model.layer.parameters(), lr=0.1)
optimizer.zero_grad()
loss = model(torch.rand(8, 5)).sum()
loss.backward()
optimizer.step()
print("param norm (after, broken)", torch.norm(torch.stack([p.norm() for p in model.parameters()])))

[...]

PR #434, opened 2021-02-25
[fix] Fix nested FlattenParamsWrapper state_dict/load_state_dict

myleott:


fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -131,21 +162,53 @@ def __getattr__(self, name: str) -> Any:
...
 
     def flat_state_dict(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
         """Return the flattened state_dict."""
-        return super().state_dict(*args, **kwargs)
+        assert self.is_flattened
+        with ExitStack() as stack:
+            # tell any nested FlattenParamsWrapper instances not to auto unflatten
+            for module in self.modules():
+                if isinstance(module, FlattenParamsWrapper):
+                    stack.enter_context(module._no_auto_unflatten_state_dict())
```
sshleifer, line 179: Very nice

min-xu-ai:
Thanking out loud, instead of trying to hack around all the recursive thing in state_dict and load_state_dict, can we can operate on the resulting dict directly?

state_dict (vanilla version, no FSDP) -> some preprocessing code -> state_dict (compatible with FSDP/flatten_param) -> hand it to load_state_dict() of pytorch.

The trick is the "some preprocessing code". I'd image that code operates on pure dict data, perhaps even remove unneeded shards to reduce its size. But would that code be easier/harder to write?

myleott:
Yeah, I think we probably should go down that route. One idea is that instead of storing the wrapped stuff in `self.module`, we could give it a more grep'able name, like `self._fsdp_wrapped_module`, `self._fpw_wrapped_module`, etc.

Alternatively we could try to build an API on top of state_dicts that looks more like this: tinyurl.com/32xm4ccb

sshleifer:
**TLDR:** happy to help, not sure which option is best.

**Thinking out loud:**


0) I think this PR is a good change.

1) I feel 7/10 strongly that any approach to the full problem is significantly better than letting the user figure it out. It's frustrating to print out two state dicts side by side, know that they are basically the same, know that someone else has printed out the exact same thing before and hacked around it and moved on, then hack around it and move on.

2) This feeling, and warm memories of hacking many a state dict for hf/transformers conversions, makes me think that min's proposal above may be the way to go for `FSDP`, especially if we can keep pretending that `FlattenParametersWrapper` doesn't know about `FSDP` but `FSDP` knows about `FPW`.

3)  Maybe a pipe dream, but this all gets a lot simpler if we can remove `flatten_parameters=True/False` optionality in `FSDP`. IIRC, @min-xu-ai has a very good reason to use `flatten_parameters=False`?

min-xu-ai:
> IIRC, @min-xu-ai has a very good reason to use flatten_parameters=False?

Don't know it was a "good reason" but I was forced to use it because BatchNorm wants to be special in mixed precision. It needs to run with fp32 states, so we can't have it flattened with other fp16 params. :-( Don't really know how common this really it in other workloads. Even with vision, they are many papers talking about training without BN. Unfortunately, presently VISSL still need BN as well as SyncBN (for DDP).

myleott:
Okay, lemme take a stab at something clean here.

Re: 3, we're stuck with flatten_params for now, due to huge performance drop without it (see benchmarks here: https://github.com/facebookresearch/fairscale/issues/430#issuecomment-785463057). After bucketing is fully implemented, it should mostly close the gap, but not fully.

Re: BatchNorm, we should come up with a solution here. Perhaps we can keep separate flattened states for some modules, not sure. We should add some tests that illustrate the BN problem though...

min-xu-ai:
>  We should add some tests that illustrate the BN problem though...
Yeah, this is on my todo.

min-xu-ai:
Did you see the state_dict hook?

https://github.com/pytorch/pytorch/blob/569d4fe3f95af913d5cc20ded09c2d000e974cc9/torch/nn/modules/module.py#L1081

Maybe it can be used to generate necessary info during state_dict generation and then used in load_state_dict?

myleott:
Nice find! `_register_load_state_dict_pre_hook` seems perfect. Each nested FSDP/FlattenParamsWrapper can modify the state dict right before loading

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -131,22 +171,61 @@ def __getattr__(self, name: str) -> Any:
...
-        with self.unflatten_params():
-            return self.module.state_dict(*args, **kwargs)
+        """Return the wrapped module's state_dict (unflattened)."""
+        if self.is_flattened and self._auto_unflatten_state_dict:
+            with self.unflatten_params(recurse=False):
+                return super().state_dict(*args, **kwargs)
+        else:
+            return super().state_dict(*args, **kwargs)
+
+    def _post_state_dict_hook(
```
min-xu-ai, line 181: make this a static method too? In fact, why not make both hooks normal methods outside of the class? They don't need "self", right?

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -110,18 +133,35 @@ def _unflatten_params(self) -> None:
...
+        If the current instance is already unflattened, then it will remain
+        unflattened after the context manager exits.
+        """
+        if recurse:
+            with ExitStack() as stack:
+                for module in self.modules():
+                    if isinstance(module, FlattenParamsWrapper):
+                        stack.enter_context(module.unflatten_params(recurse=False))
+                yield
+            return
```
min-xu-ai, line 156: how about change this to a "else:"? early return can cause future common code to be skipped by mistaken.

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -110,18 +133,35 @@ def _unflatten_params(self) -> None:
...
-        self._unflatten_params()
+    def unflatten_params(self, recurse: bool = True) -> Generator:
+        """
+        Unflatten params (optionally recursively on all nested instances).
+        If the current instance is already unflattened, then it will remain
+        unflattened after the context manager exits.
+        """
+        if recurse:
+            with ExitStack() as stack:
+                for module in self.modules():
```
min-xu-ai, line 152: add a bit comment on this loop? It is not common to see loop and enter context in a context mgr. It is unclear what's the semantics to the caller.

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -131,22 +171,61 @@ def __getattr__(self, name: str) -> Any:
             return getattr(self.module, name)  # fallback to wrapped module
 
     def state_dict(self, *args: Any, **kwargs: Any) -> "OrderedDict[str, Tensor]":  # type: ignore
-        """Return an unflattened state_dict."""
-        with self.unflatten_params():
-            return self.module.state_dict(*args, **kwargs)
+        """Return the wrapped module's state_dict (unflattened)."""
+        if self.is_flattened and self._auto_unflatten_state_dict:
+            with self.unflatten_params(recurse=False):
```
min-xu-ai, line 176: nice!

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -131,22 +171,61 @@ def __getattr__(self, name: str) -> Any:
...
+            state_dict = self.state_dict(*args, **kwargs)
+        return state_dict
+
+    @contextmanager
+    def _no_auto_unflatten_state_dict(self) -> Generator:
+        backup = self._auto_unflatten_state_dict
+        self._auto_unflatten_state_dict = False
+        yield
+        self._auto_unflatten_state_dict = backup
+        return
```
min-xu-ai, line 204: extra return?

fairscale/utils/state_dict.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Replace all keys that match a given old_prefix with a new_prefix (in-place).
+
+    Usage::
+
+        state_dict = {"layer.xyz": torch.tensor(1)}
+        replace_by_prefix_(state_dict, "layer.", "module.layer.")
+        assert state_dict == {"module.layer.xyz": torch.tensor(1)}
+    """
+    for key in list(state_dict.keys()):
```
min-xu-ai, line 68: assert old_prefix != new_prefix? If they equal, I think we will loss items.

min-xu-ai (approved):
very cool!

sshleifer (approved):

stubs/torch/__init__.pyi
```diff
@@ -28,6 +28,7 @@ from . import cuda as cuda
 from . import optim as optim
 from . import nn as nn
 from . import testing as testing
+from . import utils as utils
```
min-xu-ai, line 31: I am including these two unrelated changed in a separate PR. If it goes in first, you can remove these. otherwise, merge them in, I will deal with conflicts.

PR #440, opened 2021-02-26
[fix] fix FSDP state_dict/load_state_dict for nested wrapped instances

myleott:
Similar to #434 but for FSDP now

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -173,11 +173,11 @@ class FullyShardedDataParallel(nn.Module):
         params = list(p for p in module.parameters() if not hasattr(p, "_is_sharded"))
 
         if self.flatten_parameters and len(params) > 0:
-            self._fsdp_module: nn.Module = FlattenParamsWrapper(module, param_list=params)
+            self._fsdp_wrapped_module: nn.Module = FlattenParamsWrapper(module, param_list=params)
             del module  # free original module in case it helps garbage collection
-            self.params = [self._fsdp_module.flat_param]
+            self.params = [self._fsdp_wrapped_module.flat_param]
         else:
-            self._fsdp_module = module
+            self._fsdp_wrapped_module = module
             self.params = params
 
         # Shard module parameters in place
@@ -196,8 +196,8 @@ class FullyShardedDataParallel(nn.Module):
         # Enum to indicate if we're in the forward/backward pass, idle, etc.
         self.training_state = TrainingState.IDLE
 
-        # Register hook to be called after state_dict() to remove the
-        # "_fsdp_module." prefix and before load_state_dict() to add it back.
+        # Register hook after state_dict() to remove the "_fsdp_wrapped_module."
+        # prefix and before load_state_dict() to add it back.
         self._register_state_dict_hook(_post_state_dict_hook)
         self._register_load_state_dict_pre_hook(_pre_load_state_dict_hook)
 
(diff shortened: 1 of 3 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -384,18 +401,30 @@ def state_dict(self, *args, **kwargs):  # type: ignore
...
-            self._all_buffers_to(dtype=torch.float32)
-
-            state_dict = self.module.state_dict(*args, **kwargs)
-            # We copy the state_dict since full param will be freed after
-            # we exit the summon_full_params() context.
-            for key in state_dict.keys():
-                state_dict[key] = state_dict[key].clone()
+        # TODO
+        # Buffers dtype stays consistent with parameters.
+        # self._all_buffers_to(dtype=torch.float32)
```
myleott, line 406: Removing these doesn't seem to cause any tests failures. It's not clear to me if this should be removed? Or if we should add more tests?

min-xu-ai: more likely I think we just don't have the test. I will file an issue.

min-xu-ai: or do you wanna address it here? I am fine either way.

myleott: I see, we don't maintain the invariant that buffers are cast back to fp32 at the end of each `forward()`. Actually, I'm realizing we don't maintain fp32 buffers at all, so there could be loss of precision throughout training... we should fix that, perhaps in a separate PR: https://github.com/facebookresearch/fairscale/issues/442

sshleifer: self-assigned #442. I wrote the existing buffer logic with the incorrect assumption that buffers are static.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -172,11 +173,11 @@ def __init__(
         params = list(p for p in module.parameters() if not hasattr(p, "_is_sharded"))
 
         if self.flatten_parameters and len(params) > 0:
-            self.module: nn.Module = FlattenParamsWrapper(module, param_list=params)
+            self._fsdp_module: nn.Module = FlattenParamsWrapper(module, param_list=params)
```
sshleifer, line 176: (nit) should this be called `_child_module/_wrapped_module/`? Does not seem guaranteed be an `FSDP` instance.

sshleifer (approved):
LGTM, nice tests!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
myleott: yep, will rename to `_fsdp_wrapped_module`. Key is that it's easy to grep for in a state_dict, and unlikely to be in the user's module

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -235,6 +251,7 @@ def clip_grad_norm_(
         .. warning:: This needs to be called on all ranks, since synchronization
             primitives will be used.
         """
+        self._lazy_init()
```
min-xu-ai, line 254: 👍

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1011,3 +1072,23 @@ def alloc_storage_(data: torch.Tensor, size: torch.Size) -> None:
         return
     assert data.storage().size() == 0
     data.storage().resize_(size.numel())
+
+
+def _post_state_dict_hook(
+    module: nn.Module, state_dict: "OrderedDict[str, torch.Tensor]", prefix: str, *args: Any
+) -> "OrderedDict[str, torch.Tensor]":
+    if module.training_state == TrainingState.SUMMON_FULL_PARAMS:
+        # We copy the state_dict since full param will be freed after
```
min-xu-ai, line 1081: nice to put it in the hook!

min-xu-ai (approved):
awesome!

PR #446, opened 2021-02-27
[feat] Add context manager to FSDP for easier child module wrapping

SeanNaren:
## What does this PR do?

As discussed in https://github.com/PyTorchLightning/pytorch-lightning/pull/6152#issuecomment-785950642 this adds a context manager that assists in making child modules with similar defaults.

```python
from fairscale.nn.misc import enable_wrap, wrap

with enable_wrap(**handleful_of_important_params):
    layer_1 = wrap(torch.nn.Linear(5, 5))
    layer_2 = wrap(torch.nn.Linear(5, 5), flatten_parameters=True) # Override parameters if you'd like

...

# without the context manager, creates Linear layer
layer_1 = wrap(torch.nn.Linear(5, 5))
```

If not within the FSDP context, this would be a no-op. This makes it easier to annotate layers without having to copy any changes in parameters.

## Before submitting

## PR review

## Did you have fun?


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,53 @@ def assert_state(self, state: TrainingState) -> None:
             self.training_state == state
         ), f"expected to be in state {state} but current state is {self.training_state}"
 
+    @staticmethod
+    @contextlib.contextmanager
+    def enable_auto_wrap(**kwargs) -> Iterator[None]:
```
min-xu-ai, line 1074: The name `auto_wrap` made me think it is related to auto-determine which part to wrap. I suggest we rename this feature. How about calling it `config_context`?

i.e.:

with FSDP.config_context():
   FSDP.wrap_if_in_config_context(...)

min-xu-ai, line 1074: Also, why return Iterator[None] type not Generator type? A comment might be nice.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,53 @@ def assert_state(self, state: TrainingState) -> None:
...
+        """
+        Context manager to wrap modules in FullyShardedDataParallel.
+        Useful for when you'd like to apply the same parameters to all child modules that you wrap.
+
+            ...
+            with FullyShardedDataParallel.enable_auto_wrap(**fully_sharded_params):
```
min-xu-ai, line 1080: I think it is fine to use FSDP here in the example not the full name.

To format the doc, we often use

```
Usage::

    code
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,53 @@ def assert_state(self, state: TrainingState) -> None:
...
+
+    @staticmethod
+    def auto_wrap(module, **kwargs):
+        if FullyShardedAutoWrap.in_autowrap_context:
+            kwargs = {**FullyShardedAutoWrap.kwargs, **kwargs}
+            return FullyShardedDataParallel(module, **kwargs)
+        return module
+
+
+class FullyShardedAutoWrap:
```
min-xu-ai, line 1096: I think some docstring might be needed, which might be the reason CI is failing. I use this alias to do auto-format to make CI happy:

```
alias fs_test='black . && isort . && flake8 && mypy --ignore-missing-imports --scripts-are-modules --pretty . && echo ALL GOOD'
```

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -697,6 +697,38 @@ def _test_no_sync(self, model, batch_dim):
         assert objects_are_equal(ref_grads, accumulated_grads, raise_exception=True)
 
 
+class TestAutoWrap(unittest.TestCase):
```
min-xu-ai, line 700: if this test is not dependent on the rest of the file, move it to test_fsdp_config_context.py? This file is getting too big.

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -697,6 +697,38 @@ def _test_no_sync(self, model, batch_dim):
...
+        with FullyShardedDataParallel.enable_auto_wrap(flatten_parameters=False, process_group=self.process_group):
+            layer = FullyShardedDataParallel.auto_wrap(torch.nn.Linear(5, 5))
+        assert isinstance(layer, FullyShardedDataParallel)
+        assert layer.flatten_parameters is False
+
+    def test_auto_wrap_disabled_outside_context(self):
+        layer = FullyShardedDataParallel.auto_wrap(torch.nn.Linear(5, 5))
+        assert isinstance(layer, torch.nn.Linear)
+
+    def test_auto_wrap_override_defaults(self):
```
min-xu-ai, line 725: also, is it worth testing nested wrapping like:

with FSDP.config_context()
    FSDP.wrap_if_in...
        with FSDP.config_context()
               FSDP.wrap_if_in

min-xu-ai (commented):
Thank you for your contribution! I have some comments below.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: wrap_if_in_config_context might be too long, we can use wrap_if_in_CC.

myleott:
Thanks for doing this!

First, I just want to clarify some important details about this API and the use case. This API will be a core API for end users (e.g., Lightning users, Fairseq users) to support initializing really large models, where it's critical to shard layers **as they are initialized** to avoid system OOM. Thus, even after we solve "auto-determining which parts to wrap", we will *still* need this API to avoid system OOMs.

Re: naming, since this will be the main user-facing interface for most framework users, we should make the name really short and easy to type. At most one underscore 😄 

Reflecting on it, I agree that `auto_wrap` gives the wrong impression. Some other options:
1) FSDP.wrap: even shorter, but a bit vague
2) FSDP.maybe_wrap: a bit longer, but pretty accurate description of what's happening
3) FSDP.shard: this best reflects the use case (sharding layers as they are initialized to avoid system OOM)
4) FSDP.maybe_shard: probably the most accurate description, but at 11 characters, probably too long

Thoughts?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
myleott: See main comments re: naming. It's really important that whatever name we choose is short and easy to type

tests/nn/data_parallel/test_fsdp.py
myleott: I'm not sure if we want to support nested configuration contexts, I could imagine it getting tricky to resolve conflicting options.

One exception is we may want to allow certain layers to override a limited set of options, e.g., MoE layers will use a dummy process group of size 1 to turn off sharding. Possibly some layers would want to turn off mixed precision. I suggest we can have this limited set of overrides be directly part of the "auto_wrap" (new name TBD) call. Something like: `FSDP.auto_wrap(moe_layer, process_group=custom_process_group)` or something like that.

What do you think?

SeanNaren:
Thanks guys!! I'll address all the points ASAP @min-xu-ai and I agree with your view @myleott! I'll change the names to reflect this

I got distracted into trying to auto-wrap recursively based on parameter sizes, and I think I nearly have it! Just some weirdness when precision=mixed (@myleott @min-xu-ai  for the nested FSDP wrappers, should mixed precision be set to True or just the outer layer?)

Probably some kinks to iron out but it looks like this on top of what's been added already:

```python
    @staticmethod
    def wrap_module(x: nn.Module, max_params, **kwargs):
        num_params = sum([p.numel() for p in x.parameters()])

        if len(list(x.named_children())) == 0:
            # If the module has no children, no need to recurse, wrap it if needed
            if num_params > max_params:
                return FullyShardedDataParallel.wrap(x, **kwargs), num_params
            return x, 0

        if num_params >= max_params:
            total_wrapped_params = 0
            # Iterate through the children, recursively wrap if necessary
            for name, module in x.named_children():
                wrapped_module, num_wrapped_params = FullyShardedDataParallel.wrap_module(module, max_params, **kwargs)
                setattr(x, name, wrapped_module)
                # Keep track of how many parameters have been wrapped
                total_wrapped_params += num_wrapped_params
            # decide if we need to wrap the current module, since the left over parameters exceed the number of params to wrap
            remainder = num_params - total_wrapped_params
            if remainder >= max_params:
                return FullyShardedDataParallel.wrap(x, **kwargs), num_params
            else:
                return x, total_wrapped_params
        return x, 0

    @staticmethod
    def auto_wrap(module, recursive=False, num_params=1e8, **kwargs):
        if FullyShardedAutoWrap.in_autowrap_context and recursive:
            wrapped_module, remainder = FullyShardedDataParallel.wrap_module(
                module,
                max_params=num_params,
                **kwargs
            )
            return wrapped_module
        return FullyShardedDataParallel.wrap(module, **kwargs)

    @staticmethod
    def wrap(module, **kwargs):
        if FullyShardedAutoWrap.in_autowrap_context:
            kwargs = {**FullyShardedAutoWrap.kwargs, **kwargs}
            return FullyShardedDataParallel(module, **kwargs)
        return module
```

This will allow you to set some number threshold and wrap all modules within greater than that, trying to wrap the deepest child first. Not sure if it makes sense and probably requires a small visualization to go with.

```python
with FSDP.enable_auto_wrap(**handleful_of_important_params):
    layer_1 = FSDP.auto_wrap(TransformerBlock(self.config), max_params=1e8, recursive=True)
```

min-xu-ai:
Thanks for the notes on OOM use case. I think it should go into the docstring.

Just exploring a bit here, I wonder if we can do away *without* an new function. Checkout here:

https://stackoverflow.com/questions/3209233/how-to-replace-an-instance-in-init-with-a-different-object

Basically, we can have FDSP() return the unwrapped class if it is passed with a `passthrough=True` option. Therefore, we have two modes:

* normal mode, DDP replacement

```
m = FSDP(m, options ...)
```

or manual nesting:

```
m = FSDP(FSDP(m, options ...), options...)
```

* maybe wrap mode

```
with nesting(options ...):
   FSDP(FSDP(m, overrides...), overrides...)
```

In the second mode, we have the `nesting()` context, which is a like a stack. outer FSDP will push their arguments on to the stack and inner one will take the top of the stack options and union with its own overrides and make an object. If the option `passthrough=True`, then it will just return the underlying object without construct a new one.

Does this simplify the user facing API? I think it will only need a single additional function called `nesting` for the context. No new wrapper function is needed?

min-xu-ai:
> How about just FSDP.wrap? Then later if we introduce some other functionality to automatically wrap layers we still have the option to name that new functionality auto_wrap.
>
> Re: context manager, maybe just with FSDP.enable_wrap(...)?

if the class.__new__ trick doesn't work, perhaps we can use consider `factory` and `enable_factory` since this is a pattern of object factory? Also, do these two functions need to be attached to the class?

min-xu-ai:
@SeanNaren you are amazing!

> for the nested FSDP wrappers, should mixed precision be set to True or just the outer layer?)

I have used mixed precision and nesting and inner ones and the outer one all have mixed_precesion=True and it seems to work fine in VISSL.

SeanNaren:
>```
> with nesting(options ...):
>    FSDP(FSDP(m, overrides...), overrides...)
> ```
> 
> In the second mode, we have the `nesting()` context, which is a like a stack. outer FSDP will push their arguments on to the stack and inner one will take the top of the stack options and union with its own overrides and make an object. If the option `passthrough=True`, then it will just return the underlying object without construct a new one.
> 
> Does this simplify the user facing API? I think it will only need a single additional function called `nesting` for the context. No new wrapper function is needed?

I was eyeing this as a solution as well, we could also detect we're in a context and do some magic, but I wasn't sure if we'd be up for that. At the minimum the current implementation works reasonable, but am curious to look down this path!

> @SeanNaren you are amazing!
> 
> > for the nested FSDP wrappers, should mixed precision be set to True or just the outer layer?)
> 
> I have used mixed precision and nesting and inner ones and the outer one all have mixed_precesion=True and it seems to work fine in VISSL.

Thanks @min-xu-ai  :) :) 

I've pushed the code as well as addressed a few code review changes. I need to get it working with my small reproducible script with precision set to mixed and I'll be able to benchmark, but if you could get a chance to use it would be awesome!

myleott:
> >```
> > with nesting(options ...):
> >    FSDP(FSDP(m, overrides...), overrides...)
> > ```
> > 
> > In the second mode, we have the `nesting()` context, which is a like a stack. outer FSDP will push their arguments on to the stack and inner one will take the top of the stack options and union with its own overrides and make an object. If the option `passthrough=True`, then it will just return the underlying object without construct a new one.
> > 
> > Does this simplify the user facing API? I think it will only need a single additional function called `nesting` for the context. No new wrapper function is needed?
>
> I was eyeing this as a solution as well, we could also detect we're in a context and do some magic, but I wasn't sure if we'd be up for that.

Yeah, I'd be reluctant to adopt the more magical interface, mostly because it will make the overrides more opaque. Consider the following example:
```python
class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.l1 = FSDP(nn.Linear(5, 10))  # does this imply the default options for FSDP, i.e., mixed_precision=False, flatten_parameters=True?

with nesting(mixed_precision=True, flatten_parameters=False):
    net = MyModule()  # will net.l1 have mixed precision? flatten parameters?
```

With a separate function it seems clearer that `wrap(**overrides)` only applies *overrides* to the config.

> do these two functions need to be attached to the class?

One less import for user code 😄

tests/nn/data_parallel/test_fsdp_context.py
```diff
@@ -0,0 +1,52 @@
+import torch.nn as nn
```
myleott, line 1: nit: please add the header

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,97 @@ def assert_state(self, state: TrainingState) -> None:
...
+                    self.l1 = FSDP.wrap(torch.nn.Linear(5, 5))
+                    # Wraps children modules based on bucket_size
+                    self.l2 = FSDP.auto_wrap(TransformerBlock(), bucket_size=1e8)
+
+        """
+        with FullyShardedAutoWrap(**kwargs):
+            yield
+
+    @staticmethod
+    def wrap(module, **kwargs):
```
myleott, line 1092: nit: thoughts on renaming `**kwargs` here to `**overrides`?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,97 @@ def assert_state(self, state: TrainingState) -> None:
...
+        FullyShardedAutoWrap.kwargs = {}
+
+    def __enter__(self):
+        self.enable_autowrap_context(self.kwargs)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.disable_autowrap_context()
+
+    @staticmethod
+    def recursive_wrap(x: nn.Module, bucket_size: Union[int, float], **kwargs) -> Tuple[nn.Module, int]:
```
myleott, line 1135: How about renaming to `min_num_params` or something similar, since `bucket` already refers to bucketing of the communication ops and is typically measured in MB.
```suggestion
    def recursive_wrap(x: nn.Module, bucket_size: Union[int, float], **kwargs) -> Tuple[nn.Module, int]:
        """
        Automatically wrap child modules of *x* that meet the given criteria with ``FSDP.wrap``.
        
        Args:
            min_num_params (int): min number of parameters for a child Module to be wrapped
        """
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,97 @@ def assert_state(self, state: TrainingState) -> None:
...
+        num_params = sum([p.numel() for p in x.parameters()])
+
+        if len(list(x.named_children())) == 0:
+            # If the module has no children, no need to recurse, wrap it if needed
+            if num_params > bucket_size:
```
myleott, line 1140: `>=`?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1069,6 +1069,97 @@ def assert_state(self, state: TrainingState) -> None:
             self.training_state == state
         ), f"expected to be in state {state} but current state is {self.training_state}"
 
+    @staticmethod
+    @contextlib.contextmanager
+    def enable_wrap(**kwargs) -> Iterator[None]:
+        """
+        Context manager to wrap modules in FullyShardedDataParallel.
+        Useful for when you'd like to apply the same parameters to all child modules that you wrap.
```
myleott, line 1077: Let's add more context about the use cases, how about:
```suggestion
        Context manager to wrap modules in FullyShardedDataParallel.
        
        Useful for when you'd like to apply the same parameters to all child modules
        that you wrap. A particularly important use case is wrapping large layers so
        that they get sharded (in-place) during initialization, to avoid running out of
        system memory.
```

myleott (changes requested):
Leaving a few suggestions, will start experimenting with this in the meantime

min-xu-ai:
> One less import for user code 😄

I see. :-) It is a trade-off: it will result in longer lines and (potentially alot) more typing when using in a file after the saving on the imports; just saying. :-)

myleott:
> I see. :-) It is a trade-off: it will result in longer lines and (potentially alot) more typing when using in a file after the saving on the imports; just saying. :-)

Ah, I misunderstood. You mean something like:
```python
from fairscale.nn.data_parallel import wrap

class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.l1 = wrap(nn.Linear(5, 10))
```

You're right, that's even less chars, let's definitely support that. Should be easy to support both actually:

```python
# in fully_sharded_data_parallel.py
def wrap(module, **overrides):
    return FullyShardedDataParallel.wrap(module, **overrides)
```

min-xu-ai:
> from fairscale.nn.data_parallel.fully_sharded_data_parallel import wrap

We can put it in __init__.py so, it will be
```
from fairscale.nn.data_parallel import wrap
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: for line 1103-1104, what will happen if outer wrap and inter wrap both have same override key but different values?

tests/nn/data_parallel/test_fsdp_context.py
min-xu-ai, line 1: is there a test here for recursive_wrap? It would be good to have a simple one for now, can extend as we see more use cases. it is tricky since we perhaps don't want to wrap inner module where the outer module reaches into the inner one during forward().

SeanNaren:
> > I see. :-) It is a trade-off: it will result in longer lines and (potentially alot) more typing when using in a file after the saving on the imports; just saying. :-)
> 
> Ah, actually you are right. You mean something like:
> 
> ```python
> from fairscale.nn.data_parallel.fully_sharded_data_parallel import wrap
> 
> class MyModule(nn.Module):
>     def __init__(self):
>         super().__init__()
>         self.l1 = wrap(nn.Linear(5, 10))
> ```
> 
> That's even less chars, let's definitely support that 😄 Should be easy to support both if we want, although maybe it's simpler to just decouple it from the FSDP class as you originally suggested. @SeanNaren, thoughts?

I think I understand the API, I've made the changes now it looks like:

```python
from fairscale.nn import enable_wrap, wrap, auto_wrap

class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.l1 = wrap(nn.Linear(5, 10))

with enable_wrap():
    module = MyModule()
```

Let me know if this looks good and the code lives in the right places. I have a few things kinks to iron out, as well as figure out why in Mixed Precision my benchmark test fails with auto wrap.

Thanks for the review guys!

tests/nn/data_parallel/test_fsdp_context.py
SeanNaren: I added a very rudimentary test! I think currently I see `auto_wrap` as providing very little flexibility (and a few assumptions that a sequential module has, i.e guarantees we're just calling the forward function) and `wrap` doing the most flexible of things.

The great thing is we can interchange/swap them as we see fit, i.e `auto_wrap` the modules we know are supported and do not do any child module magic other than call `forward`, and just use `wrap` for the cases where we'd not want to modify the child modules.

fairscale/nn/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 from .data_parallel import FullyShardedDataParallel, ShardedDataParallel
 from .misc import FlattenParamsWrapper
+from .misc.auto_wrap import auto_wrap, enable_wrap, wrap
```
SeanNaren, line 8: If I put `.auto_wrap import auto_wrap, enable_wrap, wrap` into `fairscale/nn/misc/__init__.py` it was complaining about a cycling import (FSDP). I'll try finnick around but let me know what's prefered

fairscale/nn/misc/auto_wrap.py
```diff
@@ -0,0 +1,148 @@
...
+        **wrap_overrides: Configuration overrides that will take
+            priority over the values provided by the ``enable_wrap`` context
+    """
+    if ConfigAutoWrap.in_autowrap_context:
+        wrap_overrides = {**ConfigAutoWrap.kwargs, **wrap_overrides}
+        return cls(module, **wrap_overrides)
+    return module
+
+
+def auto_wrap(module, min_num_params: float = 1e8, cls: Callable = FullyShardedDataParallel, **kwargs):
```
SeanNaren, line 64: If this works, we're going to need a more sensible default here, I just chose this randomish

myleott:
Awesome! Will review this shortly. CI is also complaining about some lint, would be good to run the `alias` that @min-xu-ai shared above.

Re: “why in Mixed Precision my benchmark test fails with auto wrap” is this referring to an external benchmark or a unit test? I can try to help debug if you can share more details

SeanNaren:
> Awesome! Will review this shortly. CI is also complaining about some lint, would be good to run the `alias` that @min-xu-ai shared above.
> 
> Re: “why in Mixed Precision my benchmark test fails with auto wrap” is this referring to an external benchmark or a unit test? I can try to help debug if you can share more details

Will do!

Regarding reproducing, I'm using Will's minGPT lightning repo to benchmark/test. If you don't mind installing lightning:

```bash
git clone https://github.com/SeanNaren/minGPT.git && cd minGPT && git checkout fully_sharded
pip install git+https://github.com/PyTorchLightning/pytorch-lightning.git#feat/fsdp # Install fork https://github.com/PyTorchLightning/pytorch-lightning/pull/6152


python benchmark.py --n_layer 1 --n_head 16 --n_embd 8192 --gpus 1 --limit_train_batches 120 --precision 16 # breaks
python benchmark.py --n_layer 1 --n_head 16 --n_embd 8192 --gpus 1 --limit_train_batches 120 --precision 32 # runs fine

```

SeanNaren:
@myleott I noticed that when we wrap children modules we'll call this part of the code every time:


```python
        if self.mixed_precision:
            args, kwargs = cast_inputs_to_fp16(*args, **kwargs)
```

Do we want to do this on the children modules as well? In minGPT the inputs are of type `torch.long` so they won't be converted, but the children modules will receive float I think

myleott:
Good call, we should change that to:
```
if self._is_root and self.mixed_precision:
```
That way it only happens on the outer-most FSDP instance

SeanNaren:
Thanks @myleott this fixes the error I was running into, I made a PR here with your suggestion: https://github.com/facebookresearch/fairscale/pull/452

myleott:
I have been running into a few other issues with auto-wrapping btw:
1) wrapping an `nn.ModuleList` breaks because FSDP doesn't implement `__len__` or `__iter__`. This is a bit tricky since for `__iter__` we'd actually want to expand the params before iterating, since we'll never call the FSDP.forward method.
2) shard params can be trickly. For example, an encoder-decoder architecture with shared embeddings on the encoder input, decoder input and decoder output. In this case wrapping the encoder and decoder separately can break sharing. I think we should ideally exclude shared params from `auto_wrap` and require these to be handled by an outer-most wrapper.

SeanNaren:
Nice! Thanks for testing this out, all this makes sense.

Blacklist certain functions that we cannot wrap, have this list pass through to the recursive wrap function, add a test?

I'm thinking we'll have a blacklist in FairScale to handle most modules, and allow the user to add their own overrides if necessary? What do you think?

```python
with enable_wrap(autowrap_blacklist=[MyModuleThatShouldntBeWrapped, ...]):
    ...
```

Regarding 2, Where in the docs should I add this? I think it would be nice to summarize all the info you added in the FSDP PR into a page, I could add the details there as well. If I get time I can do the first part as well, let me know

myleott:
> Regarding 2, Where in the docs should I add this? I think it would be nice to summarize all the info you added in the FSDP PR into a page, I could add the details there as well. If I get time I can do the first part as well, let me know

@sshleifer will be working on a more detailed README under fairscale/nn/data_parallel.

For now we can put it in the auto_wrap docstring. I will push some small local changes to this branch.

SeanNaren:
Hey @myleott added blacklisting, and will need to update doc strings. We might be able to do the blacklisting cleaner, but let me know what you think!

SeanNaren:
Closer to what was discussed I added an additional list to handle nodes we should not wrap but still wrap children, and nodes we shouldn't wrap including their children. This is `FSDP_MODULE_EXCLUDE_WRAP` and `FSDP_MODULE_BLOCKLIST` respectively.

Let me know if the names need changing and if we'd like to expose the additional list. thanks for your input here Myles! really appreciate the collab on this :)

fairscale/nn/misc/auto_wrap.py
```diff
@@ -0,0 +1,197 @@
...
+        **wrapper_kwargs: Configuration settings that will be passed to all ``wrap``
+            instances inside the context
+    """
+    with ConfigAutoWrap(module_blocklist, **wrapper_kwargs):
+        yield
+
+
+def wrap(
+    module: nn.Module,
+    cls: Callable = FullyShardedDataParallel,
```
min-xu-ai, line 51: making this general is really good. I now see why this is in nn.misc now.

To solve the circular import issue,  maybe just put this file in nn.wrap or nn.wrap_context or something like that?

Another way is to let
cls = None

then, in the function:

if cls is None:
  import FSDP
  cls = FSDP

This has a bit more overhead. but probably not on critical path. I don't know any other more elegant solutions. :-(

min-xu-ai (commented):
Like this new version a lot so far. Just want to send the comment on the circular import first.

fairscale/nn/misc/auto_wrap.py
```diff
@@ -0,0 +1,197 @@
...
+            # If the module has been blocklisted from wrapping, we return
+            return module, 0
+
+        num_params = sum([p.numel() for p in module.parameters()])
+
+        if len(list(module.named_children())) == 0:
+            # If the module has no children, no need to recurse, wrap it if needed
+            if num_params >= min_num_params and not isinstance(module, tuple(FSDP_MODULE_EXCLUDE_WRAP)):
+                return wrap(module, **kwargs), num_params
+            return module, 0
```
min-xu-ai, line 178: I don't know if this is needed. when there is 0 child, we take care of it on line 183 by not looping over the children list, right?

fairscale/nn/misc/auto_wrap.py
```diff
@@ -0,0 +1,197 @@
...
+import contextlib
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Union
+
+import torch.nn as nn
+
+from fairscale.nn.data_parallel.fully_sharded_data_parallel import FullyShardedDataParallel
+from fairscale.nn.misc import checkpoint_wrapper
+
+FSDP_MODULE_EXCLUDE_WRAP = [nn.ModuleList, nn.ModuleDict]
+FSDP_MODULE_BLOCKLIST = [nn.MultiheadAttention]
```
min-xu-ai, line 15:
```suggestion
# Modules that don't wrap.
FSDP_MODULE_EXCLUDE_WRAP = [nn.ModuleList, nn.ModuleDict]
# Modules that we don't recurse down to their children.
FSDP_MODULE_BLOCKLIST = [nn.MultiheadAttention]
```

tests/nn/misc/test_wrap.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
min-xu-ai, line 1: when you merge with master, you need to add this file to tests/ci_test_list1.txt to make sure one of the GPU testers will run it in CI. Should be simple one line change. You can run `bash tests/ci_test_list_check.sh` to verify locally.

If you move wrap into nn/wrap, please move this file accordingly as well.

min-xu-ai (approved):
This is turning out really really good. I like it a lot. Thanks again for the contribution! Approving with only minor suggestions.

fairscale/nn/misc/auto_wrap.py
SeanNaren: Think you're right, great observation!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -793,7 +793,8 @@ def _pre_backward_hook(*unused: Any) -> None:
             self._prep_grads_for_backward()
 
         def _register_hook(t: torch.Tensor) -> torch.Tensor:
-            t.register_hook(_pre_backward_hook)
+            if t.requires_grad:
+                t.register_hook(_pre_backward_hook)
```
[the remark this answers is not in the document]
SeanNaren: cc @myleott who I think added this

SeanNaren:
Any suggestion on default `min_num_params`? I set it to 100M as default which seemed reasonable, but was curious if you guys had any intuition around this.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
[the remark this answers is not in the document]
myleott: No, the `if` is actually a bugfix for modules that may return outputs that don't require grad. In this case pytorch will raise an exception when you call register_hook. I only discovered it because auto_wrap sometimes puts FSDP on module boundaries I hadn't tested before :)

myleott:
> Any suggestion on default min_num_params? I set it to 100M as default which seemed reasonable, but was curious if you guys had any intuition around this.

100M seems reasonable for now. I tested this on the 210M param Attention Is All You Need transformer and it's actually fastest not to have any child wrapping, but 100M causes the encoder and decoder to get wrapped separately, which is fine too (almost the same performance, and yields some memory savings).

myleott (approved):
LGTM too! Anything remaining or shall we merge this?

SeanNaren:
It's ready to go from my end! Thanks Myle :)

Issue #442, opened 2021-02-26
[FSDP] mixed precision buffers are only maintained in FP16


myleott:
## 🐛 Bug

In FSDP with `mixed_precision=True`, we only maintain buffers in FP16. This could lead to underflow/overflow for buffers that wouldn't happen with standard AMP training.

Ideally we could keep the buffers in their original dtype, but it's unclear how/when/if to cast to `compute_dtype`. For static buffers it doesn't matter, but for BatchNorm we want updates to happen in full precision.

sshleifer:
I have struggled to create a regression test where we get over/underflow unless buffers get put back into fp32 after `fwd`. 
The problem is that the dynamic buffer I tried, `torch.nn.BatchNorm1d().running_mean` gets updating during forward, so we have no overflow.  I think this is part of the definition of buffer (no `.grad`, so no updates during `optim.step()`, so the only reason for casting buffers out of `compute_dtype` is (a) state_dict and (b) maintaining the fp32 after forward invariant. 

Am I misunderstanding the problem @min-xu-ai @myleott ?

min-xu-ai:
Is over/underflow absolutely needed in the unit test or can we assert it is equal to non-FSDP training (DDP + AMP?) being sufficient?

> Ideally we could keep the buffers in their original dtype, but it's unclear how/when/if to cast to compute_dtype. For static buffers it doesn't matter, but for BatchNorm we want updates to happen in full precision.

For the statement above, I am don't quite understand "static buffers doesn't matter part". Also, BN.half() seems like a method users are supposed to call when they do mixed precision. Should we call it too?

myleott:
If using PyTorch amp, BatchNorm should stay/happen in FP32:
```python
import torch
from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP
from fairscale.utils.testing import DummyProcessGroup

bn = torch.nn.BatchNorm1d(10).cuda()

with torch.cuda.amp.autocast():
    input = torch.randn(20, 10).cuda()
    output = bn(input)
print(output.dtype)  # torch.float32

group = DummyProcessGroup(rank=0, size=1)
fsdp_bn = FSDP(bn, mixed_precision=True, process_group=group)
with torch.cuda.amp.autocast():
    input = torch.randn(20, 10).cuda()
    output = fsdp_bn(input)
print(output.dtype)  # torch.float16
```

sshleifer:
[...]

PR #449, opened 2021-02-28
[fix] Buffers, like params go back to fp32 after fwd

sshleifer:
## 🐛 Bug (copy pasted from issue)

In FSDP with `mixed_precision=True`, we only maintain buffers in FP16. This could lead to underflow/overflow for buffers that wouldn't happen with standard AMP training.

Ideally we could keep the buffers in their original dtype, but it's unclear how/when/if to cast to `compute_dtype`. For static buffers it doesn't matter, but for BatchNorm we want updates to happen in full precision.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -219,7 +219,7 @@ class FullyShardedDataParallel(nn.Module):
         return self._fsdp_wrapped_module  # note: may be a FlattenParamsWrapper instance
 
     @torch.no_grad()
-    def _all_buffers_to(self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None) -> None:
+    def _all_buffers_to(self, dtype: torch.dtype, device: Optional[torch.device] = None) -> None:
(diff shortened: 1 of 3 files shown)


myleott (approved):
Looks good!

Can you also add a test? Maybe add a small buffer to `DeviceAndTypeCheckModule` and then add some extra `expected_buffer_dtype` checks here: https://github.com/facebookresearch/fairscale/blob/master/tests/nn/data_parallel/test_fsdp.py#L153

sshleifer:
Done, asserted that buffers are `param_dtype` during forward and `torch.float32` after forward.
There is no need for `expected_buffer_dtype` given my implem.

myleott:
Ah, please see this comment: https://github.com/facebookresearch/fairscale/issues/442#issuecomment-787553247

I think we want `buffer_dtype` to stay fp32 inside the forward pass even when using mixed precision, whereas param_dtype would be fp16 in this case. This would match the behavior of O2 Apex/AMP: https://nvidia.github.io/apex/amp.html#o2-almost-fp16-mixed-precision

min-xu-ai (approved):
LGTM

sshleifer:
we went with `buffer_dtype`

PR #450, opened 2021-02-28
[wip] [cleanup] Broken batchnorm test

sshleifer:
Currently: some batchnorm tests pass, but params don't update.

### DIFF
--- a/tests/nn/data_parallel/test_fsdp.py
+++ b/tests/nn/data_parallel/test_fsdp.py
@@ -193,6 +193,14 @@ class TestComparisonToPyTorchDDP(DistributedTest):
         # Test every combination of these options:
         spawn_and_init(functools.partial(self._test_identical_outputs, TransformerWithSharedParams, config))
 
+    def test_transformer_batch_norm_breaks(self):
+        model_fn = functools.partial(TransformerWithSharedParams, add_bn=True)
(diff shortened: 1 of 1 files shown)


min-xu-ai:
Thanks for adding this!

min-xu-ai:
@sshleifer I wonder what's have been discovered here? So far, my plan is to use BN in its own FSDP wrapper (full precision, no flattening) for now. Do you think BN can be more generally supported under FSDP?

sshleifer:
>  Do you think BN can be more generally supported under FSDP?


- I think it's very likely to be possible and very unlikely to be easy. `running_mean`, which is a buffer, needs to be sharded.  More specifically, I expect that if the vertical model slice on a worker produces features of shape X, and the unsharded `bn.running_mean` is shaped X*num_workers, there will be an exception.
- Currently, FSDP does not shard buffers, and allows the caller to pass in a `buffer_dtype`, 
- I'd  google for other batch norm implementations  (not `nn.BatchNorm2d`) that support model parallelism.
- - I'd check if [deepspeed](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py) can handle it, (and if so how).

Sorry I can't be more helpful!

min-xu-ai:
This is super helpful already!

PR #451, opened 2021-03-01
[fix] Make state_dict all-gather FP32 params

### LINKED ISSUE #445: [FSDP] state_dict should all-gather FP32 params
## 🚀 Feature

When calling `state_dict` with `mixed_precision=True`, we all-gather the FP16 params as if we were doing a forward pass. We should add an option (or probably default behavior) to instead all-gather the FP32 weights for the state_dict.

After this is fixed we should ungate these tests:
https://github.com/facebookresearch/fairscale/blob/b6dc98cfe3365b9fbc7e386c2c246765a5b5710f/tests/nn/data_parallel/test_fsdp.py#L477-L478
https://github.com/facebookresearch/fairscale/blob/b6dc98cfe3365b9fbc7e386c2c246765a5b5710f/tests/nn/data_parallel/test_fsdp.py#L484-L485

myleott:
Fixes #445.
- `summon_params` now gathers in FP32
- allow `FlattenParamsWrapper.unflatten_params` to use a custom `flat_param`

Review note: the first commit (FlattenParamsWrapper change) is completely standalone, so it may be easier to review that first, then the second one (FSDP) after.


tests/nn/data_parallel/test_fsdp_uneven.py
```diff
@@ -70,7 +79,7 @@ def _test_func(rank, world_size, model, fsdp_config, tempfile_name, unused, test
 @skip_if_single_gpu
 @pytest.mark.parametrize("test_case", [{"inputs": [torch.rand(8, 3)], "assert_ref_out": True}])
 @pytest.mark.parametrize(
-    "fsdp_config", [{}, {"flatten_parameters": False}],
+    "fsdp_config", [{}, {"flatten_parameters": False}, {"mixed_precision": True}],
```
min-xu-ai, line 82: cool! thanks for adding this case!

tests/nn/data_parallel/test_fsdp_uneven.py
```diff
@@ -29,27 +29,36 @@ def _test_func(rank, world_size, model, fsdp_config, tempfile_name, unused, test
...
     optim = SGD(model.parameters(), lr=my_lr)
     inputs = test_case["inputs"]
     assert len(inputs) == 1 or not test_case["assert_ref_out"]
     assert len(inputs[0]) >= world_size
     for in_data in inputs:
-        in_data = Tensor(in_data[rank]).cuda()
+        in_data = Tensor(in_data[rank]).to(device, dtype)
         out = model(in_data)
-        out.sum().backward()
+        out.float().sum().backward()
```
min-xu-ai, line 61: curious why is float() needed?

tests/nn/data_parallel/test_fsdp_uneven.py
```diff
@@ -29,27 +29,36 @@ def _test_func(rank, world_size, model, fsdp_config, tempfile_name, unused, test
 
     my_lr = 0.1
 
+    device = torch.device("cuda")
+    if fsdp_config.get("mixed_precision", False):
+        dtype = torch.float16
+        fsdp_config["fp32_reduce_scatter"] = True
```
min-xu-ai, line 35: why is this needed?

tests/nn/data_parallel/test_fsdp_uneven.py
```diff
@@ -29,27 +29,36 @@ def _test_func(rank, world_size, model, fsdp_config, tempfile_name, unused, test
...
             ref_forward_output_my_rank = torch.matmul(v, weight)
             # Compute one iteration global weight update.
-            v = torch.Tensor(test_case["inputs"][0][:world_size]).cuda()
-            grad = v.sum(0).repeat(weight.shape[0], 1).div(world_size)
-            ref_weight_out = weight - grad.T * my_lr
-    model.to("cuda")
+            v = torch.Tensor(test_case["inputs"][0][:world_size]).to(device, dtype)
+            grad = v.float().sum(0).repeat(weight.shape[0], 1).div(world_size)
+            ref_weight_out = fp32_weight - grad.T * my_lr
+            assert ref_weight_out.dtype == torch.float32
```
min-xu-ai, line 50: it seems that you have update this code to mixed precision. Kudos!

min-xu-ai (commented):
not a review. took a curious look. just a few questions. will review when it is ready.

tests/nn/data_parallel/test_fsdp_uneven.py
myleott: This change and the floats() below are related to mixed precision.

FSDP with `mixed_precision=True` will by default sum the grads in FP16 (reduce-scatter), then cast to FP32 and update the FP32 model params.

If I implement this in the “ref” case, then one of the tests fails in CI for one world size due to a difference of 2e-5. This is because a lot of the underlying math is happening in FP16 (e.g., grad sum), but because the end result is FP32, the default atol/rtol in assert_allclose are for FP32 which is too sensitive.

We could either adjust atol/rtol, or what I did instead is switch to a more precise mixed precision recipe: We cast the model output to FP32 to compute the loss (in this case just a sum). We then tell FSDP to do the reduce-scatter in FP32 (to match the ref case). Now all the tests pass with the original atol/rtol.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -547,13 +556,24 @@ def summon_full_params(self, recurse: bool = True) -> Generator:
...
+                    # to the old FP16 storage (_full_param_padded), so we
+                    # unflatten manually with the new FP32 storage.
+                    assert len(self.params) == 1
+                    assert isinstance(self.module, FlattenParamsWrapper)
+                    stack.enter_context(self.module.unflatten_params(recurse=False, flat_param=self.params[0]))
+                try:
+                    yield
+                finally:
+                    stack.close()
+                    if not full_precision:
```
min-xu-ai, line 573:
```suggestion
                    # _rebuild_full_param() above didn't create full params. So no need to free here.
                    if not full_precision:
```

Maybe another way to do this:

full_param_allocated = _rebuild_full_params()
...

if full_param_allocated:
   _free_full_params()

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -952,33 +972,48 @@ def _wait_for_post_backward(self) -> None:
...
+                p_size = p._full_param_padded.size()
+                assert p_size.numel() % self.world_size == 0
+                if not self.mixed_precision or not full_precision:
+                    if p._full_param_padded.storage().size() != p_size.numel():
+                        # Allocate based on full size from all shards.
+                        alloc_storage_(p._full_param_padded, size=p_size)
+                    output_tensor = p._full_param_padded
+                else:
+                    # Allocate fresh tensor in full precision.
+                    output_tensor = p_data.new_zeros(p_size)
```
min-xu-ai, line 1008: these tensors will linger in GPU memory after use? Should we add a function to free them or make _free_full_params smarter in dealing with them?

min-xu-ai (approved):
nice! Some nonblocking comments.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
myleott: Good question, I should test this. My hope was that they there would only be one reference to them (in the returned `state_dict`), so it would be up to the caller to decide how/when to free them.

In particular, the `_use_fp32_param_shard` at the end of `summon_full_params` should remove the `p.data` reference added here, so the only remaining reference should be `state_dict`. But definitely worth testing :)

myleott:
Still working on this btw. I added some tests, both for memory usage and to confirm persistence of param changes after the context manager exits. Both tests uncovered issues. Will fix them here.

min-xu-ai:
> 
> 
> Added tests for memory usage (to confirm no leaks) and to confirm persistence of param changes after the context manager exits.

Cool. I can review more whenever it is ready.

tests/nn/data_parallel/test_fsdp_summon_full_params.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
```
min-xu-ai, line 1: pretty sure you need to merge with master and add this file to the CI list fie.

min-xu-ai (approved):
This is great. Sorry for the need of merge with master and add the new file. Otherwise, I would have just merged it.

Issue #454, opened 2021-03-01
FSDP Docs


sshleifer:
/assign @sshleifer

min-xu-ai:
@myleott @Vittorio-Caggiano @sshleifer, I think FSDP doesn't work with torch 1.5 due to lack of autocast. Should we mentioned that on the first page README.md next to the feature so that people won't rush to use it only to realize that they have an pytorch version that's too old? (in general fairscale states that we support 1.5, 1.6 and 1.7)

myleott:
We don’t actually rely on autocast do we? I think FSDP would work for FP32 training with 1.5, for example

min-xu-ai:
> We don’t actually rely on autocast do we? I think FSDP would work for FP32 training with 1.5, for example

well, we have a bunch of "skips" in the unit test so that it doesn't really run under 1.5.1 tests. :-(

min-xu-ai:
To be more concrete, we have an overly big skip that skips all test in test_fsdp.py. (my bad). But skip must be done in the parent process, not the spawned children when autocast value more readily available.

Maybe we should split the test file and put autocast related things only in a separate file?

myleott:
Actually nvm, `reduce_scatter` isn't available in PyTorch 1.5.1, we should just tell people it's not compatible.

min-xu-ai:
Maybe this likely needs to be in a separate issue, but noting down this idea for a tutorial:

For some model code, they init_weight after model is created. With DDP, this is fine, with FSDP, we likely need to have a that done within summon_full_params() context. A tutorial shows this would be nice.

Doc probably should call out this fact that any direct weight accesses outside of the fwd/bwd, should be in the summon context. :-)

PR #455, opened 2021-03-01
[docs] add fsdp_tips.rst

sshleifer:
This is a dumping ground to collect things we want to document about FSDP. 

Please comment with anything random that comes to mind.

Once there have been no changes to FSDP for 48 hrs (or some other proxy for stability), I will finalize this and set it to Ready For Review.


fairscale/nn/data_parallel/README_fsdp.md
```diff
@@ -0,0 +1,80 @@
...
+```
+will use an incorrect norm (the norm over all params in a shard) when clipping gradients.
+To overcome this, you can either call
+`sharded_module.clip_grad_norm(1.0)`
+which does the extra computation required to compute the norm properly, or use `torch.nn.utils.clip_grad_value_`.
+```
+
+
+
+#### Misc
```
min-xu-ai, line 77: should this include the new wrap/auto_wrap feature?

sshleifer: Yes leave desired `md` in comment else I can take a pass tomorrow.

min-xu-ai:
```
#### The `enable_wrap` context

There are two cases where the `enable_wrap` context can be useful:

* When you'd like to apply the same parameters to all child modules that you wrap with FSDP. Calling the `wrap` function within the said context will save you from passing the same set of FSDP parameters explicitly.
* When wrapping large models that does NOT fit within the CPU memory. I.e. you don't first create the full model and then traverse it to wrap it with FSDP at different parts. Instead, you create a wrapped instance of the model incrementally as you build up the model, allowing large modules to be sharded in-place.

example:
        with enable_wrap(**fsdp_params):
            # Wraps children modules by default based on min_num_params
            self.l2 = auto_wrap(TransformerBlock(), min_num_params=1e8)
```

min-xu-ai:
One more thing to be added to the doc:

```
If model weight initialization happens deterministically after sharding, the final weights will be comprised of N piece of identical shards. This would have negative effects on the model initial state.
```

@myleott @sshleifer Does the above looks good? If so, I can also add it to the doctoring of FSDP. We generate doc from the doctoring.

sshleifer:
These are not at all perfect, please feel free to push changes to this branch or comment.
I made a separate `fsdp_tips.rst` and cross-linked with `fsdp.rst` to avoid one crowding out the other on the same page. There may be a nicer way to do this in `rst`, (e.g. linking to section 2 at the top of 1 document).

sshleifer:
[fsdp_tips.html.zip](https://github.com/facebookresearch/fairscale/files/6091635/fsdp_tips.html.zip)

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,196 @@
...
+* FSDP enables training 13B parameter models on 8 GPUs and 175B parameter models on 128 GPUs. When using the ``cpu_offload=True`` option, it's possible to train 1T parameter models on 256 GPUs.
+
+
+General usage notes
+--------------------
+
+-  For best memory efficiency use ``auto_wrap`` to wrap each layer in your network with ``FSDP`` and set ``reshard_after_forward=True``
+-  For best training speed set ``reshard_after_forward=False`` (wrapping each layer is not required, but will improve speed further)
+-  If you're using ``torch.cuda.amp.autocast`` for mixed precision, that's fully compatible with the FSDP wrapper, just set ``mixed_precision=True``
+-  Uf combining with `activation checkpointing <https://github.com/facebookresearch/fairscale/blob/master/fairscale/nn/misc/checkpoint_activations.py>`__,
```
Vittorio-Caggiano, line 34: typo: `if`

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,196 @@
...
+
+The key insight to unlock full parameter sharding is that we can decompose the
+`all-reduce <https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#allreduce>`__
+operation in DDP into separate
+`all-gather <https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#allgather>`__
+and
+`reduce-scatter <https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#reducescatter>`__
+operations:
+
+.. |Figure 1| image:: https://user-images.githubusercontent.com/231798/108780259-26870a00-7536-11eb-890d-51720f39d098.png
```
Vittorio-Caggiano, line 56: we have a higher quality image with the same font/format as the next one, do you want to use that? 
image

Vittorio-Caggiano (commented):
just a small suggestion. 
Would it possible to add also a small tutorial?

myleott:
> Would it possible to add also a small tutorial?

Yes, this is a great idea 😄

Just to cross-reference and not lose track, Min also suggested the tutorial should include custom weight init inside a `summon_full_params` context: https://github.com/facebookresearch/fairscale/issues/454#issuecomment-792223593

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,200 @@
...
+
+    import torch
+    from fairscale.nn.wrap import auto_wrap, enable_wrap, wrap
+    from fairscale.nn.data_parallel import FullyShardedDataParallel
+    from fairscale.utils.testing import DummyProcessGroup
+    tfmr = torch.nn.Transformer(num_encoder_layers=2, num_decoder_layers=2)
+
+    group = DummyProcessGroup(rank=0, size=1)
+    fsdp_params = dict(mixed_precision=True, flatten_parameters=True)
+    with enable_wrap(process_group=group, **fsdp_params):
```
myleott, line 122: switch to new syntax:
```python
with enable_wrap(wrapper_cls=FSDP, ...)
```

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,200 @@
...
+-  Sharded parameters are downcast to ``fp16`` before ``forward``, promoted to ``fp32`` after forward.
+-  buffers are kept in ``fp16``, unless ``buffer_dtype=torch.float32`` is passed. Buffers are not sharded regardless of arguments.
+-  By default, gradients will be computed and reduced
+-  ``fp32_reduce_scatter=True`` controls the quantization of the gradient communication
+-  If ``torch.amp.autocast`` is enabled it will override the output dtypes of some operations, like ``BatchNorm2D``
+
+
+Auto-wrap
+~~~~~~~~~
+Auto wrapping sub-modules with ``FSDP`` is a convenient way to improve training speed by overlapping the all-gather step across the forward passes of different submodules.
```
myleott, line 108: Add:
```
It also improves memory efficiency by freeing gathered parameters after each layer finishes executing.
```

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,200 @@
...
+
+
+Mixed Precision
+---------------
+
+When ``mixed_precision=True``:
+
+-  Sharded parameters are downcast to ``fp16`` before ``forward``, promoted to ``fp32`` after forward.
+-  buffers are kept in ``fp16``, unless ``buffer_dtype=torch.float32`` is passed. Buffers are not sharded regardless of arguments.
+-  By default, gradients will be computed and reduced
```
myleott, line 101:
```
By default, gradients will be computed and reduced in fp16. If FP32 reductions are important, set ``fp32_reduce_scatter=True``
```

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,200 @@
...
+            backward pass for layer_i
+            discard full weights for layer_i
+            reduce-scatter gradients for layer_i
+
+Saving and Loading
+------------------
+
+There are two ways to load and save FSDP instances,
+
+- ``state_dict()`` returns a dictionary containing all parameters, which can be loaded with ``load_local_state_dict()``
```
myleott, line 90: nit: this should be `load_state_dict` at the end

docs/source/api/nn/fsdp_tips.rst
```diff
@@ -0,0 +1,200 @@
...
+Using CPU RAM
+-------------
+
+``move_grads_to_cpu`` and ``cpu_offload`` control which tensors get
+moved to CPU.
+
+-  ``cpu_offload`` moves weights to CPU when they are not being used.
+-  ``move_grads_to_cpu`` moves gradients to CPU. The use of this option
+   requires that the optimizer has a copy of the model parameters on
+   CPU.
```
myleott, line 155: How about:
```
``move_grads_to_cpu`` moves gradients to CPU so that the optimizer step also happens on CPU. This option requires ``cpu_offload=True``.
```

myleott (commented):
This looks great, thanks @sshleifer! I made some comments below, but I think we can ship this and iterate. What do you think @min-xu-ai?

min-xu-ai:
> I think we can ship this and iterate. What do you think @min-xu-ai?

Yeah, totally.

PR #457, opened 2021-03-02
[feat]: checkpoint and normalization

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


tests/nn/misc/test_checkpoint_activations_norm.py
```diff
@@ -0,0 +1,90 @@
...
+
+    if mixed_precision != "fp32":
+        in_data = in_data.half()
+
+    # Needed due to checkpointing.
+    in_data.requires_grad = True
+    for model in (m_ref, m_cpt):
+        optim = SGD(model.parameters(), lr=0.1)
+        if device == "cpu" and mixed_precision != "fp32":
+            # Got: RuntimeError: "batch_norm" not implemented for 'Half'.
```
min-xu-ai, line 74: these are probably to be expected as half not implemented on cpu for both LN and BN.

tests/nn/misc/test_checkpoint_activations_norm.py
```diff
@@ -0,0 +1,90 @@
...
+            with pytest.raises(RuntimeError):
+                out = model(in_data)
+            return
+        if device == "cuda" and norm_type == BatchNorm2d and mixed_precision == "fp16":
+            # Got: RuntimeError: expected scalar type Float but found Half.
```
min-xu-ai, line 79: this is interesting, if we just change param and buffer dtype, not call half(), it fails with this error on CUDA for BN.

fairscale/nn/misc/checkpoint_activations.py
```diff
@@ -63,25 +64,55 @@ def checkpoint_wrapper(module: nn.Module, offload_to_cpu: bool = False) -> nn.Mo
...
     # The use of weakref here is to prevent creating a ref cycle: m -> m.forward -> m.
     # When such cycle exists, gc won't collect the module when the module is freed.
     # That causes GPU memory to be leaked. See the unit test for how we catch that.
     #
     # We prefer this over a class wrapper since the class wrapper would have to
     # proxy a lot of fields and methods.
-    module.forward = functools.partial(_checkpointed_forward, type(module).forward, weakref.ref(module), offload_to_cpu)  # type: ignore
+    module.forward = functools.partial(  # type: ignore
+        _checkpointed_forward, type(module).forward, weakref.ref(module), offload_to_cpu, pre_forward, post_forward
+    )
```
myleott, line 93: can we directly `register_forward_pre_hook` and `register_forward_hook` on the module instead of piping these through the checkpointing method?

fairscale/nn/misc/checkpoint_activations.py
```diff
@@ -63,25 +64,55 @@ def checkpoint_wrapper(module: nn.Module, offload_to_cpu: bool = False) -> nn.Mo
...
+
+    def pre_forward() -> None:
+        for bn_and_bool in bn_list:
+            bn_and_bool[1] = bn_and_bool[0].track_running_stats  # type: ignore
+            bn_and_bool[0].track_running_stats = False  # type: ignore
+
+    def post_forward() -> None:
+        for bn_and_bool in bn_list:
+            assert bn_and_bool[1] is not None
+            bn_and_bool[0].track_running_stats = bn_and_bool[1]  # type: ignore
```
myleott, line 83: This actually seems like a general problem, I wonder if we can make something more standalone/reusable? In particular, I feel that BatchNorm layers should not update stats when in `torch.no_grad()` mode. Currently they only respect `model.eval()`, but perhaps we can have a simple module-level wrapper that registers the relevant hooks?

Something like (untested):
```python
def patch_batch_norm(module: nn.Module):
    if isinstance(module, _BatchNorm):
        module.register_forward_pre_hook(bn_pre_hook)
        module.register_forward_hook(bn_post_hook)
    return module

def bn_pre_hook(module, input):
        module._backup_track_running_stats = module.track_running_stats
        module.track_running_stats = False

def bn_post_hook(module, input, output):
        module.track_running_stats = module._backup_track_running_stats
```

What do you think?

anj-s:
@min-xu-ai Can you give me some context about what the feature is doing? Is it for supporting BN?

fairscale/nn/misc/checkpoint_activations.py
min-xu-ai: Great idea! Thanks for suggesting. Let me give it a try. I am a bit worries that users may want to keep the behavior of tracking stats under no_grad() but that's probably a very rare thing if at all possible. The current method only affect checkpointed fwd pass.

min-xu-ai:
> @min-xu-ai Can you give me some context about what the feature is doing? Is it for supporting BN?

Thanks for taking a look, @anj-s!

Yes, this change stems from that we observed that something is off for BN under activation checkpointing (not just the version in fairscale, also under the version in pytorch). This is an attempt to workaround that.

We also observed some differences in how to we fp16 for BN affects its behavior. So the unit test includes that as well.

tests/nn/misc/test_checkpoint_activations_norm.py
min-xu-ai: @myleott What do you think about this? I guess we should workaround that too in BN?

min-xu-ai: cc @sshleifer too.

min-xu-ai:
@myleott @anj-s , I think this is ready. Mind taking another look? Note, the benchmark CI tests are currently failing due to a networking error. Will have to deal with them separately.

tests/nn/misc/test_checkpoint_activations_norm.py
```diff
@@ -0,0 +1,90 @@
...
+
+    if mixed_precision == "fp16":
+        # Set param.data and buffers as fp16
+        for p in model.parameters():
+            p.data = p.data.half()
+        for m in model:
+            for n, b in m.named_buffers():
+                setattr(b, n, b.half())
+
+    if mixed_precision == "call_half":
```
myleott, line 39: not: make this `elif`?

tests/nn/misc/test_checkpoint_activations_norm.py
```diff
@@ -0,0 +1,90 @@
...
+
+    model = Sequential(Linear(3, 2), norm_type(2))
+
+    if mixed_precision == "fp16":
+        # Set param.data and buffers as fp16
+        for p in model.parameters():
+            p.data = p.data.half()
+        for m in model:
+            for n, b in m.named_buffers():
+                setattr(b, n, b.half())
```
myleott, line 37: shouldn't this be `setattr(m, n, b.half())`?

tests/nn/misc/test_checkpoint_activations_norm.py
myleott: I think this is a bug in the test (see comment above)

myleott (approved):
Nice! Small fix needed, otherwise LGTM

tests/nn/misc/test_checkpoint_activations_norm.py
min-xu-ai: yes! what's a dumb bug!

min-xu-ai:
test bug fixed and addressed comments. Will merge if everything passes.

PR #460, opened 2021-03-03
[fix][minor] Change empty shard handling for OSS, do not rely on asserts

blefaudeux:
# Before submitting

## What does this PR do?
If a shard is empty, do not assert out but skip the broadcasting step indeed. The prior issue was that if somebody was using "python -O" on a distributed job, this would just hang.
Now if some ranks are empty, they just don't participate in the optimization problem (not updating any tensor), that's all (they can still participate in the data parallel part), which is probably a better take.

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -140,13 +140,6 @@ class OSS(Optimizer):
                     param_group_rank["params"] = params
                     self._partition_parameters[rank].append(param_group_rank)
 
-            assert min(sum(len(pg["params"]) for pg in partition) for partition in self._partition_parameters) > 0, (
-                "One or more empty shards detected, the world size is too big or the model too small.\n"
-                + "Please reduce your world size if this is the model you would like to train\n"
-                + f"Current world size: {self.world_size}\n"
-                + "Current number of parameters: {}".format(sum(len(pg["params"]) for pg in self.param_groups))
-            )
-
         return self._partition_parameters
 
     @property
@@ -552,8 +545,11 @@ class OSS(Optimizer):
 
         for device in self.buckets.keys():
             for src_rank, bucket in enumerate(self.buckets[device]):
-                global_src_rank = self.get_global_rank(self.group, src_rank)
-                last_work_handle = dist.broadcast(tensor=bucket, src=global_src_rank, group=self.group, async_op=True)
+                if bucket.numel() > 0:
+                    global_src_rank = self.get_global_rank(self.group, src_rank)
+                    last_work_handle = dist.broadcast(
+                        tensor=bucket, src=global_src_rank, group=self.group, async_op=True
+                    )
 
         # Only check on the last handle, they're all inlined on the same CUDA stream
         if last_work_handle:
@@ -597,4 +593,4 @@ class OSS(Optimizer):
                     else:
                         self.buckets[device][dst_rank] = bucket
                 else:
-                    self.buckets[device].append(torch.zeros(1, device=device))
+                    self.buckets[device].append(torch.zeros(0, device=device))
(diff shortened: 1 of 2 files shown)


fairscale/optim/oss.py
```diff
@@ -597,4 +593,4 @@ def _setup_flat_buffers(self) -> None:
                     else:
                         self.buckets[device][dst_rank] = bucket
                 else:
-                    self.buckets[device].append(torch.zeros(1, device=device))
+                    self.buckets[device].append(torch.zeros(0, device=device))
```
blefaudeux, line 596: an empty tensor is a thing, better take to catch empty shards

fairscale/optim/oss.py
```diff
@@ -552,8 +545,11 @@ def _broadcast_params(self) -> None:
 
         for device in self.buckets.keys():
             for src_rank, bucket in enumerate(self.buckets[device]):
-                global_src_rank = self.get_global_rank(self.group, src_rank)
-                last_work_handle = dist.broadcast(tensor=bucket, src=global_src_rank, group=self.group, async_op=True)
+                if bucket.numel() > 0:
```
blefaudeux, line 548: not sure about broadcasting something empty for all backends, and does not make a ton of sense, so just skip that

fairscale/optim/oss.py
```diff
@@ -140,13 +140,6 @@ def partition_parameters(self) -> List[List[dict]]:
                     param_group_rank["params"] = params
                     self._partition_parameters[rank].append(param_group_rank)
```
blefaudeux, line 142: changing the logic: if a rank has an empty shard, it will be taken care of bucket-wise, no more asserting since that prove fragile on fb infra (the job can be compiled without the asserts)

min-xu-ai (approved):
very nice. running with python -O is LoL. If that speed things up then the program is probably running too much python code. :-)

tests/optim/test_oss.py
```diff
@@ -262,19 +262,22 @@ def test_zero_grad():
     mp.spawn(run_test_zero_grad, args=(world_size, temp_file_name), nprocs=world_size, join=True)
 
 
-def run_test_catch_empty_shardd(rank, world_size, tempfile_name):
+def run_test_empty_shardd(rank, world_size, tempfile_name):
     dist_init(rank, world_size, tempfile_name, backend="gloo")
```
anj-s, line 266: Can we test this on nccl as well?

tests/optim/test_oss.py
```diff
@@ -262,19 +262,22 @@ def test_zero_grad():
...
+    o.step()
 
     dist.destroy_process_group()
 
 
 def test_empty_shard():
     world_size = 4
 
-    mp.spawn(run_test_catch_empty_shardd, args=(world_size, tempfile.mkstemp()[1]), nprocs=world_size, join=True)
+    mp.spawn(run_test_empty_shardd, args=(world_size, tempfile.mkstemp()[1]), nprocs=world_size, join=True)
```
anj-s, line 280: typo:extra d?

fairscale/optim/oss.py
anj-s: maybe add this same line as a comment for better code readability?

fairscale/optim/oss.py
anj-s: Just to clarify: this happens when one of the nodes does not have any params assigned to it? Thats an empty shard right?

fairscale/optim/oss.py
blefaudeux: hmm ok, the comment was about the change actually but if that helps I can always write something. The if clause is if(no params) so I thought that was kind of clear (outside of this very narrow PR view)

PR #463, opened 2021-03-03
[fix] wait for all handles.. supposed to be safer with gloo

blefaudeux:
# Before submitting

## What does this PR do?
Should not be needed with NCCL, but seems that Gloo does not offer the same guarantees. Hopefully fixing sporadic issues like [this](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1984/workflows/0353c715-d633-4b6c-8797-29de335ff212/jobs/11530)

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -626,17 +626,19 @@ class ShardedDataParallel(nn.Module):
     # Flush all the buckets, just in case
     def _flush_buckets(self) -> None:
         if self._bucket_list is not None:
-            last_handle = None
+            handles = []
             for bucket in self._bucket_list:
                 if not bucket.sent:
                     # Normalize the bucket in one go
                     bucket.buffer.mul_(self.world_size_scaling)
 
                     # Reduce the bucket
-                    last_handle = dist.reduce(
-                        tensor=bucket.buffer, dst=bucket.destination, group=self.process_group, async_op=True,
+                    handles.append(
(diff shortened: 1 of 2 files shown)


anj-s (approved):
Thanks for fixing this flaky test so quickly!

Do you know why this happens for 1.5 and not for the later versions? Just wondering if it is something specific that was fixed or that we were just unlucky to see it fail at this version.

blefaudeux:
> Thanks for fixing this flaky test so quickly!
> 
> Do you know why this happens for 1.5 and not for the later versions? Just wondering if it is something specific that was fixed or that we were just unlucky to see it fail at this version.

it does not seem to properly fix it actually :/ I'm guessing that this could be a bug in gloo actually, since it only happens in this configuration (this torch version + this python version) as far as I can see, but a bit wary of switching off the unit test for that reason :/

min-xu-ai:
> 
> 
> > Thanks for fixing this flaky test so quickly!
> > Do you know why this happens for 1.5 and not for the later versions? Just wondering if it is something specific that was fixed or that we were just unlucky to see it fail at this version.
> 
> it does not seem to properly fix it actually :/ I'm guessing that this could be a bug in gloo actually, since it only happens in this configuration (this torch version + this python version) as far as I can see, but a bit wary of switching off the unit test for that reason :/

were you able to reproduce it consistently locally with a similar python + torch version or it is only the CI? I think it is OK if we detect some combination of version doesn't work, then in the doc (and in the code) give user a warning or runtime check when such combination is detected.

PR #464, opened 2021-03-03
[chore] move a test script and a CI test improvement

min-xu-ai:
move a test script so that top level dir is less crowded.


anj-s (approved):

anj-s (approved):
Do we need to change something that references this?

min-xu-ai:
> 
> 
> Do we need to change something that references this?

Oh, I should've mentioned in the first place. I don't think anything refer to it. It is likely a script used on interactively for testing.

.circleci/config.yml
```diff
@@ -116,6 +116,9 @@ install_dep_180: &install_dep_180
   - run:
       name: Install Dependencies with torch 1.8.0 nightly
       command: |
+        # check if we have restored cache correctly, if so, just skip
+        python -c 'import torch; assert torch.__version__.split(".")[:2] != ["1", "8"]' || exit 0
```
anj-s, line 120: Should this be in the same PR? Or is this related?

min-xu-ai: Yeah, CI run for this PR hit a failure related to venv installation. I decide to improve on that here. Ideally should split the PR, but it is very small change.

min-xu-ai:
Thanks @anj-s for quick review!

min-xu-ai:
@blefaudeux , it seems that we hit network error from time to time when the benchmark is download dataset or something?

https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1999/workflows/78eeac83-1cef-4977-8c82-e3288f9b1d76/jobs/11752

perhaps we should download it once in the cache dir and then load it from the CI cache?

min-xu-ai:
> 
> 
> @blefaudeux , it seems that we hit network error from time to time when the benchmark is download dataset or something?
> 
> https://app.circleci.com/pipelines/github/facebookresearch/fairscale/1999/workflows/78eeac83-1cef-4977-8c82-e3288f9b1d76/jobs/11752
> 
> perhaps we should download it once in the cache dir and then load it from the CI cache?

I could download from this URL: http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz from my browser. But I think the server must blocked CI's IP or something. It always is error 403.

blefaudeux:
Yes I just shared that with Anjali, that's new. We used to try a couple of times if needed but ai think that this changed

min-xu-ai:
> 
> 
> Yes I just shared that with Anjali, that's new. We used to try a couple of times if needed but ai think that this changed

Here are alternative URLs that seems to work:

https://github.com/awslabs/dynamic-training-with-apache-mxnet-on-aws/blob/master/cpp-package/example/get_data.sh

Maybe you can use them for now?

Longer term, it is better to cache the data in the setup steps and put it in the cache. But that can be done later. Right now, this is blocking all PRs I think. :-(

PR #467, opened 2021-03-04
[refactor] enhance wrap and auto_wrap

min-xu-ai:
- Two things were done in this PR
  1. We don't need to import FSDP in wrap.py since the wrapper class
     type is stored in the context now.
  2. We can use a `should_wrap` function to customize wrapping policy
     for auto_wrap, including size of module, blacklist, exclude list
- The auto_wrap function got simplified a bit as a minor side effect.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


min-xu-ai:
cc @SeanNaren. I will comment on the diff to make it easier to review.

fairscale/nn/__init__.py
```diff
@@ -4,20 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from .data_parallel import FullyShardedDataParallel, ShardedDataParallel
-from .misc import FlattenParamsWrapper
+from .misc import FlattenParamsWrapper, checkpoint_wrapper
 from .moe import MOELayer, Top2Gate
 from .pipe import Pipe, PipeRPCWrapper
-from .wrap import auto_wrap, enable_wrap, wrap
-
-__all__ = [
```
min-xu-ai, line 12: I remove those since it seems unnecessary when we basically export everything imported.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -34,26 +83,25 @@ def enable_wrap(module_blocklist: Optional[List] = None, **wrapper_kwargs: Any)
...
             # Wraps layer in FSDP by default if within context
             self.l1 = wrap(torch.nn.Linear(5, 5))
-            # Wraps children modules by default based on min_num_params
-            self.l2 = auto_wrap(TransformerBlock(), min_num_params=1e8)
+            # Wraps children modules based on a different min_num_params
+            my_should_wrap = functools.partial(should_wrap, min_num_params=1e7)
+            self.l2 = auto_wrap(TransformerBlock(), shuold_wrap=my_should_wrap)
 
     Args:
-        module_blocklist: List of additional Module Classes to not wrap when
```
min-xu-ai, line 41: module blocklist is no longer needed since we can do all customization with should_wrap now.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -130,68 +175,85 @@ class ConfigAutoWrap:
...
+            module (nn.Module):
+                module to recursively wrap
+            should_wrap (Callable, Optional):
+                optionally, override the :func:`should_wrap` from the context.
+
+        Returns:
+            (nn.Module, int):
+                Wrapped module and the number parameters wrapped recursively.
         """
-        if isinstance(module, tuple(ConfigAutoWrap.module_blocklist)):
```
min-xu-ai, line 175: this if is not needed since the first should_wrap decision takes care of it now.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -130,68 +175,85 @@ class ConfigAutoWrap:
...
+        # Make sure no child is not already wrapped.
+        for _, child in module.named_modules():
+            assert not isinstance(child, cast(type, ConfigAutoWrap.wrapper_cls))
 
+        # We count all params, assuming none of them is already wrapped.
         num_params = sum([p.numel() for p in module.parameters()])
 
-        if num_params >= min_num_params:
+        assert should_wrap is not None
+        if should_wrap(True, module, num_params):
```
min-xu-ai, line 242: this should_wrap decides if a module should be broken and wrapped.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -130,68 +175,85 @@ class ConfigAutoWrap:
...
+                    module=child, should_wrap=should_wrap, **kwargs
                 )
                 setattr(module, name, wrapped_child)
                 # Keep track of how many parameters have been wrapped
                 total_wrapped_params += num_wrapped_params
             # decide if we need to wrap the current module,
             # since the left over parameters exceed the number of params to wrap
             remainder = num_params - total_wrapped_params
-            if remainder >= min_num_params and not isinstance(module, tuple(FSDP_MODULE_EXCLUDE_WRAP)):
+            if should_wrap(False, module, remainder):
```
min-xu-ai, line 255: this should_wrap decides if a module should be wrapped.

tests/nn/wrap/test_wrap.py
```diff
@@ -25,7 +26,7 @@ def setUp(self) -> None:
         self.process_group = DummyProcessGroup(rank=0, size=1)
 
     def test_wrap(self):
-        with enable_wrap(flatten_parameters=False, process_group=self.process_group):
+        with enable_wrap(wrapper_cls=FSDP, flatten_parameters=False, process_group=self.process_group):
```
min-xu-ai, line 29: user specify the wrapper class here now.

tests/nn/wrap/test_wrap.py
```diff
@@ -45,11 +46,12 @@ def test_auto_wrap(self):
         Test to ensure with auto wrap, we wrap child modules correctly based on the min_num_params.
         ``nn.Linear(5, 5)`` does not exceed the bucket size, but combined they do.
         """
-        with enable_wrap(process_group=self.process_group, flatten_parameters=False):
+        with enable_wrap(wrapper_cls=FSDP, process_group=self.process_group, flatten_parameters=False):
             sequential = nn.Sequential(
                 nn.Linear(5, 5), nn.Linear(5, 5), nn.Sequential(nn.Linear(5, 5), nn.Linear(5, 5))
             )
-            model = auto_wrap(sequential, min_num_params=40)
+            my_should_wrap = functools.partial(should_wrap_default, min_num_params=40)
```
min-xu-ai, line 53: user can override should_wrap with auto_wrap

tests/nn/wrap/test_wrap.py
```diff
@@ -74,31 +77,38 @@ def test_auto_wrap_preset_exclude_wrap_include_children(self):
...
         """
-        Test to ensure blocklisted modules are not wrapped.
+        Test to ensure blacklisted modules are not wrapped.
         """
-        with enable_wrap(module_blocklist=[nn.Linear], process_group=self.process_group, flatten_parameters=False):
+        my_should_wrap = functools.partial(
+            should_wrap_default, min_num_params=40, blacklist=should_wrap_default.MODULE_BLACKLIST.union({nn.Linear})
+        )
+        with enable_wrap(
+            should_wrap=my_should_wrap, wrapper_cls=FSDP, process_group=self.process_group, flatten_parameters=False
```
min-xu-ai, line 108: user can also override should_wrap with the context

tests/nn/wrap/test_wrap.py
```diff
@@ -74,31 +77,38 @@ def test_auto_wrap_preset_exclude_wrap_include_children(self):
...
 
-    def test_auto_wrap_preset_blocklist_custom(self):
+    def test_auto_wrap_preset_blacklist_custom(self):
         """
-        Test to ensure blocklisted modules are not wrapped.
+        Test to ensure blacklisted modules are not wrapped.
         """
-        with enable_wrap(module_blocklist=[nn.Linear], process_group=self.process_group, flatten_parameters=False):
+        my_should_wrap = functools.partial(
+            should_wrap_default, min_num_params=40, blacklist=should_wrap_default.MODULE_BLACKLIST.union({nn.Linear})
```
min-xu-ai, line 105: user can union() add more modules to the blacklist or the exclusion list.

min-xu-ai:
> 
> 
> cc @SeanNaren. I will comment on the diff to make it easier to review.

I think this may cause some grief on your integration effort with lightning. Apologize in advance. I didn't think about those changes until today. I hope they are small enough changes and longer term better for the API in terms of being more generic (wrapper is more pure) and more customizable (more flexible should_wrap).

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -34,26 +83,25 @@ def enable_wrap(module_blocklist: Optional[List] = None, **wrapper_kwargs: Any)
...
+            Custom function to control how to do :func:`auto_wrap`. This is
+            useful to exclude unsupported modules or wrap based on sizes when
+            wrapping recursively.
+            (default: :func:`should_wrap_default`)
+        **wrapper_kwargs:
+            Configuration settings that will be passed to all ``wrap``
             instances inside the context
     """
-    with ConfigAutoWrap(module_blocklist, **wrapper_kwargs):
+    with ConfigAutoWrap(should_wrap_default, **wrapper_kwargs):
```
SeanNaren, line 100:
```suggestion
    with ConfigAutoWrap(should_wrap, **wrapper_kwargs):
```

SeanNaren (approved):
> I think this may cause some grief on your integration effort with lightning. Apologize in advance. I didn't think about those changes until today. I hope they are small enough changes and longer term better for the API in terms of being more generic (wrapper is more pure) and more customizable (more flexible should_wrap).

Absolutely no worries here, want to get the integration right so I can make the changes! Thanks so much for refactoring this, it's starting to look really nice.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -4,23 +4,72 @@
...
+def should_wrap_default(
+    should_recurse: bool,
+    module: nn.Module,
+    unwrapped_params: int,
+    # These are customizable for this default policy function.
+    min_num_params: int = int(1e8),
+    blacklist: Optional[Set] = None,
+    exclude_wrap: Optional[Set] = None,
+) -> bool:
+    """Default policy function on with :func:`auto_wrap`.
```
myleott, line 23: nit: replace "function on with ..." -> "function for ..."

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -4,23 +4,72 @@
...
 
-from fairscale.nn.data_parallel.fully_sharded_data_parallel import FullyShardedDataParallel
 from fairscale.nn.misc import checkpoint_wrapper
 
-# Modules that don't wrap.
-FSDP_MODULE_EXCLUDE_WRAP = {nn.ModuleList, nn.ModuleDict}
-# Modules that we don't recurse down to their children.
-FSDP_MODULE_BLOCKLIST = {nn.MultiheadAttention}
+
+def should_wrap_default(
```
myleott, line 14: what do you think about renaming this slightly to make the intent clearer, e.g., `default_should_wrap_policy`?

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -4,23 +4,72 @@
...
+
+       Return if a module should be wrapped during :func:`auto_wrap`.
+
+       The first three parameters are used by :func:`auto_wrap`. If
+       you write a custom version of this policy function, your version
+       needs to at least accept the first three parameters and free
+       to do whatever you want in the function.
+
+    Args:
+       should_recurse (bool):
```
myleott, line 33: let's just call this `recurse`, this is what PyTorch calls similar args (e.g., in `nn.Module.parameters(recurse=True)`)

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -4,23 +4,72 @@
...
+           on whether we should wrap the said module.
+       module (nn.Module):
+           The module to be considered in this decision.
+       unwrapped_params (int):
+           The number of parameters yet to be wrapped in this module.
+
+       min_num_params (int):
+           Customizable policy input. It controls the size threshold
+           on how big should a module be to be considered wrapped.
+       blacklist (set):
```
myleott, line 46: this needs to be renamed, there's a larger effort to remove blacklist/whitelist terminology across the company (more context, internal link only: https://fburl.com/z8pu53eq)

I also find the naming a bit unclear w.r.t. the `exclude_wrap` option.

How about we rename this to `force_leaf_modules`:
```python
force_leaf_modules (Set[Type[nn.Module]]): set of module types to
    keep as leaves, i.e., their children will never be wrapped.
```

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -34,26 +83,25 @@ def enable_wrap(module_blocklist: Optional[List] = None, **wrapper_kwargs: Any)
...
         yield
 
 
-def wrap(
-    module: nn.Module,
-    cls: Callable = FullyShardedDataParallel,
-    activation_checkpoint: bool = False,
-    **wrap_overrides: Any
-) -> nn.Module:
+def wrap(module: nn.Module, activation_checkpoint: bool = False, **wrap_overrides: Any) -> nn.Module:
```
myleott, line 104: similar to how you moved `cls` to `enable_wrap`, should we remove `activation_checkpoint`? In fairseq I found it more intuitive to do this in the context manager too.

For example:
```python
def my_wrapper(module, **wrap_overrides):
    return FSDP(activation_checkpoint(module), **wrap_overrides)

with enable_wrap(my_wrapper, ...):
    (...)
```

Btw, activation checkpointing works with FSDP as long as it's aligned to FSDP boundaries. This makes me wonder if we should add a `checkpoint_activations` convenience flag to FSDP. Then one could do:
```python
with enable_wrap(FSDP, checkpoint_activations=True):
    (...)
```

myleott (changes requested):
I like this a lot! Let's merge this soon so we can start to stabilize the API

fairscale/nn/__init__.py
msbaines: This is not accurate. __all__ affects the behavior of 'import *'. I would recommend restoring it.

https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: I am fine either way. I slightly worry that we marry FSDP and checkpoint too early and removing some flexibility since people can do checkpointing inside a FSDP without wrapping. So, let me just first move checkpointing option into the context?

myleott: Yep makes sense

fairscale/nn/__init__.py
min-xu-ai: Thanks for taking a look! I can do that. OTOH, We don't maintain this consistently though. Also, is "import *" a good thing to have?

msbaines: Not sure. It might be nice to properly support `import *` but I don't have strong opinions:

https://twitter.com/jeremyphoward/status/1233062899406405636

msbaines: We should really document what should be in `__all__` and what we should be importing via `__init__.py`. I've been trying to follow PyTorch conventions but we should make those conventions explicit in the CONTRIBUTING guide.

min-xu-ai: Good points. I will restore it back here. Let me open an issue for further discussion? I think it is slightly better not to support import *. I will provide more details in the issue.

min-xu-ai: put it back and cleaned it up for now.

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: done.

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: done. I must have been debating between "on" and "with". Lol.

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: yeah, better.

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: done. also changed exclude_wrap to exclude_wrap_modules to be similar to force_leaf_modules. Also, renamed the two constants accordingly.

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: done. I like this a lot.

min-xu-ai:
Sorry for the delay. I think this is ready.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -92,16 +97,16 @@ def enable_wrap(should_wrap: Optional[Callable] = None, **wrapper_kwargs: Any) -
...
             useful to exclude unsupported modules or wrap based on sizes when
             wrapping recursively.
-            (default: :func:`should_wrap_default`)
+            (default: :func:`default_should_wrap_policy`)
         **wrapper_kwargs:
             Configuration settings that will be passed to all ``wrap``
             instances inside the context
     """
-    with ConfigAutoWrap(should_wrap_default, **wrapper_kwargs):
+    with ConfigAutoWrap(default_should_wrap_policy, **wrapper_kwargs):
```
min-xu-ai: this bug is fixed in the merge commit.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -92,16 +97,16 @@ def enable_wrap(should_wrap: Optional[Callable] = None, **wrapper_kwargs: Any) -
...
         yield
 
 
-def wrap(module: nn.Module, activation_checkpoint: bool = False, **wrap_overrides: Any) -> nn.Module:
+def wrap(module: nn.Module, **wrap_overrides: Any) -> nn.Module:
```
min-xu-ai, line 109: no more activation_checkpoint, very clean.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -130,68 +169,85 @@ class ConfigAutoWrap:
...
         """
-        if isinstance(module, tuple(ConfigAutoWrap.module_blocklist)):
-            # If the module has been blocklisted from wrapping, we return
-            return module, 0
+        if should_wrap is None:
+            should_wrap = ConfigAutoWrap.should_wrap
+
+        # Make sure no child is not already wrapped.
+        for _, child in module.named_modules():
+            assert not isinstance(child, cast(type, ConfigAutoWrap.wrapper_cls))
```
myleott, line 230: I'm a bit confused by this, where is `type` defined? Also, tradeoffs of this vs. `isinstance(child, ConfigAutoWrap.wrapper_cls)`?

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -130,68 +169,85 @@ class ConfigAutoWrap:
...
 
+        # We count all params, assuming none of them is already wrapped.
         num_params = sum([p.numel() for p in module.parameters()])
 
-        if num_params >= min_num_params:
+        assert should_wrap is not None
+        if should_wrap(True, module, num_params):
```
myleott, line 236: nit: perhaps add kwargs to make this easier to read? `should_wrap(recurse=True, module=module, unwrapped_params=num_params)`

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -130,68 +169,85 @@ class ConfigAutoWrap:
...
+                    module=child, should_wrap=should_wrap, **kwargs
                 )
                 setattr(module, name, wrapped_child)
                 # Keep track of how many parameters have been wrapped
                 total_wrapped_params += num_wrapped_params
             # decide if we need to wrap the current module,
             # since the left over parameters exceed the number of params to wrap
             remainder = num_params - total_wrapped_params
-            if remainder >= min_num_params and not isinstance(module, tuple(FSDP_MODULE_EXCLUDE_WRAP)):
+            if should_wrap(False, module, remainder):
```
myleott, line 249: same, kwargs would make this easier to read

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -34,26 +88,25 @@ def enable_wrap(module_blocklist: Optional[List] = None, **wrapper_kwargs: Any)
...
     Args:
-        module_blocklist: List of additional Module Classes to not wrap when
-            using :func:`auto_wrap`. This is useful to exclude unsupported
-            modules when wrapping recursively.
-        **wrapper_kwargs: Configuration settings that will be passed to all ``wrap``
+        should_wrap (Callable, Optional):
+            Custom function to control how to do :func:`auto_wrap`. This is
+            useful to exclude unsupported modules or wrap based on sizes when
+            wrapping recursively.
+            (default: :func:`default_should_wrap_policy`)
```
myleott, line 100: it's not totally clear from this description if this applies to `wrap` layers or just `auto_wrap`. For example:
```python
with enable_wrap(should_wrap=lambda _, module, _: False):
    l1 = wrap(nn.Linear(1, 1))
```
In this case the Linear would be wrapped, correct?

Assuming yes, then perhaps update this description to something like:
```
Custom function to control how to do :func:`auto_wrap`. This is
useful to exclude unsupported modules or wrap based on sizes when
wrapping recursively. Note: modules annotated with :func:`wrap`
ignore this policy and will always be wrapped
(default: :func:`default_should_wrap_policy`)
```

Given the above, I'm also wondering if it'd be clear to rename `should_wrap` to `auto_wrap_policy`. What do you think?

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -4,23 +4,77 @@
...
-from fairscale.nn.data_parallel.fully_sharded_data_parallel import FullyShardedDataParallel
-from fairscale.nn.misc import checkpoint_wrapper
 
-# Modules that don't wrap.
-FSDP_MODULE_EXCLUDE_WRAP = {nn.ModuleList, nn.ModuleDict}
-# Modules that we don't recurse down to their children.
-FSDP_MODULE_BLOCKLIST = {nn.MultiheadAttention}
+def default_should_wrap_policy(
+    recurse: bool,
+    module: nn.Module,
```
myleott, line 14: minor nit: what do you think about swapping the order of these two? In some sense, `module` is the main argument to this function and `recurse` is secondary

myleott (commented):
Nice!

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: this is to make mypy happy. casting from callable to type, which is builtin type of python. :-)

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: good idea.

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: done

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: done

fairscale/nn/wrap/auto_wrap.py
min-xu-ai: good idea. done.

min-xu-ai:
Thanks for review again, Myle! Even better now. Take another look?

myleott (approved):
Thanks for this refactor, it came out really nice!

SeanNaren:
Great job @min-xu-ai!!

min-xu-ai:
@myleott this seems like an unrelated to this change?

https://app.circleci.com/pipelines/github/facebookresearch/fairscale/2064/workflows/3de30f4d-5c84-4b8a-bb51-8f8e6dc231db/jobs/12695/tests#failed-test-0

It is fsdp & moe fail to compare with DDP.

myleott:
Yes, it seems to be failing intermittently even on master, which is concerning. Probably something wrong with stream synchronization, I'll look into it

PR #473, opened 2021-03-04
[hotfix]: fix a bug in CI command

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,7 @@ install_dep_151: &install_dep_151
       command: |
         sudo apt-get install -y libopenmpi-dev
         # check if we have restored venv cache (/home/circleci/venv) correctly, if so, just skip
+        if [ -f /home/circleci/venv/check_version.py ]; then python /home/circleci/venv/check_version.py torch ne 1.5 && exit 1; fi
         # start installing
         pip install --progress-bar off torch==1.5.1+cu101 torchvision==0.6.1+cu101 -f https://download.pytorch.org/whl/torch_stable.html
         pip install --progress-bar off -r requirements-test.txt
@@ -88,6 +89,7 @@ install_dep_160: &install_dep_160
         python -c 'import torch; print("Torch version:", torch.__version__)'
         python -c 'import torch; assert torch.__version__.split(".")[:2] == ["1", "6"], "wrong torch version"'
         python -m torch.utils.collect_env
+        wget -O /home/circleci/venv/check_version.py https://raw.githubusercontent.com/min-xu-ai/check_verion/main/check_version.py
 
 install_dep_171: &install_dep_171
   - run:
@@ -103,6 +105,7 @@ install_dep_171: &install_dep_171
         python -c 'import torch; print("Torch version:", torch.__version__)'
         python -c 'import torch; assert torch.__version__.split(".")[:2] == ["1", "7"], "wrong torch version"'
         python -m torch.utils.collect_env
+        wget -O /home/circleci/venv/check_version.py https://raw.githubusercontent.com/min-xu-ai/check_verion/main/check_version.py
 
 install_dep_171_cu110: &install_dep_171_cu110
   - run:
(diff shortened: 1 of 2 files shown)


min-xu-ai:
A incoming PR uncovered a bug in the CI config: when import torch fails, the command should return OK so that we go ahead with pip installation. This version should fixes it.

cc @vfdev-5

msbaines:
> A incoming PR uncovered a bug in the CI config: when import torch fails, the command should return OK so that we go ahead with pip installation. This version should fixes it.
> 
> cc @vfdev-5

Can you elaborate. Why should we continue if import torch fails? Isn't that catastrophic.

min-xu-ai:
> Can you elaborate. Why should we continue if import torch fails? Isn't that catastrophic.

Still testing. Converted to draft.

The check here is that we want to continue to do the pip install of torch if it doesn't exist.

The old code is that we do always do pip install even if restoring from the cache was successful. That can fail from time to time due to networking issues.

I first put in a check for the right torch version to skip the pip install but that check was buggy. It skipped install when torch is not there.

min-xu-ai:
This is ready. The failure is unrelated. I tested on CI and verified that we skip pip install when cache is correctly restored and with right torch version.

.circleci/config.yml
```diff
@@ -65,22 +65,23 @@ install_dep_151: &install_dep_151
...
+        wget -O /home/circleci/venv/check_version.py https://raw.githubusercontent.com/min-xu-ai/check_verion/main/check_version.py
 
 install_dep_160: &install_dep_160
   - run:
       name: Install Dependencies with torch 1.6.0
       command: |
         sudo apt-get install -y libopenmpi-dev
         # check if we have restored venv cache (/home/circleci/venv) correctly, if so, just skip
-        python -c 'import torch; assert torch.__version__.split(".")[:2] != ["1", "6"]' || exit 0
+        if [ -f /home/circleci/venv/check_version.py ]; then python /home/circleci/venv/check_version.py torch eq 1.6 && exit 0; fi
```
blefaudeux, line 84: alrighty, makes sense

.circleci/config.yml
```diff
@@ -146,8 +150,8 @@ install_dep_180: &install_dep_180
         pip install --progress-bar off  git+https://github.com/min-xu-ai/torch_pg.git@c723ab4#egg=torch-pg
         python -c 'import torch; print("Torch version:", torch.__version__)'
         python -c 'import torch; assert torch.__version__.split(".")[:2] == ["1", "8"], "wrong torch version"'
-        pip list|grep torch
         python -m torch.utils.collect_env
+        wget -O /home/circleci/venv/check_version.py https://raw.githubusercontent.com/min-xu-ai/check_verion/main/check_version.py
```
blefaudeux, line 154: typo, no ? 'check_verion'

.circleci/config.yml
```diff
@@ -432,14 +436,14 @@ jobs:
       # Cache the venv directory that contains dependencies
       - restore_cache:
           keys:
-            - cache-key-gpu-151-{{ checksum "setup.py"}}-{{ checksum "requirements-test.txt"}}
+            - cache-key-gpu-151-1-{{ checksum "setup.py"}}-{{ checksum "requirements-test.txt"}}
```
blefaudeux, line 439: given the setup.py change, this is not needed, is it ?

blefaudeux:
LGTM except for a couple of questions, mind checking Min ? Thanks in any case, I realize that this is a bit arduous but needed

.circleci/config.yml
min-xu-ai: actually, it is not. It is how I made my repo. (by accident)

.circleci/config.yml
min-xu-ai: right. I can remove it

min-xu-ai:
> LGTM except for a couple of questions, mind checking Min ? Thanks in any case, I realize that this is a bit arduous but needed

Checked. Good call on the "-1-" part. Will remove.

blefaudeux (commented):
LGTM with the additional read, thanks again Min !

PR #474, opened 2021-03-05
[perf][minor] cache the rank lookups, small shardedddp perf fix

blefaudeux:
# Before submitting

## What does this PR do?
- small cleanup of the logic on the shardedddp side since the introduction of the bucket flush (no need to count the grads, ..). Small perf improvements thanks to that, most (but not all) of what was lost with this feature, but I think that it's worth it, just safer
- removing global rank lookups
- should fix the intermittent perf errors for OSS
 
## PR review

## Did you have fun?


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -153,10 +156,6 @@ def __init__(
         self._trainable_param_to_rank: Dict[torch.Tensor, int] = {}
         self._reference_trainable_mask = list(map(_trainable, self._all_params))
 
-        # - keep track of the grads which have already been reduced
-        self._reduced_grads = 0
```
blefaudeux, line 157: no need for these anymore, it was there prior to using the Variable queue method

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -425,27 +422,21 @@ def cleanup() -> None:
                             param.grad.data = param.grad.data.to(dtype=param.dtype)
 
                     # Async reduce for this buffer, log the future
-                    dst_global_rank = OSS.get_global_rank(self.process_group, dst_rank)
```
blefaudeux, line 428: cache these lookups, probably nothing perf-wise but not really needed

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -425,27 +422,21 @@ def cleanup() -> None:
...
-                    self._reduced_grads += 1
 
-                    # Opportunistically try to empty the queue
+                    # Opportunistically try to empty the queue, free memory
                     self._try_consume_work_handle()
 
-                    # If all the reduce operations have been called,
-                    # make sure that all the asynchronous calls have concluded before moving on
-                    # and execute the delayed actions (release gradients, unroll the buckets)
-                    if self._reduced_grads == self._reduced_grads_max:
```
blefaudeux, line 446: this was sometimes blocking, but not needed given the Variable queue callback, and makes the code for this callback from complicated to grok

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -480,17 +471,10 @@ def reduce(*_: Any) -> None:
                                 callback=None,
                             )
                         )
-                        self._reduced_grads += 1
 
                     # Opportunistically try to empty the queue
                     self._try_consume_work_handle()
 
-                    # If all the reduce operations have been called,
```
blefaudeux, line 488: same, not needed anymore, less is more

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -640,3 +617,5 @@ def _flush_buckets(self) -> None:
 
             if last_handle is not None:
                 last_handle.wait()
+
+        self._consume_work_handles()
```
blefaudeux, line 621: this is called once per BW pass, consume all the futures and we're good to go

tests/ci_test_list_3.txt
```diff
@@ -1,8 +1,8 @@
 tests/nn/data_parallel/test_fsdp_uneven.py
 tests/nn/data_parallel/test_fsdp_grad_scaler.py
 tests/nn/data_parallel/test_fsdp_summon_full_params.py
-tests/nn/data_parallel/test_features_sharded_ddp.py
-tests/nn/data_parallel/test_pytorch_parity_sharded_ddp.py
+tests/nn/data_parallel/test_sharded_ddp_features.py
```
blefaudeux, line 4: renamed for consistency, and just more logical I believe (name & sort)

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -107,7 +107,7 @@ def __init__(
         self.enable_broadcast_buffers = broadcast_buffers
         self.auto_refresh_trainable = auto_refresh_trainable
         self.reduce_fp16 = reduce_fp16
-        if reduce_buffer_size > 0:
+        if reduce_buffer_size > 0 and reduce_fp16:
```
blefaudeux, line 110: does not change the logic, just removes a warning which could be useless

fairscale/nn/data_parallel/sharded_ddp.py
min-xu-ai: we discovered that variable queue method and activation_checkpointing have some bad interactions. Basically, if you queue the finally callback in a checkpointed bwd, the callback can come before the overall bwd finishes. You may want to verify that's not affecting SDP.

min-xu-ai (approved):
Looks like good cleanup!

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: ohh, thanks for the headsup. It's already the case before this PR anyway, so I'm guessing that this can go through but to keep in mind, not sure how to grab that case ? (practically you can disable the buckets in that case for SDP, so there's at least one solution)

PR #483, opened 2021-03-05
[fix] OSS dict: simplified logic, bugfix for interleaved param groups 

blefaudeux:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -85,7 +85,6 @@ class OSS(Optimizer):
         self._per_device_params: Dict[torch.device, List[List[Parameter]]] = OrderedDict()  # device, rank, params
         self._param_rank: Dict[torch.Tensor, int] = {}
         self._partition_parameters: List[List[dict]] = []
-        self._index_to_param: Dict[int, torch.Tensor] = {}
         self._param_to_index: Dict[int, int] = {}
         self._local_params: Optional[List[torch.Tensor]] = None
 
@@ -159,15 +158,6 @@ class OSS(Optimizer):
 
         # Make sure that the iterator is not consumed, only expose a copy
         return self._local_params
-
-    @property
-    def index_to_param(self) -> Dict[int, torch.Tensor]:
-        """ Hash table in between parameter indices in the global optimizer scheme, and the actual params
-        """
-        if len(self._index_to_param) == 0:
-            self._index_to_param = {i: p for i, p in enumerate(chain(*(g["params"] for g in self.param_groups)))}
-
-        return self._index_to_param
 
     @property
     def param_to_index(self) -> Dict[int, int]:
@@ -376,7 +366,7 @@ class OSS(Optimizer):
                         global_id = self.param_to_index[local_index_to_param_id[local_param_index]]
                         state_dict["state"][global_id] = s["state"][local_param_index]
 
-        # Make sure that the parameters are sorted in the state, as expected
+        # Make sure that the parameters are sorted in the state, as expected for a pytorch dict
         state_dict["state"] = dict(sorted(state_dict["state"].items()))
 
         return state_dict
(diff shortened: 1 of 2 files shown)


blefaudeux:
Pytorch1.5 needs fixing, as always..

fairscale/optim/oss.py
```diff
@@ -160,15 +159,6 @@ def local_params(self) -> List[torch.Tensor]:
         # Make sure that the iterator is not consumed, only expose a copy
         return self._local_params
 
-    @property
-    def index_to_param(self) -> Dict[int, torch.Tensor]:
```
blefaudeux, line 164: this was an optimisation for very big models, which I think was a mistake: this increased the code footprint, and really, really big models will not be trained with OSS anyway. On top of that, this is not needed with this PR except for torch 1.5, which will be deprecated soonish

tests/optim/test_oss.py
```diff
@@ -811,9 +817,11 @@ def check_optimizer_equivalence(optimizer: Type[torch.optim.Optimizer], change_t
 
         # Define a model to be trained by OSS
         oss_module = torch.nn.Sequential(trunk, head)
+
+        # Make sure that the param groups are interleaved, to catch an ordering bug in the state dict
         oss_trainable_params = [
-            {"params": trunk.parameters(), "lr": 1e-5},
```
blefaudeux, line 815: this unit test let a "subtle" bug pass: it was testing param groups which are sequential with respect to the model parameters, turns out that there was a bug if this was not the case.

fairscale/optim/oss.py
```diff
@@ -389,17 +379,34 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
                 from a call to :meth:`state_dict`
         """
 
-        # NOTE: PyTorch 1.5 does not index linearly but with the id(params) at saving time
-        # we work around that here by using the fact that the params are ordered as in the param_groups
-        pytorch15_index_redirect = {k: i for i, k in enumerate(state_dict["state"].keys())}
```
blefaudeux, line 394: this was the arguably a little subtle bug: this implied that the loaded state dict and the param groups were sorted in the same way, which is not always the case actually (happened to be the case for the unit test in place..)

fairscale/optim/oss.py
```diff
@@ -389,17 +379,34 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+        # Update the state, trusting the ordering in param_groups
+        # Apart from the removal of states not owned by this rank, the pytorch logic is kept
+        # (See torch.optim.optimizer)
+        id_map = {
```
blefaudeux, line 385: this is exactly the same logic as pytorch (as commented), which I think is a safer and sane-er (if this word exist ?) way to go

fairscale/optim/oss.py
```diff
@@ -389,17 +379,34 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+            old_id: p
+            for old_id, p in zip(
+                chain.from_iterable((g["params"] for g in state_dict["param_groups"])),
+                chain.from_iterable((g["params"] for g in self.param_groups)),
+            )
+        }
+
+        # FIXME: pytorch1.5 compatibility, to be removed when 1.5 support ends
+        _param_list = list(chain.from_iterable((g["params"] for g in self.param_groups)))
```
blefaudeux, line 394: the assumption (same as pytorch) is not that param_groups are ordered in the same way in between checkpoint and current state. Previous assumption was that checkpointed *state* and current param_groups are ordered in the same way, which turned out to be false in some cases. cc @vedanuj

blefaudeux:
ping reviewers

min-xu-ai (approved):
lgtm

Issue #498, opened 2021-03-08
[SPD] flaky unit test


blefaudeux:
see https://app.circleci.com/pipelines/github/facebookresearch/fairscale/2127/workflows/477e8512-fca3-4d54-a74a-3143c5d31b47/jobs/13546

and 
https://app.circleci.com/pipelines/github/facebookresearch/fairscale/2126/workflows/e8cc7dd1-4de8-4644-a8a1-e555ba67b414/jobs/13531

both checking parity while using Gloo

PR #499, opened 2021-03-08
[fix] flaky SDP tests with Gloo, checking all handles

blefaudeux:
# Before submitting

## What does this PR do?
Seems to fix https://github.com/facebookresearch/fairscale/issues/498 (see [CI](https://app.circleci.com/pipelines/github/facebookresearch/fairscale?branch=gloo_check_all_handles), I kicked the jobs a couple of times in a row, all green). This is a change which is only needed for Gloo actually, but testing with several backends has other advantages so my take is that it's worth the very small perf hit

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -319,7 +319,7 @@ class ShardedDataParallel(nn.Module):
                 dist.broadcast(buffer.data, self.reference_global_rank, self.process_group, async_op=True)
             )
 
-        if blocking:
+        if blocking and work_handles:
             if self.backend != dist.Backend.NCCL:
                 _ = list(filter(lambda x: x.wait(), work_handles))
             else:
@@ -518,7 +518,7 @@ class ShardedDataParallel(nn.Module):
         # gloo does not guarantee inlining like NCCL, wait for all requests
         if self.backend != dist.Backend.NCCL:
             _ = list(filter(lambda x: x.wait(), work_handles))
-        else:
+        elif work_handles:
             work_handles[-1].wait()
 
     def _passing_sync_batchnorm_handle(self, module: nn.Module) -> None:
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -120,6 +120,7 @@ class ShardedDataParallel(nn.Module):
 
         # Communication related attributes
         self.process_group = process_group if process_group is not None else dist.group.WORLD
+        self.backend = dist.get_backend(self.process_group)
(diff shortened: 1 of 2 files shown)


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -505,16 +504,15 @@ def _sync_params_and_buffers(self) -> None:
...
-                t, src=self.reference_global_rank, group=self.process_group, async_op=True
+            work_handles.append(
+                dist.broadcast(t, src=self.reference_global_rank, group=self.process_group, async_op=True)
             )
 
-        # Only wait for the last handle, they're inlined in the same CUDA stream
-        if last_work_handle:
-            last_work_handle.wait()
+        # gloo does not guarantee inlining like NCCL, wait for all requests
+        _ = list(filter(lambda x: x.wait(), work_handles))
```
blefaudeux, line 515: the crux of the diff: Gloo is actually not guaranteeing that the network calls will be sequential (or so I've been told), which means that it's safer to actually check for all handles (not needed for NCCL, but I figured that a backend check here would really be ugly). same below

blefaudeux:
(the broken test is not ShardedDDP, looks like a flaky MoE that @myleott fixed earlier)

min-xu-ai (approved):

fairscale/nn/data_parallel/sharded_ddp.py
anj-s: is there a perf difference for NCCL? I vaguely remember this change helped with perf but I could have misremembered the PR.

blefaudeux: yes there's a small perf hit, you can see all the checks on the trace which take a non-negligible amount of time (like, %), but I think that having something super robust is still worth it for now, until we have a real perf-constrained usecase ? right now the broadcast part is slowing things down so much that this is o()

blefaudeux: (done)

fairscale/optim/oss.py
```diff
@@ -540,19 +540,19 @@ def _sync_param_groups(source: List[Dict[Any, Any]], destination: List[Dict[Any,
...
-                    global_src_rank = self.get_global_rank(self.group, src_rank)
-                    last_work_handle = dist.broadcast(
-                        tensor=bucket, src=self._local_to_global_rank[src_rank], group=self.group, async_op=True
+                    work_handles.append(
+                        dist.broadcast(
+                            tensor=bucket, src=self._local_to_global_rank[src_rank], group=self.group, async_op=True
+                        )
                     )
 
         # Only check on the last handle, they're all inlined on the same CUDA stream
```
anj-s, line 554: Do we need to update this as well?

PR #500, opened 2021-03-09
[feat] Make OSS state available on all ranks

blefaudeux:
# Before submitting

## What does this PR do?
Makes it possible to fix https://github.com/PyTorchLightning/pytorch-lightning/issues/6387

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -318,7 +318,9 @@ class OSS(Optimizer):
             if rank == self.rank:
                 if should_collect_state:
                     logging.debug("Saving self state")
-                    self._all_states.append(self.optim.state_dict())
+                    self._all_states.append(
+                        recursive_copy_to_device(self.optim.state_dict(), non_blocking=True, device=torch.device("cpu"))
+                    )
 
                 # Sync with other replicas
                 state_to_share = (
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -449,7 +449,6 @@ class OSS(Optimizer):
                     self.optim.state[param] = recursive_copy_to_device(value, non_blocking=True, device=param.device)
             else:
                 # Not a param, copied as-is (backward compatibility or exotic optimizers)
-                print(key, "not in idmap")
                 param = _param_list[key]
                 self.optim.state[param] = recursive_copy_to_device(value, non_blocking=True, device=param.device)
 
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -297,19 +297,53 @@ class OSS(Optimizer):
     def consolidate_state_dict(self, recipient_rank: int = 0) -> None:
         """Update the consolidated state_dict list, one per rank.
 
+        Arguments:
+            recipient_rank (int): on which rank to materialize the full state dict.
+            -1 is a special value, which means that all ranks should have the state
+
         .. warning: This needs to be called on all replicas"""
 
         # Sync lr and other attributes in case its been updated
         OSS._sync_param_groups(self.param_groups, self.optim.param_groups)
 
-        if self.rank == recipient_rank:
-            # Pull the sharded state from all the other replicas
-            # Store all the states in order, rank by rank
-            logging.debug("Pulling the sharded optimizer state from all replicas")
-            self._all_states = self._collect_sharded_states()
-        else:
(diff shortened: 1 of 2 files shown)


blefaudeux:
cc @SeanNaren @ananthsub

fairscale/optim/oss.py
```diff
@@ -422,69 +465,6 @@ def refresh_trainable(self) -> None:
 
         self._setup_flat_buffers()
 
-    def _broadcast_state_dict(self) -> None:
```
blefaudeux, line 425: consolidate the two following functions into one, since with the "all ranks" behavior the split did not make a lot of sense anymore

fairscale/optim/oss.py
```diff
@@ -334,29 +368,38 @@ def local_state_dict(self) -> dict:
         """
         return self.optim.state_dict()
 
-    def state_dict(self) -> Dict[str, Any]:
+    def state_dict(self, all_ranks: bool = False) -> Dict[str, Any]:
```
blefaudeux, line 371: why all this: make it possible to get the state on all ranks. In that case a prealable "consolidate_state_dict" is not needed anymore

fairscale/optim/oss.py
```diff
@@ -334,29 +368,38 @@ def local_state_dict(self) -> dict:
...
+            if `all_ranks` was not set to `True`. In that case, the state may also not be up to date,
+            depending on when `consolidate_state_dict` was last called.
         """
 
-        if len(self._all_states) == 0:
+        if not all_ranks and len(self._all_states) == 0:
             raise RuntimeError(
                 "Optimizer state has not been consolidated on this rank. \
                 Please call `consolidate_state_dict()` on all ranks beforehand if you meant to save the global state"
             )
```
ananthsub, line 397: Would it be possible to change the runtime error to a warning ? 

The error case for us is if a user is using the sharded optimizer outside of sharded DDP. If we're not using sharded DDP, we assume the optimizer is a PyTorch optimizer. We have a simple callback that does this:
```
def on_save_checkpoint(self, trainer: Trainer, _pl_module: LightningModule) -> None:
        for optimizer in trainer.optimizers:
            if isinstance(optimizer, OSS):
                    optimizer.consolidate_state_dict()
```
After this runs, we fetch the state dict on all ranks when creating the checkpoint dict. This runs into this error on non-zero ranks. If we instead change this to become a warning and returns even an empty dictionary, then that will resolve the issue. But it depends on what the expectations are for `state_dict` and whether a warning + empty dict is a hack

blefaudeux: I would prefer not to, because it's not doing what the contract says it will do, it cannot return the state_dict in that case, silently failing (nobody reads warnings) feels like something which will bite the users in the future. 

I don't understand why you would fetch an empty dict in that case, since you do something differently in between the ranks anyway why not skip the state_dict() call ?

blefaudeux: example failure case: a user kicks a very expensive training at scale, and thinks that the checkpointing is fine. For some reason he/she is saving an empty dict from a different rank than the one which got the state, and all the checkpoints are not useable, which he/she realizes long after the training is done. I would really much prefer not to open up for a failure case like that

SeanNaren: Agree with @blefaudeux on this one, we've actually seen the example that Ben brought up when someone was using our pipe implementation in PL, since there is no forcing of save logic there.

As Ben highlighted, what's the use case of saving on all ranks?

blefaudeux:
looks like the comms behaviour was different with 1.7, looking into that, should not be structural.

fairscale/optim/oss.py
blefaudeux: @ananthsub would that work for you ?

anj-s (approved):

fairscale/optim/oss.py
blefaudeux: ok, I just re-read 

> "The error case for us is if a user is using the sharded optimizer outside of sharded DDP. If we're not using sharded DDP, we assume the optimizer is a PyTorch optimizer.". 


The issue there is that this has nothing to do with ShardedDDP, OSS is the one sharding the optimizer state, so **with or without ShardedDDP the optimizer state is sharded**, and the vanilla Pytorch API does not work (unless we move to RPC, may be the case in the future but not really possible right now)

PR #502, opened 2021-03-09
[perf] Further improve performance for FSDP.no_sync

myleott:
The previous fix was still slow for nested FSDP instances with reshard_after_forward=False

### DIFF
--- a/tests/nn/data_parallel/test_fsdp_no_sync.py
+++ b/tests/nn/data_parallel/test_fsdp_no_sync.py
@@ -4,15 +4,17 @@
 # LICENSE file in the root directory of this source tree.
 
 import functools
+import itertools
 import unittest
 from unittest.mock import patch
 
+from parameterized import parameterized
 import torch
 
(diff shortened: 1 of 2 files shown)


tests/nn/data_parallel/test_fsdp_no_sync.py
```diff
@@ -94,37 +96,82 @@ def _test_no_sync(self, model, batch_dim):
...
+            model = self.get_wrapped_model(group, config=config)
+
+        num_fsdp = 0
+        for child in model.modules():  # includes self
+            if isinstance(child, FullyShardedDataParallel) and len(child.params) > 0:
+                num_fsdp += 1
+
+        if config.get("reshard_after_forward", True):
+            # inside no_sync:
+            #   num_fsdp all-gathers in the forward
```
zhengwy888, line 135: it's a bit difficult to wrap my head around, why inside no_sync the forward all gather is one more than outside no_sync's forward?

PR #512, opened 2021-03-11
[OSS] Adding a hard sync stream barrier before broadcast

blefaudeux:
# Before submitting

## What does this PR do?
It's possibly just me, but pytorch grabbing a non-manageable cuda stream for all NCCL related things is a good way to trip over. Make sure that the compute streams are done before sending the buckets, ideally everything could be lined up, but this is as good as it gets for now

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -549,7 +549,7 @@ class OSS(Optimizer):
 
         # if NCCL broadcasts will be done in an independent stream
         # make sure that prior compute work is complete
-        if self.backend == dist.Backend.NCCL:
+        if torch.device("cuda").type == self._default_device.type:
             for device in self.per_device_params.keys():
                 torch.cuda.synchronize(device=device)
 
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -94,6 +94,7 @@ class OSS(Optimizer):
 
         self.group = group if group is not None else dist.group.WORLD
         self.world_size = dist.get_world_size(self.group)
+        self.backend = dist.get_backend(self.group)
         self.rank = dist.get_rank(self.group)
         self.global_rank = self.get_global_rank(self.group, self.rank)
         self._local_to_global_rank = [self.get_global_rank(self.group, i) for i in range(self.world_size)]
@@ -546,6 +547,12 @@ class OSS(Optimizer):
     def _broadcast_params(self) -> None:
         """Helper function to broadcast all the parameters from a given device"""
 
+        # if NCCL broadcasts will be done in an independent stream
+        # make sure that prior compute work is complete
+        if self.backend == dist.Backend.NCCL:
+            for device in self.per_device_params.keys():
+                torch.cuda.synchronize(device=device)
+
         work_handles = []  # Work handles are consumed within this scope, no callback
 
         for device in self.buckets.keys():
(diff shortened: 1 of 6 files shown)


fairscale/optim/oss.py
```diff
@@ -546,6 +547,12 @@ def _sync_param_groups(source: List[Dict[Any, Any]], destination: List[Dict[Any,
     def _broadcast_params(self) -> None:
         """Helper function to broadcast all the parameters from a given device"""
 
+        # if NCCL broadcasts will be done in an independent stream
+        # make sure that prior compute work is complete
+        if self.backend == dist.Backend.NCCL:
+            for device in self.per_device_params.keys():
+                torch.cuda.synchronize(device=device)
```
min-xu-ai, line 554: this syncs the specified GPU and CPU, right? Does this actually sync streams? Is there a way to have a unit test and trigger this and ensure the fix is right?

blefaudeux: it should just sync all the streams, which does not imply necessarily a cpu/gpu sync I think. It's a good question, I was also wondering about a unit test, I'll try to make up a fake optimizer which waits on GPU for a long time

blefaudeux:
adding a unit test with a fake optimizer which holds onto the compute for a long time

fairscale/utils/testing.py
```diff
@@ -548,3 +548,14 @@ def rank(self) -> int:
...
         return self._size
+
+
+class SGDWithPausingCompute(torch.optim.SGD):
+    def step(self, closure: Optional[Any] = None) -> Any:
+        loss = super().step(closure=closure)
+
+        # Add a long cuda wait on the default compute stream
+        # This is used to make sure that OSS and ShardedDDP enforce a proper stream synchronization
+        torch.cuda._sleep(1000000)
```
blefaudeux, line 559: so this does not work, because no real computation depends on that, so the gpu is waiting on one stream while broadcasting no the other. I need to find something a little better than that

blefaudeux:
See tests on https://app.circleci.com/pipelines/github/facebookresearch/fairscale?branch=oss_stream_sync, a previous revision was not doing the sync for Gloo and the unit test modification (slow optim + rank specific) caught that

fairscale/utils/testing.py
```diff
@@ -548,3 +548,27 @@ def rank(self) -> int:
...
+
+        super().__init__(*args, **kwargs)
+
+    def step(self, closure: Optional[Any] = None) -> Any:
+        loss = super().step(closure=closure)
+
+        # This is used to make sure that OSS and ShardedDDP enforce a proper stream synchronization
+        # - Add a long cuda wait on a compute stream, non blocking from the CPU perspective
+        with torch.cuda.stream(torch.cuda.Stream()):
+            torch.cuda._sleep(100000000)
```
blefaudeux, line 566: the gist of this is to add a slow rank-specific workload which will modify the params. If the broadcast does not wait for this to be done, then the params will eventually differ in between the ranks

min-xu-ai (approved):
super great

fairscale/optim/oss.py
blefaudeux: so it seems that it's somehow not needed for NCCL actually, because pytorch is doing the same (but it's not doing this for gloo..)

PR #513, opened 2021-03-11
[fix] FSDP: multi-pass autograd graph and mixed precision

min-xu-ai:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -886,7 +886,7 @@ class FullyShardedDataParallel(nn.Module):
         Due to multiple-pass forward, this function can be called on
         the same parameter multiple times in a single forward pass. If we register
         the hook multiple time, we end up getting called multiple times. We
-        could try to get a new hook every time and delete the previou one
+        could try to get a new hook every time and delete the previous one
         registered. However, due to *unknown reason* (I have debugged it for
         a long time!), in mixed precision mode, we get two different ``grad_acc``
         objects below during different calls of this function (in the same
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -34,6 +34,13 @@ class TrainingState(Enum):
     Simple enum to indicate what state FSDP is in. Used for asserting
     to make sure APIs are called in the correct state.
 
+    ..note::
+
+        BACKWARD_PRE and BACKWARD_POST states are used to ensure we
+        receives backward hooks in the correct order. It is used to catch
+        unexpected order of hooks being called (likely due to our
+        hook registration logic or autograd engine logic changes).
+
     TODO (Min): It would be nice to capture the stepping state as well.
         Maybe we can use the model.zero_grad() call, but not sure if it
         is called if optim.zero_grad() is used instead.
@@ -45,7 +52,8 @@ class TrainingState(Enum):
 
     IDLE = auto()
     FORWARD = auto()
-    BACKWARD = auto()
+    BACKWARD_PRE = auto()
+    BACKWARD_POST = auto()
     SUMMON_FULL_PARAMS = auto()
 
 
@@ -581,7 +589,7 @@ class FullyShardedDataParallel(nn.Module):
             # Set the state so that we assert when trying to go into
             # forward/backward.
             self.training_state = TrainingState.SUMMON_FULL_PARAMS
-            full_tensors = self._rebuild_full_params(full_precision=True)
+            full_tensors = self._rebuild_full_params(force_full_precision=True)
             assert full_tensors is not None
             with contextlib.ExitStack() as stack:
                 if self.flatten_parameters and self.module.is_flattened:
@@ -835,7 +843,18 @@ class FullyShardedDataParallel(nn.Module):
             pre_backward_hook_has_run[0] = True
 
             # Start of a backward pass.
-            self.training_state = TrainingState.BACKWARD
+            if self.rank == 0 and self.world_size > 1:
+                print(
+                    "XXX start bwd",
+                    self.training_state,
+                    self._is_root,
+                    self.world_size,
+                    len(unused),
+                    unused[0].shape,
+                    unused[0].sum().item(),
+                )
+            self.assert_state([TrainingState.IDLE, TrainingState.BACKWARD_PRE])
+            self.training_state = TrainingState.BACKWARD_PRE
 
             # All-gather full parameters.
             if self.reshard_after_forward:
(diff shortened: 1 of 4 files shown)


min-xu-ai:
Not ready. No need to review

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -857,7 +866,40 @@ def _register_hook(t: torch.Tensor) -> torch.Tensor:
...
+        2. If it fires more than once, we end up incorrectly shard the grad
+        multiple times. (could lead to dimension too small)
+        3. If it fires once but too early or doesn't fire, we leave gradients
+        unsharded. (could lead to dimension too large)
+
+        Due to multiple-pass forward, this function can be called on
+        the same parameter multiple times in a single forward pass. If we register
+        the hook multiple time, we end up getting called multiple times. We
+        could try to get a new hook every time and delete the previou one
+        registered. However, due to *unknown reason* (I have debugged it for
```
min-xu-ai, line 890: This is not very satisfying since I didn't find out what's different with mixed precision. I did bunch of tests with and without FSDP, I couldn't reproduce the same behavior (different grad_acc objects) outside of FSDP. Within FSDP, nothing related to the param or the data within the param seems to trigger this. If someone else can figure this out, definitely help!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1014,35 +1074,47 @@ def _wait_for_post_backward(self) -> None:
...
+                _remove_shard_bwd_hook(m)
+                if m._has_params:
+                    m.assert_state(TrainingState.BACKWARD_POST)
+                else:
+                    m.assert_state(TrainingState.BACKWARD_PRE)
                 m.training_state = TrainingState.IDLE
 
     @torch.no_grad()
-    def _rebuild_full_params(self, full_precision: bool = False) -> Optional[List[Tuple[torch.Tensor, bool]]]:
+    def _rebuild_full_params(self, force_full_precision: bool = False) -> Optional[List[Tuple[torch.Tensor, bool]]]:
```
min-xu-ai, line 1085: renamed this parameter, no logic change.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1055,35 +1127,36 @@ def update_p_data(custom_output_tensor: Optional[torch.Tensor] = None) -> None:
...
                 else:
-                    # If self.cpu_offload and full_precision, we need to cast
+                    # If self.cpu_offload and force_full_precision, we need to cast
                     # the FP32 CPU param to CUDA for the all-gather.
                     p_data = p.data.to(p._full_param_padded.device)
 
                     p_size = p._full_param_padded.size()
                     assert p_size.numel() % self.world_size == 0
-                    if not self.mixed_precision or not full_precision:
+                    if self.mixed_precision and force_full_precision:
```
min-xu-ai, line 1151: switched the if condition's order, easier to read, no logic change.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1180,11 +1253,19 @@ def _free_fp16_param_shard(self, params: Optional[List[Parameter]] = None) -> No
                 p._fp16_shard.record_stream(current_stream)
                 free_storage_(p._fp16_shard)
 
-    def assert_state(self, state: TrainingState) -> None:
+    def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None:
```
min-xu-ai, line 1256: made this assert an error checking since it seems to be useful even when people run python -O.

sshleifer (approved):
Interesting edge case!
From my reading, it seems we only add _shard_bwd_hook once, do we still need to `_remove_shard_bwd_hook`?

min-xu-ai:
> Interesting edge case!
> From my reading, it seems we only add _shard_bwd_hook once, do we still need to `_remove_shard_bwd_hook`?

Thanks for the review!

We still need _remove_shard_bwd_hook since it appears that we need to add the hook once per-forward (again, for some modules, they get called multiple times per-forward). If we don't do the hook per-forward, the hook may not get called. I think all this has something todo with how torch generate that GradAccumulate object when the forward is executing with torch.grad().

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -857,7 +866,40 @@ def _register_hook(t: torch.Tensor) -> torch.Tensor:
...
+        are accumulated for a param.
+        2. If it fires more than once, we end up incorrectly shard the grad
+        multiple times. (could lead to dimension too small)
+        3. If it fires once but too early or doesn't fire, we leave gradients
+        unsharded. (could lead to dimension too large)
+
+        Due to multiple-pass forward, this function can be called on
+        the same parameter multiple times in a single forward pass. If we register
+        the hook multiple time, we end up getting called multiple times. We
+        could try to get a new hook every time and delete the previou one
```
myleott, line 889: nit: typo
```suggestion
        could try to get a new hook every time and delete the previous one
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
myleott: Interesting! I'm not sure what's the cause. But there was something in the 1.8 release notes about backward hooks not behaving properly, which was fixed in 1.8. Search for "register_full_backward_hook" here: https://github.com/pytorch/pytorch/releases/tag/v1.8.0

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1001,7 +1048,20 @@ def _queue_wait_for_post_backward(self) -> None:
     def _wait_for_post_backward(self) -> None:
         """Wait for post-backward to finish. Only called on root instance."""
         assert self._is_root
-        self.assert_state(TrainingState.BACKWARD)
+        if self._has_params:
+            self.assert_state(TrainingState.BACKWARD_POST)
+        else:
+            self.assert_state(TrainingState.BACKWARD_PRE)
```
myleott, line 1054: nice :)

tests/nn/data_parallel/test_fsdp_multiple_forward.py
```diff
@@ -0,0 +1,79 @@
...
+
+    for _ in range(3):
+        in_data = torch.rand(64, 4).cuda()
+        in_data.requires_grad = True
+        out = model(in_data)
+        out.sum().backward()
+        optim.step()
+        optim.zero_grad()
+
+    model.assert_state(TrainingState.IDLE)
```
myleott, line 54: is it worth testing that the results are identical to DDP too? I do like the cleanliness of this test though 😄

myleott (approved):
Nice catch!

tests/nn/data_parallel/test_fsdp_multiple_forward.py
min-xu-ai: Yeah, I thought about it. My plan is to have another test with double-forward and conv/bn etc with numerical testing with DDP. This test is only for hooks I guess. :-)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: The new hook api is very interesting. I wonder what's the behavior will be like for the new hook function when we are doing nesting. We are not using the old API. I wonder if the way we do hooks is equivalent to the API there. I will do some digging.

Issue #517, opened 2021-03-13
[ShardedDDP] Grad norm disparity with DDP when reduce buckets are enabled


blefaudeux:
## 🐛 Bug
The grad norms become false when bucketing is enabled with ShardedDDP, spotted by Weiyi (@zhengwy888), many thanks !

blefaudeux:
cc @zhengwy888

PR #518, opened 2021-03-13
[fix][DRAFT] fixing a grad norm bug with ShardedDDP and buckets

blefaudeux:
# Before submitting

## What does this PR do?
Should fix https://github.com/facebookresearch/fairscale/issues/517, I added a new unit test to make sure that this was fixed, and that this would not be re-introduced, fairly easy to miss. Spotted by @zhengwy888, thank you !

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -462,9 +462,9 @@ class ShardedDataParallel(nn.Module):
                         bucket.buffer.mul_(self.world_size_scaling)
 
                         def cleanup() -> None:
-                            # Clear the reduced buffer
+                            # Clear the reduced buffers
                             # so that they don't interfere with clip grad norm
-                            if dst_rank != self.global_rank:
+                            if bucket.destination != self.global_rank:
                                 bucket.buffer.fill_(0.0)
 
                         # Reduce the bucket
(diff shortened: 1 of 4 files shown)


tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -173,6 +181,13 @@ def closure_sharded(input_tensor=input_tensor):
                 next(ddp_model.parameters()).requires_grad = not next(ddp_model.parameters()).requires_grad
                 check_same_model_params(sharded_ddp_model, ddp_model, f"Rank: {rank} - Trainability refresh {i} broke")
 
+                if clip_grad_norm:
```
blefaudeux, line 184: this is extensively tested on the OSS side, but in that case the gradients are not sharded, so this bug was not caught. I added a simple variant here, checking with one norm only but with the sharded gradients

blefaudeux:
Not ready, the unit test is doing a good job it seems.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -471,7 +477,7 @@ def reduce(*_: Any) -> None:
                                     group=self.process_group,
                                     async_op=True,
                                 ),
-                                callback=None,
+                                callback=cleanup,
```
zhengwy888, line 480: this should also be in _flush_reduce_calls()

blefaudeux: I forgot about that, I should store the callback in the bucket actually, that would be  cleaner and make it easier to call it from different places. It's not clear to me that it's what the unit test catches though

blefaudeux: ok, there was something wrong in the unit test, that second part makes sense

blefaudeux:
to give some more context, this showed up because fairseq was not using https://github.com/facebookresearch/fairscale/blob/master/fairscale/optim/oss.py#L270 but its own variant, which does not check for parameter ownership. Not sure how to guard against that in the future, a little too easy to trip over

blefaudeux:
bigger update incoming, couple of new corner cases fixed

PR #519, opened 2021-03-15
[feat] experimental.nn.multiprocess_pipe: re-implemented using rpc

msbaines:
Re-implemented nn.MultiProcessPipe based on the rpc pipeline parallel
tutorial:

(https://pytorch.org/tutorials/intermediate/dist_pipeline_parallel_tutorial.html).

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


msbaines:
@mrshenli @pritamdamania

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -0,0 +1,198 @@
...
+BOUNCE_TENSORS = True
+
+
+def _verify_module(module: List[LayerSpec]) -> None:
+    if not isinstance(module, List):
+        raise TypeError("module must be a list")
+
+    for elem in module:
+        if not isinstance(elem, tuple):
+            raise TypeError("module must be a list of tuple")
```
anj-s, line 34: Should we say "elem must be of ..."?

anj-s: Also, does it make sense to print out what was passed in as part of the error message? just for easier debugging.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -0,0 +1,198 @@
...
+    chunks: int = 1
+
+    #: The checkpoint mode to determine when to enable checkpointing. It is one
+    #: of ``'always'``, ``'except_last'``, or ``'never'``.
+    checkpoint: str = "never"
+
+    def __init__(
+        self,
+        module: List[LayerSpec],
+        *,
```
anj-s, line 148: Is this right? what does a single * mean?

msbaines: Everything after the '*' must be passed as a keyword argument.

fairscale/experimental/nn/multiprocess_pipe.py
msbaines: module is correct in this context. module is the name constructor param. The error is telling the API user they passed in a value of the wrong type.

msbaines: > Also, does it make sense to print out what was passed in as part of the error message? just for easier debugging.

I usually like to print things out but in this case is a constructor parameter so it should fail pretty quickly. If it fails, it essentially means you are using the API incorrectly.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class _ToHere(Module):
+    def __init__(self, device: str):
+        super().__init__()
+        self.device = device
+
+    def forward(self, x_rref: rpc.RRef) -> Tensor:  # type: ignore
+        if BOUNCE_TENSORS:
+            return x_rref.remote().cpu().to_here().to(self.device)
```
sidgoyal78, line 54: I was hoping to understand this better, this moves the tensor to the cpu first?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -0,0 +1,198 @@
...
+            workers.append(worker)
+            rmodule.append(rlayer)
+
+        self.chunks = chunks
+        self.checkpoint = checkpoint
+        self.module = module
+        self.workers = workers
+        self.rmodule = rmodule
+
+    def forward(self, x: Tensor) -> rpc.RRef:  # type: ignore
```
sidgoyal78, line 187: If i am understanding correctly, this forward() impl doesn't support checkpointing right?

sidgoyal78 (commented):
Thanks for the PR. I have left a few questions.

fairscale/experimental/nn/multiprocess_pipe.py
msbaines: rpc (until v1.8.0) only supported cpu tensors so I had to bounce to cpu first. I recently learned of GPU support in v1.8.0 which I am now trying to incorporate.

fairscale/experimental/nn/multiprocess_pipe.py
msbaines: It does not yet although it may support PyTorch's regular checkpointing. I will test that.

anj-s (approved):
Thank you for this awesome change! It is so simple to comprehend. Just curious if you had a chance to see how the perf compares to the current implementation.

tests/experimental/nn/test_multiprocess_pipe.py
```diff
@@ -0,0 +1,184 @@
...
+    assert y.shape == torch.Size([8, 4])
+    assert y.requires_grad is True
+    assert torch.allclose(y.sum(), expected_sum), f"{y.sum()} != {expected_sum}"
+
+
+@rpc_test(world_size=2)
+@pytest.mark.parametrize("devices", DEVICES)
+def backward(devices):
+    torch.random.manual_seed(3)
+    criterion = DistributedLoss(torch.nn.MSELoss)
```
pritamdamania87, line 153: `DistributedLoss` is really a nice abstraction to deal with this problem!

PR #520, opened 2021-03-15
[utils] Compute the grad norm in fp32 no matter what

blefaudeux:
# Before submitting

## What does this PR do?
It used to be the case, top my head that's something that @joshim5 had to introduce following accuracy issues when computing the accumulated grad norm. Compute the grad norm in fp32 then cast back to the parameter type

## PR review

## Did you have fun?


myleott:
Yes, this makes sense. I'm not even sure you need/want to cast back to the original parameter type. Should be fine to return in FP32.

zhengwy888 (approved):
looks great!

blefaudeux:
> Yes, this makes sense. I'm not even sure you need/want to cast back to the original parameter type. Should be fine to return in FP32.

yeah I thought about that, I figured that somebody who had a hard type constraint would expect it it to have the same type as the parameters ?

PR #521, opened 2021-03-15
[feat][fix] ShardedDDP - better tests, removing broken reduce fp16 (to be reintroduced properly)

### LINKED ISSUE #517: [ShardedDDP] Grad norm disparity with DDP when reduce buckets are enabled

blefaudeux:
# Before submitting

## What does this PR do?
- Fixes #517 as far as I can tell
- Removes the reduce fp16 option from ShardedDDP, which was broken, other PR incoming to re-implement it properly
- Adds a couple of cases in the DDP parity unit test, among which
-- check that the grads are the same
-- handle AMP on/off
-- handle Gloo/NCCL

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -78,10 +78,6 @@ class ShardedDataParallel(nn.Module):
                 compatible with PytorchAMP.
 
     .. warning:
-        Pytorch AMP support is limited to two cases: either the whole model supports fp16, and the reduce buffers can be used, or some parts of the
-        model do not support fp16 and in that case reduce buffer size should be set to 0.
-
-    .. warning:
         If `auto_refresh_trainable` is set to `True` (this is the default) then any trainability change in the model graph will be handled
         automatically.
         If `auto_refresh_trainable` is set to `False`, ShardedDDP will not refresh its assumptions with respect to trainable parameters
@@ -188,9 +184,6 @@ class ShardedDataParallel(nn.Module):
 
         # Optionally check whether the trainable parameters have changed
         if self.auto_refresh_trainable:
-            # Re-establish the buckets if:
-            # - there are new trainable or untrainable parameters
-            # - we're under an autocast context and the grads will be fp16
             trainable_mask = list(map(_trainable, self._all_params))
             if trainable_mask != self._reference_trainable_mask:
                 logging.warning("ShardedDDP detected that the trainable params changed, updating the partitioning")
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -462,9 +462,9 @@ class ShardedDataParallel(nn.Module):
                         bucket.buffer.mul_(self.world_size_scaling)
 
                         def cleanup() -> None:
-                            # Clear the reduced buffer
+                            # Clear the reduced buffers
                             # so that they don't interfere with clip grad norm
-                            if dst_rank != self.global_rank:
+                            if bucket.destination != self.global_rank:
                                 bucket.buffer.fill_(0.0)
 
                         # Reduce the bucket
@@ -620,12 +620,18 @@ class ShardedDataParallel(nn.Module):
                     bucket.buffer.mul_(self.world_size_scaling)
 
                     # Reduce the bucket
+                    def cleanup() -> None:
+                        # Clear the reduced buffers
+                        # so that they don't interfere with clip grad norm
+                        if bucket.destination != self.global_rank:
+                            bucket.buffer.fill_(0.0)
+
                     self._work_handles.append(
                         Workhandle(
                             handle=dist.reduce(
                                 tensor=bucket.buffer, dst=bucket.destination, group=self.process_group, async_op=True,
                             ),
-                            callback=None,
+                            callback=cleanup,
                         )
                     )
                     bucket.sent = True
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -409,7 +409,8 @@ class ShardedDataParallel(nn.Module):
                         if dst_rank != self.global_rank:
                             param.grad = None
 
-                    torch.cuda.synchronize()
+                    if self.backend == dist.Backend.NCCL:
+                        torch.cuda.synchronize()
 
                     # Async reduce for this buffer, log the future
                     self._work_handles.append(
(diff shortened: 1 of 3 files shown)


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -300,8 +290,8 @@ def reduce(self) -> None:
         # Trigger all the current BW hooks
         _ = map(lambda x: x(), self._grad_accs)
 
-        # Make sure that all the futures are consumed
-        self._consume_work_handles()
+        # Make sure that the buckets are flushed and all the futures are consumed
+        self._flush_reduce_calls()
```
blefaudeux, line 294: manual reduce case, spotted by the unit test, this was not flushing the buckets (so manual reduce + unused params would not be correct)

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -413,22 +403,16 @@ def reduce(*_: Any) -> None:
...
-                        else:
-                            assert param.grad is not None
-                            param.grad.data = param.grad.data.to(dtype=param.dtype)
 
                     # Async reduce for this buffer, log the future
                     self._work_handles.append(
                         Workhandle(
                             handle=dist.reduce(
-                                tensor=param.grad.data,
+                                tensor=param.grad,
```
blefaudeux, line 415: to my understanding this does not change anything -unit test agrees-, just better not to use .data when not needed I think

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -461,6 +445,12 @@ def reduce(*_: Any) -> None:
                         # Normalize the bucket in one go
                         bucket.buffer.mul_(self.world_size_scaling)
 
+                        def cleanup() -> None:
```
blefaudeux, line 448: this should not be strictly needed if users
- use OSS.clip_grad_norm
- use grad accumulation with the no_sync() context
but in case one of these was not done, it would lead to gradients being both in the original buckets and in the reduced buckets, which can lead to errors. This is then arguably cleaner, reduced buckets are empty

fairscale/optim/oss.py
```diff
@@ -247,6 +248,7 @@ def clip_grad_norm(
         Arguments:
             max_norm (float or int): max norm of the gradients
             norm_type (float or int): type of the used p-norm. Can be ``'inf'`` for infinity norm.
+            skip_nan_grads (bool): auto-skip scaling if some grads are NaNs (which can happen with Pytorch AMP, in that case the model update is skipped)
```
blefaudeux, line 251: if NaN grads are ok (basically if you're using torch AMP), then this can be passed

fairscale/optim/oss.py
```diff
@@ -286,10 +288,20 @@ def clip_grad_norm(
             total_norm = total_norm ** (1.0 / norm_type)
 
         clip_coef = torch.tensor(max_norm, dtype=total_norm.dtype, device=total_norm.device) / (total_norm + 1e-6)
+
+        if skip_nan_grads and torch.isnan(clip_coef):
+            return total_norm
+
+        assert not torch.isnan(clip_coef), (
```
blefaudeux, line 295: by default we assume that normalizing by a NaN is most probably the sign that something is broken, better signal it early

fairscale/optim/utils.py
```diff
@@ -120,5 +120,6 @@ def calc_grad_norm(parameters: List[torch.nn.Parameter], p: float) -> torch.Tens
     if p == inf:
         local_norm = max(par.grad.detach().abs().max() for par in parameters)  # type: ignore
     else:
-        local_norm = torch.norm(torch.stack([torch.norm(par.grad.detach(), p) for par in parameters]), p)  # type: ignore
+        # Compute the norm in full precision no matter what
+        local_norm = torch.norm(torch.stack([torch.norm(par.grad.detach(), p, dtype=torch.float32) for par in parameters]), p).to(dtype=parameters[0].dtype)  # type: ignore
```
blefaudeux, line 124: fixed in https://github.com/facebookresearch/fairscale/pull/520, in some cases the sum on fp16 could overflow. The cost is negligible basically, and this minimizes the computations done in fp16

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -136,35 +138,42 @@ def step():
...
+                    _scaler.step(_optimizer)
+                    _scaler.update()
 
             check_same_model_params(sharded_ddp_model, ddp_model, f"Rank: {rank} - Step {i} broke")
 
+            # Check that the two grad norm are equivalent
+            # NOTE: The grads can occasionally be NaNs, the scaler will skip the step in that case
+            # This is not ShardedDDP specific. If the grads are not NaN for DDP then they should also
+            # be valid for ShardedDDP
+            if clip_grad_norm:
```
blefaudeux, line 169: *new* check that clip_grad_norm is also at parity in between DDP and ShardedDDP

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -201,10 +204,11 @@ def closure_sharded(input_tensor=input_tensor):
 @pytest.mark.parametrize("reduce_buffer_size", [0, 2 ** 20])
 @pytest.mark.parametrize("grad_accumulation", [True, False])
 @pytest.mark.parametrize("change_train_graph", [True, False])
-@pytest.mark.parametrize("fp16_reduction", _test_fp16_reduction)
-def test_ddp_parity(reduce_buffer_size, grad_accumulation, change_train_graph, fp16_reduction):
+@pytest.mark.parametrize("clip_grad_norm", [True, False])
```
blefaudeux, line 207: bigger test matrix, since #517 showed that the current test matrix was not enough..

blefaudeux:
no NaNs, but the different in the grad norm is noticeably bigger with torch 1.6 and 1.7, worth investigating I think

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -413,22 +404,19 @@ def reduce(*_: Any) -> None:
...
                     # Future work includes clearing up the buffer if possible
                     def cleanup() -> None:
                         if dst_rank != self.global_rank:
                             param.grad = None
-                        else:
-                            assert param.grad is not None
-                            param.grad.data = param.grad.data.to(dtype=param.dtype)
+
+                    if self.backend == dist.Backend.NCCL:
+                        torch.cuda.synchronize()
```
blefaudeux, line 413: doing that completely tanks the perf, but the results are not correct if I remove just this line (same for the 2 other same lines below).. cc @ngimel @mrshenli, any other way to make sure that the NCCL stream(s) wait for a given event on the current one ?

blefaudeux:
changing more things to get to the bottom of all the cuda sync issues

PR #527, opened 2021-03-17
[feature] FSDP: enable pytorch SyncBN

min-xu-ai:
- not fully validated yet but at least not asserting
- this enables VISSL to move forward with its next PR

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/utils/parallel.py
+++ b/fairscale/utils/parallel.py
@@ -45,7 +45,7 @@ def validate_process_group(device: torch.device, process_group: ProcessGroup) ->
         )
 
 
-def enable_pytorch_sync_bn(module: torch.nn.Module):
+def enable_pytorch_sync_bn(module: torch.nn.Module) -> None:
     """Call _specify_ddp_gpu_num for all pytorch SyncBN layers so that it
        is happily running even without DDP. E.g. this is used by FSDP.
     """
@@ -54,4 +54,4 @@ def enable_pytorch_sync_bn(module: torch.nn.Module):
             # Number "1" below meant to be the number of GPUs for each DDP worker.
             # (i.e. "device_ids" in DDP. As far as I see, the value is not actually
             # used, but this call needs to be made to avoid an exception.
-            layer._specify_ddp_gpu_num(1)
+            layer._specify_ddp_gpu_num(1)  # type: ignore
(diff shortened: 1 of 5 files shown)


fairscale/utils/parallel.py
```diff
@@ -43,3 +43,15 @@ def validate_process_group(device: torch.device, process_group: ProcessGroup) ->
...
+def enable_pytorch_sync_bn(module: torch.nn.Module) -> None:
+    """Call _specify_ddp_gpu_num for all pytorch SyncBN layers so that it
+       is happily running even without DDP. E.g. this is used by FSDP.
+    """
+    for layer in module.modules():
+        if isinstance(layer, torch.nn.modules.SyncBatchNorm):
+            # Number "1" below meant to be the number of GPUs for each DDP worker.
+            # (i.e. "device_ids" in DDP. As far as I see, the value is not actually
+            # used, but this call needs to be made to avoid an exception.
+            layer._specify_ddp_gpu_num(1)  # type: ignore
```
myleott, line 57: Wow, I didn't realize this was being done in upstream PyTorch DDP, nice find 😄 

For reference: https://github.com/pytorch/pytorch/blob/e09e97ebf9946aa9b9582b27811bb7a9c3de392f/torch/nn/parallel/distributed.py#L1284-L1290

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -0,0 +1,91 @@
...
+            self.bn = FSDP(
+                BatchNorm2d(2),
+                mixed_precision=False,
+                process_group=dist.new_group(ranks=[rank]),
+                flatten_parameters=False,
+            )
+
+        def forward(self, x):
+            x = self.conv(x)
+            x = self.bn(x)
```
myleott, line 48: is the output type here always float32? Perhaps add:
```python
assert x.dtype == torch.float32
```

myleott (approved):
Awesome!

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: It is a very good suggestion to add that. The result is something I wasn't expecting. The input to this "FP32 BN" can be either fp16 or fp32 and the output always match the input ind type. The internal weights of BN is kept at FP32. Let me check what kind of precision is in fact required in vision.

min-xu-ai: I also verified that autocast doesn't change this behavior for BN. It seems that we need to explicitly manage it for BN.

min-xu-ai: OK, I found out the answer. I think we only keep the weights fp32 and inputs are still fp16, base on the description here:

image

from this page: https://nvidia.github.io/apex/amp.html

PR #528, opened 2021-03-17
[DRAFT][SDP][BigPR] Base all sync on torch futures

blefaudeux:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

@girifb

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -8,12 +8,11 @@ A nn.Module wrapper to go with a Sharded Optimizer in order to handle targeted g
 reduction automatically.
 """
 
-from collections import deque
 import contextlib
 import functools
 from itertools import chain
 import logging
-from typing import Any, Callable, Deque, Dict, Generator, List, Optional, Union
+from typing import Any, Callable, Dict, Generator, List, Optional, Union
 
 import torch
 from torch import nn
@@ -22,7 +21,6 @@ import torch.distributed as dist
 
 from fairscale.nn.misc import GradBucket
 from fairscale.optim import OSS
-from fairscale.optim.utils import Workhandle
 
 
 def _trainable(param: torch.Tensor) -> bool:
@@ -107,11 +105,6 @@ class ShardedDataParallel(nn.Module):
         self.enable_broadcast_buffers = broadcast_buffers
         self.auto_refresh_trainable = auto_refresh_trainable
         self.reduce_fp16 = reduce_fp16
-        if reduce_buffer_size > 0 and reduce_fp16:
-            self.reduce_fp16 = False
-            logging.warning(
-                "fp16 gradient reduction is not compatible with reduction buffers, which are requested. fp16 grad reduction is deactivated."
-            )
 
         # Handle a no_sync() context which prevents the gradient synchronization,
         # accumulate in place
@@ -120,7 +113,9 @@ class ShardedDataParallel(nn.Module):
 
         # Communication related attributes
         self.process_group = process_group if process_group is not None else dist.group.WORLD
-        self.backend = dist.get_backend(self.process_group)
+        assert (
+            dist.get_backend(self.process_group) == dist.Backend.NCCL
+        ), "SharedDDP requires a backend with torch.futures support, for now NCCL is the only option"
         self.world_size_scaling = 1.0 / dist.get_world_size(self.process_group)  # > 0
         self.reference_global_rank = OSS.get_global_rank(self.process_group, 0)  # picking rank 0 as the reference
         self.rank = dist.get_rank(self.process_group)
@@ -154,6 +149,7 @@ class ShardedDataParallel(nn.Module):
         )
         self._trainable_params: List[torch.Tensor] = []
         self._grad_to_be_reduced: List[bool] = []
+        self._grad_fp16: List[Any] = []
         self._trainable_param_to_rank: Dict[torch.Tensor, int] = {}
         self._reference_trainable_mask = list(map(_trainable, self._all_params))
 
@@ -179,6 +175,7 @@ class ShardedDataParallel(nn.Module):
         # - setup backward hooks which will be called by Torch's autograd in due time
         self._grad_accs: List[Callable] = []
         self._manual_reduce: List[Callable] = []
+        self._grad_hooks: List[Any] = []
 
         # passing a handle to torch.nn.SyncBatchNorm layer
         self._passing_sync_batchnorm_handle(self.module)
@@ -187,7 +184,7 @@ class ShardedDataParallel(nn.Module):
         if sync_models_at_startup:
             self._sync_params_and_buffers()
 
-        self._work_handles: Deque[Workhandle] = deque()
+        self._futures: List[torch.futures.Future] = []
         self._bucket_flush_callback_set = False
 
         self.refresh_trainable()
@@ -200,11 +197,7 @@ class ShardedDataParallel(nn.Module):
 
         # Optionally check whether the trainable parameters have changed
         if self.auto_refresh_trainable:
-            trainable_mask = list(map(_trainable, self._all_params))
-            if trainable_mask != self._reference_trainable_mask:
-                logging.warning("ShardedDDP detected that the trainable params changed, updating the partitioning")
-                self.refresh_trainable()
-                self._reference_trainable_mask = trainable_mask
+            self._auto_update_train_graph()
 
         if self.enable_broadcast_buffers:
             # NCCL communications are on a different stream, needs to be blocking
@@ -267,13 +260,14 @@ class ShardedDataParallel(nn.Module):
         """ If the module trainability has changed, update all the assumptions """
 
         # Make sure that this is not done while gradients are waiting to be reduced (if no_sync context for instance)
-        assert not functools.reduce(
+        assert self.should_accumulate_grads or not functools.reduce(
             lambda x, y: x or y, self._grad_to_be_reduced, False
         ), "Grads waiting to be reduced: {}".format(self._grad_to_be_reduced)
 
         self._trainable_params = list(filter(lambda x: x.requires_grad, self._all_params))
         self._trainable_params.sort(key=lambda x: x.numel())
         self._grad_to_be_reduced = [True for _ in self._trainable_params]
+        self._grad_fp16 = [None for _ in self._trainable_params]
 
         self._trainable_param_to_rank = {}
         for optim in self.sharded_optimizers:
@@ -288,7 +282,7 @@ class ShardedDataParallel(nn.Module):
                     for param in filter(lambda x: x.requires_grad, device_params):
                         self._trainable_param_to_rank[param] = optim.param_to_rank[param]
 
-        self._setup_bucket_strategy()
+        self._setup_buckets()
         self._setup_backward_hooks()
 
     def reduce(self) -> None:
@@ -306,8 +300,8 @@ class ShardedDataParallel(nn.Module):
         self._bucket_flush_callback_set = True  # no need to flush in the end, we own the callback execution
         _ = list(map(lambda x: x(), self._manual_reduce))
 
-        # Make sure that all the futures are consumed
-        self._consume_work_handles()
+        # Make sure that the buckets are flushed and all the futures are consumed
+        self._flush_reduce_calls()
 
     @torch.no_grad()
     def sync_buffers(self, blocking: bool = False) -> None:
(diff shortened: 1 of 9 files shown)


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -416,35 +403,28 @@ def reduce(*_: Any) -> None:
 
                     # Make sure that this is not fired twice
                     self._grad_to_be_reduced[index] = False
-                    param.grad.mul_(self.world_size_scaling)
 
-                    if self.reduce_fp16:
-                        param.grad.data = param.grad.data.half()
+                    param.grad.mul_(self.world_size_scaling)
```
msbaines, line 407: What happened to the` if self.reduce_fp16`

fairscale/optim/oss.py
```diff
@@ -292,10 +294,20 @@ def clip_grad_norm(
...
+        assert not torch.isnan(clip_coef), (
+            "Clipping grads with a NaN, cannot be right.\n"
+            + f"Local norm: {local_norm.item()}. Total norm: {total_norm.item()}"
+        )
+
         if clip_coef < 1:
-            for device, device_params in self.per_device_params.items():
-                for p in filter(lambda x: x.grad is not None, device_params[self.rank]):
-                    p.grad.detach().mul_(clip_coef.to(device))  # type: ignore
+            with torch.no_grad():
```
msbaines, line 307: This seems like unrelated cleanup. With large changes it is much easier to review if there is a single thesis.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: it did not give the correct results, I did not test it well enough. It can be re-implemented on top of this PR, I was genuinely trying to keep things "small"

fairscale/optim/oss.py
blefaudeux: you're right, I'll move that to another PR

blefaudeux:
I was planning to write down some explanations @msbaines, that's why it was in draft. Writing some history here: basically this started with a report from @zhengwy888 that he was seeing a different grad norm with shardedDDP when the buckets were activated (which led to https://github.com/facebookresearch/fairscale/issues/517).

I could repro that in a new unit test, which is part of this PR, turns out that the synchronization in between compute and comms was not correct, and shardedDDP did in fact produce different results from DDP. I really tried to get things to work without the torch futures, but the TL; DR is that you don't get to synchronize with `torch.distributed.reduce()` for instance, because this goes to a "hidden" cuda stream (a pool actually). The handles that `torch.distributed` returns are not a good synchronization primitive, they just mark the op being inserted in the cuda stream, not completeness. So the only reasonable way that I could find, which prove reliable, was to use `torch.futures`, which implied rewriting a fair bunch of things. Gloo does not support torch.futures, hence the trail of changes in the unit test, other than the new ones the other ones are unchanged (just backend limited to NCCL).

If you can think of another method I would love to learn about it. cc @mrshenli just in case I writing something wrong in there

fairscale/optim/oss.py
```diff
@@ -245,6 +245,7 @@ def clip_grad_norm(
         max_norm: Union[float, int],
         norm_type: Union[float, int] = 2.0,
         filter_params_fn: Callable[[Any], Any] = None,
+        skip_nan_grads: bool = False,
```
blefaudeux, line 248: will be moved to another PR, unrelated

tests/nn/data_parallel/test_sharded_ddp_features.py
```diff
@@ -22,7 +22,6 @@
 from fairscale.utils.testing import (
```
blefaudeux, line 22: no changes in this file except from
- removing duplicated test with the other sharded_ddp test file (probably a botched merge)
- removing gloo tests

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -201,28 +196,30 @@ def closure_sharded(input_tensor=input_tensor):
 @pytest.mark.parametrize("reduce_buffer_size", [0, 2 ** 20])
 @pytest.mark.parametrize("grad_accumulation", [True, False])
 @pytest.mark.parametrize("change_train_graph", [True, False])
-@pytest.mark.parametrize("fp16_reduction", _test_fp16_reduction)
-def test_ddp_parity(reduce_buffer_size, grad_accumulation, change_train_graph, fp16_reduction):
+@pytest.mark.parametrize("clip_grad_norm", [True, False])
```
blefaudeux, line 199: two new cases, which showed the problem that @zhengwy888 had spotted: check clip_grad_norm parity, with and without AMP, without and without buckets, with and without grad accumulation, and change in training graph. This reliably showed an error when the buckets were activated

blefaudeux: Screen Shot 2021-03-16 at 10 53 09 AM

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -290,341 +285,9 @@ def run_ddp_parity_two_optim(rank, world_size, backend, temp_file_name, reduce_b
...
-    backend = dist.Backend.NCCL
     mp.spawn(
         run_ddp_parity_two_optim,
-        args=(world_size, backend, tempfile.mkstemp()[1], reduce_buffer_size),
-        nprocs=world_size,
-        join=True,
-    )
-
-
-def run_test_two_inputs(rank, world_size, backend, device, temp_file_name, reduce_buffer_size):
```
blefaudeux, line 302: all the test below were duplicates, probably a merge which failed on somebody. These are not ddp parity, they are test of sharded_ddp features and are handled in the other test file

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -59,9 +58,6 @@ class ShardedDataParallel(nn.Module):
             (default: True) Check whether the parameters trainability (`requires_grad`) has changed and update both ShardedDDP
             and OSS automatically if this is the case. If set to False, `refresh_trainable()` needs to be called anytime
             a parameter is frozen or unfrozen.
-        reduce_fp16 (bool):
```
blefaudeux, line 62: this was broken and could be re-introduced later, not too complicated to do once the syncs are correct. I did try to keep this PR as small as possible

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -317,18 +304,15 @@ def sync_buffers(self, blocking: bool = False) -> None:
             blocking (bool): wait for the operation to conclude.
         """
 
-        work_handles = []
+        futures = []
 
         for buffer in self.module.buffers(recurse=True):
-            work_handles.append(
```
blefaudeux, line 323: basically the crux of the issue is that, the handles are not a good sync primitive, they are just not correct, and the cuda events are not accessible because the stream being used is not accessible

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -339,12 +323,18 @@ def zero_grad(self, set_to_none: bool = False) -> None:
                 See :meth:`torch.optim.Optimizer.zero_grad` for details.
         """
 
+        if self.auto_refresh_trainable:
```
blefaudeux, line 326: this is orthogonal but a bugfix: if the model trainability changed, and people zero_grad, then some of the gradients could skip being set to zero, which would break the subsequent grad norm computation..

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -491,6 +454,10 @@ def _setup_backward_hooks(self) -> None:
         This makes the gradient reduction automatic whenever there's a backward pass
         """
 
+        # Empty the possible previous hooks
```
blefaudeux, line 457: suggested in the doc, doing that for correctness, I'm not sure of the effect. Remove the previous hooks before attaching new ones

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -551,82 +515,68 @@ def _setup_bucket_strategy(self) -> None:
...
+        self._should_bucket_grad = []
+        self._should_bucket_grad = [False for _ in range(len(self._trainable_params))]
 
-        for param in self._trainable_params:
+        for index, param in enumerate(self._trainable_params):
             device = param.device
             dst_rank = self._trainable_param_to_rank[param]
 
             if param.device not in self.buckets.keys():
-                self.buckets[param.device] = [
```
blefaudeux, line 560: this could lead to a subtle over provision, if not all ranks where present for all devices. Moving that to allocating the bucket on demand, when it's really required

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -551,82 +515,68 @@ def _setup_bucket_strategy(self) -> None:
...
+            bucket.shrink()
 
-            else:
-                self._should_bucket_grad.append(False)
+            # - store the callback which releases the bucket
+            def cleanup(fut: torch.futures.Future) -> None:
+                if bucket.destination != self.global_rank:
+                    fut.wait()
+                    # FIXME: Autograd dies with this
+                    # bucket.collapse()
```
blefaudeux, line 555: this is an extra, not done in the current (master) version of ShardedDDP: release the bucket when reduced, rebuild in the next FW. For some reason autograd segfaults on that, opinions welcome

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -551,82 +515,68 @@ def _setup_bucket_strategy(self) -> None:
...
 
-        self._bucket_list = list(chain(*[self.buckets[device] for device in self.buckets.keys()]))
+            bucket.callback = cleanup
 
-        # Resize the buckets to remove lost space in the end
-        for bucket in self._bucket_list:
-            bucket.buffer.resize_(bucket.fill)
-            bucket.sent = True
-
-    def _consume_work_handles(self) -> None:
```
blefaudeux, line 596: all this was nice, but handle.wait() does not do the same thing as (torch.future).wait()... this was an expensive mistake

fairscale/nn/misc/grad_bucket.py
```diff
@@ -0,0 +1,148 @@
...
+                self._add_grad_as_view(p)
+
+            self._is_collapsed = False
+
+    @torch.no_grad()
+    def scale_and_reduce(self, scale: float) -> torch.futures.Future:
+        assert self._buffer is not None, "Cannot send a collapsed bucket, please rebuild"
+
+        self._buffer.mul_(scale)
+        torch.cuda.synchronize()
```
blefaudeux, line 111: this is very expensive in terms of time, but the results are not correct without it. If a reviewer can think of a better synchronization that would be great..

fairscale/nn/misc/grad_bucket.py
```diff
@@ -0,0 +1,148 @@
+from typing import Any, Callable, List, Optional, Union
+
```
blefaudeux, line 2: the idea behind that was to simplify sharded_ddp, which is already complicated enough. The reasoning is the same as before, just with the torch.futures and not workhandle

blefaudeux:
(rebased and some unrelated changes removed, still not ready for review)

tests/nn/data_parallel/test_sharded_ddp_features.py
blefaudeux: test removal has been moved to another PR, now just gloo backend removal as expected

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: resolved in the newer version

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
blefaudeux: checked in with another PR

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
blefaudeux: checked in in another PR

blefaudeux:
This kills compatibility with torch 1.6 it seems, looks like the torch.distributed calls are not torch.futures compatible :(

blefaudeux:
not supported on torch 1.6, will need to wait for 1.9 so that we deprecate it

PR #533, opened 2021-03-18
[perf] [FSDP] micro-optimization for memory usage

myleott:
When doing mixed precision training with reshard_after_forward (i.e., nested FSDP, ZeRO-3 mode), we don't need the FP16 param shard, since it will be cast back from the FP32 shard before the backward anyway. This saves a little bit of memory.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -814,6 +814,8 @@ class FullyShardedDataParallel(nn.Module):
 
         if self.reshard_after_forward:
             self._free_full_params()
+            if self.mixed_precision:
+                self._free_fp16_param_shard()
 
         # Switch to main FP32 param shard. We maintain this invariant throughout
         # the code, i.e., ``p.data == p._fp32_shard`` after each function. This


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -814,6 +814,8 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
 
         if self.reshard_after_forward:
             self._free_full_params()
+            if self.mixed_precision:
+                self._free_fp16_param_shard()
```
min-xu-ai, line 818:
```suggestion
                # Save some memory by freeing fp16 shard since it will be recreated from
                # fp32 shard on backward. TODO (Min): may not need to free and recast
                # for the very last block.
                self._free_fp16_param_shard()
```

min-xu-ai (approved):
very cool

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: never mind about the suggestion since it is merged. It is minor.

PR #537, opened 2021-03-19
[FSDP][feature] optimizer state dict save and load

sshleifer:
### Overview
```python


# save
fsdp = FSDP(world_size=4)
optim = Adam(fsdp.parameters())
full_state_dict = fsdp.gather_full_optim_state_dict(optim, recipient_rank=0))
# this is None if you are not on a recipient rank
# recipient_rank=None  performs the same consolidation on all ranks.

# load with different world size
fsdp2 = FSDP(world_size=2)
optim2 = Adam(fsdp2.parameters()
state_shard = fsdp2.get_shard_from_optim_state_dict(full_state_dict)
optim2.load_state_dict(state_shard)
```

### Future Work
- fairseq integration
- support `flatten_parameters=False`
- support param groups
- test more nested setups. An FSDP with no params could certainly break this.

On the fairseq side, I tested running with 4 gpus and loading with 2 and this worked.

### Assumptions
(0) `flatten_parameters=True`

(1) if there is a tensor in optimizer state, it is the same size and corresponds to a tensor in model state. If there are singleton tensors in the optimizer, or tensors that correspond to the average update for a column of params (so shaped differently), things will break.

(2) We assume that these two lists are the same if we account for padding:

```python
mlist = [sum(m._param_numels) for m in self.modules() if isinstance(m, FullyShardedDataParallel)]
params = [p.numel() for p in self.parameters()]
```
we use this assumption to call 
`mlist[i].get_params_view(flat_param=params_unpadded[i])`.

### New overhead introduced
- `_get_shard` now returns how many padding elements it introduced.


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,244 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
                 traceback.print_stack()
             raise ValueError(msg)
 
+    # Optim State dict functions
```
sshleifer, line 1355: I considered moving these to a separate `FSDPOptimizerMixin` in `fsdp_optimizer_utils.py`, but decided it wasn't really a mixin since it depends heavily on FSDP.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+
+    def consolidate_optim_state_dict(self, optim: torch.optim.Optimizer, recipient_rank: int = 0) -> None:
+        """Update the consolidated state_dict list, one per rank.
+
+        Arguments:
```
myleott, line 1360: very minor nit: convention seems to be calling this "Args:"

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+            recipient_rank (int): on which rank to materialize the full state dict.
+            -1 is a special value, which means that all ranks should have the state
```
myleott, line 1362: thoughts on making `None` the special value? I had original (mistakenly) assumed -1 meant the last worker

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
                 traceback.print_stack()
             raise ValueError(msg)
 
+    # Optim State dict functions
+
+    def consolidate_optim_state_dict(self, optim: torch.optim.Optimizer, recipient_rank: int = 0) -> None:
```
myleott, line 1357: I like `recipient_rank=0` as a default here, makes it harder for people to mess up

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+        """Update the consolidated state_dict list, one per rank.
+
+        Arguments:
+            recipient_rank (int): on which rank to materialize the full state dict.
+            -1 is a special value, which means that all ranks should have the state
+
+        .. warning: This needs to be called on all replicas"""
+        _default_device = torch.device("cuda")
```
myleott, line 1365: how about calling `self._lazy_init()` and then using `self.compute_device`? That's what we're doing for `apply`/`summon_all_params`

It also makes it work in case the user hasn't set the current_device properly

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+        should_collect_state = self.rank == recipient_rank or recipient_rank == -1
+        should_send_state = (self.rank != recipient_rank and recipient_rank != -1) or recipient_rank == -1
+        _all_optimizer_states: List[Dict[str, Any]] = []
+        for rank in range(self.world_size):
+            if rank == self.rank:
+                sd = optim.state_dict()
+                sd["num_padded"] = [m.num_padded for m in self.modules() if isinstance(m, FullyShardedDataParallel)]
+                if should_collect_state:
+                    _all_optimizer_states.append(
+                        recursive_copy_to_device(sd, non_blocking=True, device=torch.device("cpu"))
```
myleott, line 1379: remove `non_blocking`, it will actually be blocking no matter what, since GPU->CPU transfer can only be non-blocking if the destination is pinned memory.

Roughly speaking (not tested for syntax):
```python
x = torch.ones(5).cuda()
buffer = torch.zeros(5)
buffer.copy_(x, non_blocking=True)  # blocking
buffer.pin_memory_()
buffer.copy_(x, non_blocking=True)  # non-blocking
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+                state_to_share = (
+                    sd if should_send_state else torch.tensor([0], dtype=torch.uint8, device=_default_device)
+                )
+                broadcast_object(
+                    state_to_share, src_rank=self.rank, group=self.process_group, dist_device=_default_device,
+                )
+            else:
+                # Fetch the optim state from the other replicas
+                replica_state = broadcast_object(
+                    torch.tensor([0], dtype=torch.uint8, device=_default_device),
```
myleott, line 1392: replace with `obj=None` to make clearer there's no data being sent

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+            if rank == self.rank:
+                sd = optim.state_dict()
+                sd["num_padded"] = [m.num_padded for m in self.modules() if isinstance(m, FullyShardedDataParallel)]
+                if should_collect_state:
+                    _all_optimizer_states.append(
+                        recursive_copy_to_device(sd, non_blocking=True, device=torch.device("cpu"))
+                    )
+
+                # Sync with other replicas
+                state_to_share = (
```
myleott, line 1383: doesn't every rank need to share its own state at some point?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+                    torch.tensor([0], dtype=torch.uint8, device=_default_device),
+                    src_rank=rank,
+                    group=self.process_group,
+                    dist_device=_default_device,
+                )
+
+                if should_collect_state:
+                    _all_optimizer_states.append(
+                        recursive_copy_to_device(replica_state, non_blocking=True, device=torch.device("cpu"))
+                    )
```
myleott, line 1401: can this be rearranged to remove some duplication? Something like:
```python
for rank in range(self.world_size):
    if rank == self.rank:
        state = optim.state_dict()
        sd["num_padded"] = ...
        state = broadcast_object(state, src_rank=rank, ...)
    else:
        state = broadcast_object(None, src_rank=rank, ...)

    if should_collect_state:
        _all_optimizer_states.append(recursive_copy_to_device(state, device=torch.device("cpu"))
```

myleott (commented):
About half-way through, leaving initial comments and will post rest in second batch

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
sshleifer: Just copy pasted this func from OSS. I think the reason for `should_collect_state` is to save useless communication from rank 0.

sshleifer: I have the simplified implem working with `torch.distributed.broadcast_object_list`.
I no longer need compute_device. Still calling lazy_init_for safety.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
sshleifer: Thx for teaching!

min-xu-ai:
Looks like really cool stuff. Some high level questions about the context:

1. in the example, do users need to put the optimizer back into it is original state after the consolidation? If not, perhaps make a comment about it in the example?
2. this assumes there is enough GPU memory to hold the state at each or one rank? what's the solution for very large models?

It seems that this is only needed when we change world size between save/restore? If the world size not changed, normal save/restore with the only the sharded data is OK?

sshleifer:
@min-xu-ai 
1) No resetting is needed by the user. This doesn't mutate `optimizer` it just combines `optimizer.state_dict` between ranks. I'll add a comment and test to that effect.
2) Consolidation is happening in CPU memory (the cast is in the `consolidate` method). If we see use cases where there is not enough CPU memory to handle consolidation on one node, we will iterate :)

myleott (commented):
Do we envision a use case for calling `consolidate_optim_state_dict` without calling `gather_full_optim_state_dict`?

If not, perhaps simplify interface to:
```python
fsdp = FSDP(world_size=4)
optim = Adam(fsdp.parameters())
full_state_dict = fsdp.gather_full_optim_state_dict(optim, recipient_rank=-1)
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1352,242 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+                            next_global_param_id += 1
+                        global_id = local_to_global_param_id[local_id][i]
+                        if global_id not in new_state_dict["state"]:
+                            new_state_dict["state"][global_id] = copy.deepcopy(constant_state[local_id])
+                        assert k not in new_state_dict["state"][global_id], f"already added {k} to new[{global_id}]"
+                        new_state_dict["state"][global_id][k] = param_view
+
+            if next_global_param_id == 0:  # stateless optimizer
+                num_params = sum([len(m._param_numels) for m in instance_list])  # type: ignore
+                new_state_dict["param_groups"][pg_id]["params"] = list(range(num_params))
```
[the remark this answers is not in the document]
sshleifer: Are you talking about `list(range(num_params))`? If so, it affects both cases.
I'll leave a TODO

fairscale/nn/data_parallel/fsdp_optim_utils.py
```diff
@@ -0,0 +1,151 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+"""These files are used by fsdp to help consolidate and shard optimizer states."""
```
myleott, line 5: love this

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -185,6 +186,8 @@ def __init__(
         self.buffer_dtype = buffer_dtype or self.compute_dtype
         self.move_grads_to_cpu = cpu_offload if move_grads_to_cpu is None else move_grads_to_cpu
         self.bucket_cap_mb = bucket_cap_mb
+
+        self.num_padded: List[int] = []
```
min-xu-ai, line 190: rename to numel_padded_per_param?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,83 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+    def _consolidate_optim_state_dict(
+        self, optim: torch.optim.Optimizer, recipient_rank: Optional[int] = None
+    ) -> List[Dict]:
+        """Update the consolidated state_dict list, one per rank.
+
+        Args:
+            recipient_rank (int): on which rank to materialize the full state dict.
+            None is a special value, which means that all ranks should have the state
+
+        .. warning: This needs to be called on all replicas"""
```
min-xu-ai, line 1365: add doc on return values?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,83 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+            raise NotImplementedError("optim state dict requires flatten_parameters=True")
+        world_optim_states = self._consolidate_optim_state_dict(optim, recipient_rank)
+        if self.rank != recipient_rank and recipient_rank is not None:
+            return None
+        # Unify the shard states by concatenating tensors and unflattening params
+        new_state_dict = ou.build_unflat_state_dict(self._fsdp_instances, world_optim_states)
+        return new_state_dict
+
+    @property
+    def _fsdp_instances(self) -> List[nn.Module]:
```
min-xu-ai, line 1409: nice

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,121 @@
...
+        if torch.is_tensor(v):
+            return v.numel()
+    return 0
+
+
+def assert_equal(a, b):
+    assert a == b, f"{a} != {b}"
+
+
+from time import time
```
min-xu-ai, line 32: move to top?

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,121 @@
+import functools
```
min-xu-ai, line 1: copyright?

min-xu-ai (commented):
Seems like super solid work. I added some minor comments. I didn't check the logic in detail mainly because I have two high level questions:

1. should we consider some optimizer wrapper that work together with fsdp to get the full state? It seems right now everything is in fsdp. Will an optimizer wrapper help more? I haven't thought through this.
2. I have been thinking that fsdp should support a "streaming" mode for full state so that no single rank's work need to hold all state (non-shard state) in memory. Should this PR try to do streaming to avoid overly big state?

Both 1 and 2 above are kind of independent of PR. Just wanted to put them out there in case they are helpful. If not, just let me know and I will dive deep into this version of the code and give it a more detailed review. Thanks!

fairscale/nn/data_parallel/fsdp_optim_utils.py
```diff
@@ -0,0 +1,151 @@
...
+import copy
+from typing import Dict, Generator, List, Tuple
+
+import torch
+
+
+# This function helps shard an
```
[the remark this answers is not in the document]
sshleifer: Done, nice catch

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -19,9 +19,10 @@
 from torch.nn import Parameter
 import torch.nn.functional as F
 
+import fairscale.nn.data_parallel.fsdp_optim_utils as ou
```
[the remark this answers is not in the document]
sshleifer: `SyntaxError: invalid syntax` :(

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,83 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
             raise ValueError(msg)
 
+    def _consolidate_optim_state_dict(
+        self, optim: torch.optim.Optimizer, recipient_rank: Optional[int] = None
+    ) -> List[Dict]:
+        """Update the consolidated state_dict list, one per rank.
+
+        Args:
+            recipient_rank (int): on which rank to materialize the full state dict.
+            None is a special value, which means that all ranks should have the state
```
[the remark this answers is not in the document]
sshleifer: Myle was confused by -1 so we went to `None`

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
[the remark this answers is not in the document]
min-xu-ai: got it. perhaps from . import fsdp_optim_utils as ou?

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+import functools
+from time import time
+
+from parameterized import parameterized
+import torch
+from torch.optim import SGD, Adadelta, Adam  # type: ignore
```
min-xu-ai, line 10: we usually don't have typing in test files. so "type: ignore" is not needed?

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+            fsdp = self.get_wrapped_model(group, config=config).cuda()
+            unwrapped_model = TransformerWithSharedParams(group).cuda()
+        else:
+            fsdp = FullyShardedDataParallel(NestedWrappedModule(group, wrapper_config=config), group, **config).cuda()
+            unwrapped_model = NestedWrappedModule(group, wrapper_config=None).cuda()
+
+        try:
+            fsdp_optim = optim_fn(fsdp.parameters(), lr=0.01,)
+            optim_unwrapped = optim_fn(unwrapped_model.parameters(), lr=0.01)
+        except TypeError:  # AdaScale
```
min-xu-ai, line 65: do you actually mean "AdaScale" here? I don't see AdaScale being used here in this test.

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        fsdp.module.run_backward(loss)
+        fsdp_optim.step()
+
+        output = unwrapped_model(*x)
+        loss = unwrapped_model.get_loss(x, output)
+        unwrapped_model.run_backward(loss)
+        optim_unwrapped.step()
+        unwrapped_sd = optim_unwrapped.state_dict()
+
+        # first_key = unwrapped_sd['state'][0].keys()
```
min-xu-ai, line 84: debugging statement?

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        tstart = time()
+        sd = fsdp.gather_full_optim_state_dict(fsdp_optim, recipient_rank=0)
+        duration = time() - tstart
+        # Switching from fairscale.optim.utils.broadcast_object to torch.broadcast_object_list will cause this to raise
+        assert duration < fsdp.world_size, f"gather optim state took {duration} seconds, suspect change in _consolidate"
```
min-xu-ai, line 89: this is interesting. thanks for the comment. what the usual value for `duration`? I am surprised that it is somehow connected with `world_size`, which is not in the unit of seconds even.

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+            assert objects_are_equal(shard_sd["state"][0], original_shard_sd["state"][0])
+        assert objects_are_equal(shard_sd["state"], original_shard_sd["state"])
+
+    def test_named_params_ordering(self):
+        """Test assumption of consolidate_optimizer_state_dict"""
+        group = DummyProcessGroup(0, 1)
+        model = TransformerWithSharedParams(group)
+        named_pars = [p for n, p in model.named_parameters()]
+        for i, p in enumerate(model.parameters()):
+            assert p.shape == named_pars[i].shape
```
min-xu-ai, line 123: why not p == named_pars[i]? they should be the same object by my quick test.

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        assert duration < fsdp.world_size, f"gather optim state took {duration} seconds, suspect change in _consolidate"
+
+        if fsdp.rank > 0:
+            return
+
+        assert_equal(len(sd["state"]), len(unwrapped_sd["state"]))
+        assert_equal(len(sd["param_groups"][0]["params"]), len(unwrapped_sd["param_groups"][0]["params"]))
+        assert_equal(
+            sum([first_tensor_shape(v) for k, v in sd["state"].items()]),
+            sum([first_tensor_shape(v) for k, v in unwrapped_sd["state"].items()]),
```
min-xu-ai, line 98: perhaps `norm` will be slightly better than `sum` for comparison in case both tensors sum to the same values? same with line 110, 111.

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+        shard_sd = fsdp.get_shard_from_optim_state_dict(sd)
+
+        original_shard_sd = fsdp_optim.state_dict()
+        assert_equal(len(shard_sd["state"]), len(original_shard_sd["state"]))
+        assert_equal(shard_sd.keys(), original_shard_sd.keys())
+        torch.save(shard_sd, f"new_shard_{fsdp.world_size}.pt")
```
min-xu-ai, line 106: can you please save this test file in /tmp with a temp name? otherwise, test may leave it behind in the git tree?

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        original_shard_sd = recursive_copy_to_device(original_shard_sd, non_blocking=False, device="cpu")
+
+        assert_equal(
+            sum([first_tensor_shape(v) for k, v in shard_sd["state"].items()]),
+            sum([first_tensor_shape(v) for k, v in original_shard_sd["state"].items()]),
+        )
+        if shard_sd["state"]:
+            assert objects_are_equal(shard_sd["state"][0], original_shard_sd["state"][0])
+        assert objects_are_equal(shard_sd["state"], original_shard_sd["state"])
```
min-xu-ai, line 115: is this test a superset of the test on line 114? why do we need both?

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
min-xu-ai: actually not seeing the file being used. could it be `/dev/null`?

min-xu-ai (commented):
Reviewed the test file first. It looks very good. Minor comments.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,87 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
                 traceback.print_stack()
             raise ValueError(msg)
 
+    def _consolidate_optim_state_dict(
+        self, optim: torch.optim.Optimizer, recipient_rank: Optional[int] = None
+    ) -> List[Dict]:
+        """Update the consolidated state_dict list, one per rank.
```
min-xu-ai, line 1359: should this be called only on the root FSDP instance?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,87 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+
+        .. warning: This needs to be called on all replicas"""
+        self._lazy_init()
+        # NOTE(SS): we do not support param groups yet, as they seem to break FSDP
+        # Pull the sharded state from all the other replicas
+        # Store all the states in order, rank by rank
+        should_collect_state = recipient_rank is None or (self.rank == recipient_rank)
+        all_states: List[Dict[str, Any]] = []
+        dummy_tensor = torch.tensor([0], dtype=torch.uint8, device=self.compute_device)
+        for rank in range(self.world_size):
```
min-xu-ai, line 1378: there might be complications here when nested FSDP instance have different world_size, right? For example, if BN layers are in their own world_size == 1 process groups, then we collect duplicated states for them? add a TODO?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,87 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+                assert isinstance(sd, dict), f"{self.rank} received {type(sd)} from {rank}, expected dict"
+                all_states.append(recursive_copy_to_device(sd, non_blocking=False, device=torch.device("cpu")))
+
+        return all_states
+
+    def gather_full_optim_state_dict(
+        self, optim: torch.optim.Optimizer, recipient_rank: Optional[int] = 0
+    ) -> Optional[Dict[str, Any]]:
+        """Return the last known global optimizer state. The returned state is compatible with Pytorch, in that the
+        sharded properties are not exposed. Multiple parameter groups are not yet supported.
```
min-xu-ai, line 1395:
```suggestion
        sharded properties are not exposed. Multiple parameter groups are not yet supported.
        
        This should be called only on the root FSDP instance.
        
        Different world_size groups in nested FSDP instances is not supported.
        Args:
                    optim (Optimizer): an optimizer instance for this FSDP rank. Its state is
                    used in the consolidation. However, its state is not modified.
                    recipient_rank (int): on which rank to materialize the full state dict.
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,87 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+        new_state_dict = ou.build_unflat_state_dict(self._fsdp_instances, world_optim_states)
+        return new_state_dict
+
+    @property
+    def _fsdp_instances(self) -> List[nn.Module]:
+        """Returns all fsdp modules in self.modules() including self."""
+        return [m for m in self.modules() if isinstance(m, FullyShardedDataParallel)]
+
+    def get_shard_from_optim_state_dict(self, full_optim_state_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """Get the portion of the optimizer state dict associated with the shard"""
```
min-xu-ai, line 1418:
```suggestion
        """Get the portion of the optimizer state dict associated with the shard
        
        This can be used to get the right sharded optimizer state to be loaded
        into the sharded optimizer for this FSDP rank.
        
        Args:
            full_optim_state_dict (dict): consolidated optimizer state returned by ``gather_full_optim_state``, or loaded from a checkpoint.
            
        Returns:
            (dict): a shard of the optimizer state.
        """
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1346,6 +1353,87 @@ def assert_state(self, state: Union[TrainingState, List[TrainingState]]) -> None
...
+        # Assert nesting is the same as it was at save time
+        instance_list = self._fsdp_instances
+        ou.check_param_counts_before_sharding(full_optim_state_dict, len(instance_list))
+        if self.flatten_parameters:
```
min-xu-ai, line 1422: does this assume all inner FSDP instances also have flatten == True?

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -122,15 +122,15 @@ def _flatten_params(self, flat_param: Optional[nn.Parameter] = None) -> None:
         # register the views as plain attributes
         self._unflatten_params_as_views()
 
-    def _get_param_views(self, flat_param: Tensor) -> Generator:
+    def get_param_views(self, flat_param: Tensor) -> Generator:
```
min-xu-ai, line 125: since this is becoming an public method, can you please:

1. add docstring with proper doc
2. assert flat_param is valid before using it?

min-xu-ai (commented):
Reviewed fsdp changes. I am not sure if nested FSDP cases are well supported by this change.

1. the APIs are really only intended for the root instance?
2. root and all inner instances should have flatten == True?
3. all instance needs to have world_size == default world_size?

If so, can you assert those are the cases in the APIs so that we don't accidentally produce incorrect optim states or crash with non-obvious errors?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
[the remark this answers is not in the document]
sshleifer: That works!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
sshleifer: Yes, more specifically it should be called on the instance that was the argument to `optimizer(model.parameters()`. Are there other cases?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
sshleifer: added TODO in the caller

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
sshleifer: Yes, will assert

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
sshleifer: I was getting "torch.optim has no Attribute Adadelta" from mypy without this, using 

```bash
mypy --ignore-missing-imports --scripts-are-modules --pretty .
```
from `fs_test`.

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
sshleifer: yes, nice catch

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
sshleifer: It takes longer to gather from 8 nodes than 4 than 2.
This actually takes 4 ms, but I accidentally regressed it during development and caused it to take 8 seconds for world size 2, 13 for world size 4. 
Now that it's fixed I want to prevent it happening again, agreed that the units are arbitrary.

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
sshleifer: This just checks that we have the same num elements as the base model after unflattening.
I renamed `first_tensor_shape` -> `first_tensor_numel` to make it clearer.

sshleifer (commented):
Thanks for the comments!

tests/nn/data_parallel/test_fsdp_optimizer_utils.py
min-xu-ai: I see. magic mypy. I thought it would skip the whole file since there isn't any type annotation in it.

fairscale/nn/data_parallel/fsdp_optim_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        consolidated_pid = param_id_map[expanded_pid]
+        for buffer_name, p in buffers.items():
+            if torch.is_tensor(p):
+                if buffer_name not in new_state[consolidated_pid]:
+                    new_state[consolidated_pid][buffer_name] = []
+                new_state[consolidated_pid][buffer_name].append(p.reshape(-1))
+            else:
+                constant_state[buffer_name] = p
+
+    for consolidated_pid, state in new_state.items():
```
min-xu-ai, line 34:
```suggestion
    # Now combine all tensors in each buffer using torch.cat().
    for consolidated_pid, state in new_state.items():
```

fairscale/nn/data_parallel/fsdp_optim_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        for buffer_name, tensors in state.items():
+            new_state[consolidated_pid][buffer_name] = torch.cat(tensors)
+        new_state[consolidated_pid].update(constant_state)
+    new_sd = {"state": new_state, "param_groups": sd["param_groups"]}
+
+    for pg_id, _ in enumerate(sd["param_groups"]):
```
min-xu-ai, line 40:
```suggestion
    # add pointers from the `params` dict.
    for pg_id, _ in enumerate(sd["param_groups"]):
```

fairscale/nn/data_parallel/fsdp_optim_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        f"there were {n_local_params_in_opt}"
+    )
+    stateless = len(full_optim_state_dict["state"]) == 0
+    assert stateless or (n_instances == n_local_params_in_opt), msg
+
+
+# All functions below here help saving the list of optimizer states, one from each rank
+# build_unflat_state_dict is the interface used by FSDP
+def _extract_constant_state(combined_state: Dict[int, Dict[str, List]], param_id: int) -> Dict:
+    constant_state = {}  # This state is like step in Adam, not a tensor so we dont unpad or cat it.
```
min-xu-ai, line 60:
```suggestion
    constant_state = {}  # This state is like the `step` count in Adam, not a tensor so we dont unpad or cat it.
```

min-xu-ai (approved):
Finished reviewing. Great step forward. I wished there were more comments in the fsdp_optim_utils.py for me follow along better. I tried my best and it seems to make sense. It might be able to be simplified and individually tested. But we can iterated on them later as we learn more.

sshleifer:
I'm gunna merge this tomorrow AM unless further comments or CI failure @myleott

PR #540, opened 2021-03-19
[feat][refactor][OSS] Param buckets + fp16 broadcasts

blefaudeux:
# Before submitting

## What does this PR do?
Similar to the grad buckets but with a different feature set (and small enough so that I felt that it was worth a minor duplication), add a class to represent the bucketed parameters, and make it easy to compress them to fp16 and subsequently decompress. ~20% speed boost with slow interconnect, no real accuracy loss with AMP (the FW is done with fp16 parameters anyway). Useless with fairseq for instance (parameters are already fp16 to begin with)

## PR review
cc @mannatsingh, a proper implementation of what talked about a couple of months back

## Did you have fun?

### DIFF
--- a/fairscale/nn/misc/grad_bucket.py
+++ b/fairscale/nn/misc/grad_bucket.py
@@ -16,6 +16,7 @@ class GradBucket:
     def __init__(self, size: int, dtype: torch.dtype, device: torch.device, destination: int) -> None:
         self._max_size = size
         self._params: List[torch.Tensor] = []
+        self._param_ids: List[int] = []
         self._fill = 0
         self._is_collapsed = False
 
@@ -39,9 +40,9 @@ class GradBucket:
         return len(self._params) == self.params_checked_in
 
     def can_add_grad_view(self, param: torch.Tensor) -> bool:
-        """ Is there enough room in the bucket to add this parameter gradient ?
+        """ Is there enough room in the bucket to add this parameter gradient, and is this param not already checked in ?
         """
-        return self._fill + param.numel() < self._max_size
+        return self._fill + param.numel() < self._max_size and id(param) not in self._param_ids
 
     def to(  # type: ignore
         self,
@@ -70,11 +71,15 @@ class GradBucket:
         """
         Add a new parameter gradient to the bucket. Param.grad becomes a view of this bucket buffer
         """
+
+        assert id(param) not in self._param_ids, "The same gradients cannot be checked in twice"
+
         if param.grad is None:
             param.grad = torch.zeros_like(param)
 
         self._add_grad_as_view(param)
         self._params.append(param)
+        self._param_ids.append(id(param))
 
     @torch.no_grad()
     def collapse(self) -> None:
--- a/fairscale/nn/misc/param_bucket.py
+++ b/fairscale/nn/misc/param_bucket.py
@@ -3,14 +3,14 @@
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import List, Optional, Union
+from typing import Any, Callable, List, Optional, Union
 
 import torch
 
 
-class ParamBucket:
+class Bucket:
     """
-    Helper class to simplify the handling of parameter buckets
+    Helper class to simplify the handling of buckets, which unify the underlying storage of multiple tensors
     """
 
     def __init__(self, size: int, dtype: torch.dtype, device: torch.device) -> None:
@@ -20,7 +20,6 @@ class ParamBucket:
 
         # The actual flat tensor
         self.buffer: torch.Tensor = torch.zeros(size, dtype=dtype, device=device)
-        self.sent = True
 
     def to(  # type: ignore
         self,
@@ -34,6 +33,27 @@ class ParamBucket:
         """
         assert self.buffer is not None, "Cannot move a collapsed bucket, please rebuild it"
         self.buffer.to(device, dtype, non_blocking)
+
+
+class ParamBucket(Bucket):
+    """
+    Helper class to simplify the handling of parameter buckets
+    """
+
+    def __init__(self, size: int, dtype: torch.dtype, device: torch.device) -> None:
+        super().__init__(size, dtype, device)
+
+    def to(  # type: ignore
+        self,
+        device: Optional[Union[int, torch.device]],
+        dtype: Optional[torch.dtype] = None,
+        non_blocking: bool = False,
+        keep_param_alignment: bool = True,
+    ) -> "ParamBucket":
+        """
+        Move the underlying buffer
+        """
+        super().to(device, dtype, non_blocking)
 
         if keep_param_alignment:
             self._reattach_params()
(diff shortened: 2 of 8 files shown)


fairscale/nn/misc/grad_bucket.py
```diff
@@ -16,6 +16,7 @@ class GradBucket:
     def __init__(self, size: int, dtype: torch.dtype, device: torch.device, destination: int) -> None:
         self._max_size = size
         self._params: List[torch.Tensor] = []
+        self._param_ids: List[int] = []
```
blefaudeux, line 19: small extension to fool proof these: forbid the same params being checked in twice

fairscale/nn/misc/grad_bucket.py
```diff
@@ -70,11 +71,15 @@ def add_grad(self, param: torch.Tensor) -> None:
         """
         Add a new parameter gradient to the bucket. Param.grad becomes a view of this bucket buffer
         """
+
+        assert id(param) not in self._param_ids, "The same gradients cannot be checked in twice"
```
min-xu-ai, line 75: when a module is forwarded multiple times, like the multi_res_input_forward in vissl, perhaps this will happen?

fairscale/optim/oss.py
```diff
@@ -52,6 +54,10 @@ class OSS(Optimizer):
             torch.distributed group (default: group.WORLD)
         broadcast_buffer_size (int):
             (deprecated) used to cap the size of the broadcast buffers, not being used anymore.
+        broadcast_fp16 (bool):
+            Compress the model shards in fp16 before sharing them in between ranks.
```
min-xu-ai, line 58: model shards or optimizer state shards?

min-xu-ai (commented):
I am wondering can grad bucket and param bucket share something? Should they have separate classes? To me, they are both tensors and both bucketing "horizontally", i.e. between params. Is there some opportunity to simplify the code here?

fairscale/nn/misc/grad_bucket.py
anj-s: Just curious if you have seen this happen. Can you have grads checked in twice?

fairscale/optim/oss.py
blefaudeux: model shards, really, not like in FDSP since each rank has all of them, but these are real model shards

fairscale/nn/misc/grad_bucket.py
blefaudeux: it's defined at construction time, not per forward, so this should be fine. I think that the assert is important because the grad becomes a view of a tensor, and it cannot be a (correct) view of two things at the same time

fairscale/nn/misc/param_bucket.py
```diff
@@ -0,0 +1,77 @@
...
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List, Optional, Union
+
+import torch
+
+
+class ParamBucket:
```
anj-s, line 11: Why not just a Bucket class? Seem pretty similar to GradBuckets?

blefaudeux:
> I am wondering can grad bucket and param bucket share something? Should they have separate classes? To me, they are both tensors and both bucketing "horizontally", i.e. between params. Is there some opportunity to simplify the code here?

I was wondering about that, the shared part is that there's an underlying flat buffer and that tensors are checked in. This being said they also have significant differences (grads can be None, not params, when checking in in the grad bucket the param is the reference but the grad data is checked in, vs. param all the way for the param bucket, collapsing/rebuilding is only needed for grad bucket, etc..). I think that both could inherit a base class and this would remove ~10 lines of code, not sure whether that's worth it but could be done in another PR if you'd like

fairscale/nn/misc/param_bucket.py
```diff
@@ -0,0 +1,77 @@
...
+        assert param.dtype == self.buffer.dtype
+        assert param.device == self.buffer.device
+
+        fill_next = self._fill + param.numel()
+        assert fill_next <= self.buffer.numel()
+
+        # Copy the current param value
+        if keep_existing_value:
+            self.buffer[self._fill : fill_next].copy_(param.data.flatten())
+        param.data = self.buffer[self._fill : fill_next].view_as(param.data)
```
anj-s, line 65: why do you need to do this?

fairscale/nn/misc/param_bucket.py
blefaudeux: see https://github.com/facebookresearch/fairscale/pull/540#issuecomment-803023686, there are differences still, I can make them inherit from the same base class if you'd like but don't expect huge line count savings. I'm being told that my PRs are too big, so probably better to keep that for another one ?

fairscale/optim/oss.py
```diff
@@ -52,6 +54,10 @@ class OSS(Optimizer):
...
+            This is safe to use when PyTorch AMP is activated. Without torch AMP this will lead to a slight
+            degradation in terms of accuracy.
```
anj-s, line 60: Should we disallow it in that case?

fairscale/nn/misc/param_bucket.py
blefaudeux: "this" being ? this precise line is the tensor view

fairscale/nn/misc/grad_bucket.py
blefaudeux: never say never :) Some models share weights in different places, so yes I think this could easily happen, I was trying to be cautious

tests/optim/test_oss.py
```diff
@@ -501,7 +501,7 @@ def run_test_reproducibility(rank, world_size, tempfile_name):
     loss_fn = torch.nn.L1Loss()
     loss_fn.to(device)
 
-    optimizer = optim.OSS(model.parameters(), optim=torch.optim.RMSprop, lr=0.1)
+    optimizer = optim.OSS(model.parameters(), optim=torch.optim.RMSprop, lr=0.1, broadcast_fp16=broadcast_fp16)
```
anj-s, line 504: Can we do a weight check as well? I think you have a test for that as well if i remember correctly.

tests/optim/test_oss.py
```diff
@@ -534,12 +534,13 @@ def closure():
...
 @skip_if_single_gpu
-def test_reproducibility():
+@pytest.mark.parametrize("broadcast_fp16", [False, True])
+def test_reproducibility(broadcast_fp16: bool):
     world_size = 2
     temp_file_name = tempfile.mkstemp()[1]
 
     mp.spawn(
-        run_test_reproducibility, args=(world_size, temp_file_name), nprocs=world_size, join=True,
+        run_test_reproducibility, args=(world_size, temp_file_name, broadcast_fp16), nprocs=world_size, join=True,
```
anj-s, line 543: Shall we add a benchmark for this? WDYT?

tests/optim/test_oss.py
blefaudeux: the bucket test is doing that, here I think that the comparison would be not completely trivial, because some accuracy is lost when doing fp32->fp16->fp32, so the result will not be strictly the same as not doing it. The bucket unit test does check that the results are the same as doing this conversion chain outside of the bucket

tests/optim/test_oss.py
blefaudeux: ideally it would be on fb cluster, it only really makes sense when inter node

blefaudeux:
@min-xu-ai @anj-s I can make both bucket inherit from a base class, it's true that they share some things (and for instance the type change would be handy for fp16 reduce support with buckets for ShardedDDP), this PR or another one ?

fairscale/nn/misc/param_bucket.py
anj-s: I was only thinking in terms of making changes in two places. If you end up doing that you can rethink merging this. For now this looks great! very clean and easy to understand.

fairscale/nn/misc/param_bucket.py
anj-s: From offline discussion: The params and buckets share the same storage but have different views.

tests/optim/test_oss.py
anj-s: discussed offline: Since the weights may not be exactly equal we want to know what the tolerance diff value could be.

tests/optim/test_oss.py
anj-s: from offline discussion since I was not clear why this only makes sense for inter node: Turns our the perf gain is insignificant due to fast intranode comms. We still want to measure performance to make sure there are no regressions.

fairscale/optim/oss.py
anj-s: Makes sense. Might be good to document the accuracy loss that you came across in your testing.

anj-s:
> @min-xu-ai @anj-s I can make both bucket inherit from a base class, it's true that they share some things (and for instance the type change would be handy for fp16 reduce support with buckets for ShardedDDP), this PR or another one ?

I always vote for new PRs so that we can quickly close and iterate on this one :) Upto you though, whatever you feel is best.

anj-s (approved):

blefaudeux:
@min-xu-ai @anj-s unified buckets (well, through inheritance), still ok for you ?

tests/optim/test_oss.py
blefaudeux: I can add that in another PR if that's ok ? I tested locally and not a big difference, if anything a bit slower with this option turned on, as expected

blefaudeux:
> @min-xu-ai @anj-s unified buckets (well, through inheritance), still ok for you ?

guess is yes :)

PR #543, opened 2021-03-20
[fix][FSDP] fix weight init when using apply() (fixes #490 and #444)

### LINKED ISSUE #444: [FSDP] improve robustness to mismatch between torch.cuda.current_device and model's device
## 🚀 Feature

Currently we require the user to set the current CUDA device appropriately when using FSDP, e.g.:
```python
torch.cuda.set_device(device)
model = FSDP(model, ...).cuda()
```

PyTorch DDP is more robust to this, see this example: https://github.com/facebookresearch/fairscale/issues/423#issuecomment-784633275

We currently use `.cuda()` and `torch.device("cuda")` in a few places, we can add a new `compute_device` arg and use that instead.

### LINKED ISSUE #490: [FSDP] issues around early init_weight
cc @prigoyal @myleott 

Documenting several issues around early init_weight.

By early, I mean in vissl's case, we do the following:

```
# building trunk
trunk = FSDP(...)
set_seed(distributed_rank)
trunk.init_weight()

# build model
model = FSDP(sequential(trunk, head))

```

two issues observed so far:

1. when init_weight is called, I used summon_full_params(), which called lazy_init. However, when root FSDP is later called during forward, it calls lazy_init again and try to set the is_root flag for children FSDP. There, we assert since children's FSDP is_root flags are already set.
2. vissl seems to init_weight at each rank differently. This means each worker starts with different weights? This is not standard DDP. Priya can you confirm? If so, FSDP might not be able to support that exact behavior.

myleott:
- add `compute_device` so that we can use `summon_full_params` immediately after `FSDP.__init__`, even if the params are still on CPU (this also fixes #444)
- override `FSDP.apply` so that it calls `summon_full_params` first. This makes it possible to do weight inits via `model.apply(custom_weight_init_fn)` without segfaulting, and should give identical results to not using FSDP

This also required reworking `_all_buffers_to` to no longer use `apply`, since it is called from within `_lazy_init` and created some circular logic: `apply -> summon_full_params -> _lazy_init -> _all_buffers_to -> apply`.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -279,7 +279,6 @@ class FullyShardedDataParallel(nn.Module):
                     module._reset_lazy_init()
         return return_value
 
-    @torch.no_grad()
     def _cast_buffers(
         self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None, memo: Optional[Set] = None
     ) -> None:
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -9,7 +9,7 @@ from enum import Enum, auto
 import functools
 from math import inf
 import traceback
-from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, List, NamedTuple, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, List, NamedTuple, Optional, Set, Tuple, Union
 
 import torch
 from torch.autograd import Variable
@@ -280,16 +280,39 @@ class FullyShardedDataParallel(nn.Module):
         return return_value
 
     @torch.no_grad()
-    def _all_buffers_to(self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None) -> None:
-        """Move all buffers to the specified device and dtype, recursively."""
+    def _cast_buffers(
+        self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None, memo: Optional[Set] = None
+    ) -> None:
+        """Move all buffers to the given *device* and *dtype*.
+
+        If *device* or *dtype* are not given, then they will default to
+        ``self.compute_device`` and ``self.buffer_dtype``, respectively. In the
+        case of nested FSDP instances, we will respect the child instance's
+        ``compute_device`` and ``buffer_dtype`` configuration.
+
+        Args:
+            device (torch.device, Optional):
+                device to cast buffers to (defaults to compute_device)
+            dtype (torch.dtype, Optional):
+                dtype to cast buffers to (defaults to buffer_dtype)
+            memo (Set, Optional):
+                set of modules that have already been processed
+        """
+        if memo is None:
+            memo = set()
         for module in self.modules():
-            for name, buf in module.named_buffers(recurse=False):
-                if buf is None:
-                    continue
-                buf = buf.to(device=device)
-                if torch.is_floating_point(buf):
-                    buf = buf.to(dtype=dtype)
-                setattr(module, name, buf)
+            if module is not self and isinstance(module, FullyShardedDataParallel):
+                # Allow any child FSDP instances to handle their own buffers.
+                module._cast_buffers(device=device, dtype=dtype, memo=memo)
+            elif module not in memo:
+                memo.add(module)
+                for name, buf in module.named_buffers(recurse=False):
+                    if buf is None:
+                        continue
+                    buf = buf.to(device=device or self.compute_device)
+                    if torch.is_floating_point(buf):
+                        buf = buf.to(dtype=dtype or self.buffer_dtype)
+                    setattr(module, name, buf)
 
     @property
     def params_with_grad(self) -> List[Parameter]:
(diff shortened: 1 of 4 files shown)


myleott:
Doesn't work with SyncBN, will fix

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -239,11 +252,44 @@ def __init__(
...
-        cast_fn = functools.partial(cast_buffers_, device=device, dtype=dtype)
-        self.apply(cast_fn)
+        for module in self.modules():
+            for name, buf in module.named_buffers(recurse=False):
+                if buf is None:
+                    continue
+                buf = buf.to(device=device)
+                if torch.is_floating_point(buf):
+                    buf = buf.to(dtype=dtype)
+                setattr(module, name, buf)
```
sshleifer: LGTM

tests/nn/data_parallel/test_fsdp_apply.py
```diff
@@ -0,0 +1,65 @@
...
+    TransformerWithSharedParams,
+    rename_test,
+    spawn_and_init,
+)
+
+
+class TestApply(DistributedTest):
+    @parameterized.expand(CONFIG_OPTIONS, name_func=rename_test)
+    def test_transformer_weight_init(self, config):
+        model_init_fn = functools.partial(apply_custom_weight_init, TransformerWithSharedParams)
```
sshleifer, line 25: This would have failed, right?

sshleifer (approved):
LGTM, nice tests!

tests/nn/data_parallel/test_fsdp_apply.py
```diff
@@ -0,0 +1,65 @@
...
+        spawn_and_init(test_fn)
+
+    @parameterized.expand(CONFIG_OPTIONS, name_func=rename_test)
+    def test_nested_wrapped_weight_init(self, config):
+        model_init_fn = functools.partial(apply_custom_weight_init, NestedWrappedModule)
+        test_fn = functools.partial(self._test_identical_outputs, model_init_fn, config, lr=0.01)
+        spawn_and_init(test_fn)
+
+
+def apply_custom_weight_init(model_init_fn, *args, **kwargs):
```
min-xu-ai, line 36: rename to `model_init_and_apply_custom_weight_init`?

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -77,6 +77,49 @@ def get_wrapped_model(group, cuda_first=False, config={}, **model_kwargs) -> Ful
             model = FullyShardedDataParallel(TransformerWithSharedParams(group, **model_kwargs), group, **config).cuda()
         return model
 
+    @classmethod
```
min-xu-ai, line 80: assume this is just code movement, no logical changes?

tests/ci_test_list_2.txt
```diff
@@ -33,3 +33,4 @@ tests/nn/pipe/test_deferred_batch_norm.py
 tests/nn/pipe/test_dependency.py
 tests/nn/pipe/test_stream.py
 tests/experimental/nn/test_multiprocess_pipe.py
+tests/nn/data_parallel/test_fsdp_apply.py
```
min-xu-ai, line 36: see below. I was thinking that this test file is in the wrong dir.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -239,11 +252,67 @@ def __init__(
...
+        Args:
+            fn (nn.Module): function to be applied to each submodule
+
+        Returns:
+            Module: self
+        """
+        is_uninitialized = self._is_root is None
+        with self.summon_full_params(recurse=False):
+            return_value = super().apply(fn)
+        # summon_full_params will call _lazy_init, which sets _is_root. However,
```
min-xu-ai, line 273: this is very nice! Thank you for fixing this in an elegant way!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -239,11 +252,67 @@ def __init__(
     def module(self) -> nn.Module:
         return self._fsdp_wrapped_module  # note: may be a FlattenParamsWrapper instance
 
+    def apply(self, fn: Callable[[nn.Module], None]) -> "FullyShardedDataParallel":
+        """
+        Applies ``fn`` recursively to every submodule (as returned by
+        ``.children()``) as well as self. Typical use includes initializing the
+        parameters of a model.
+
+        Compared to ``torch.nn.Module.apply``, this version additionally gathers
```
min-xu-ai, line 261: we should mention what training state should this be called in. I guess only idle state. It would be weird if this is called in the summon state, right?

min-xu-ai (approved):
Super nice. Some nonblocking comments.

min-xu-ai:
The test failures might be related to the cast_buffer change?

tests/nn/data_parallel/test_fsdp.py
myleott: yep, just moved it into DistributedTest to be reused in the other file

tests/nn/data_parallel/test_fsdp_apply.py
myleott: correct, in the case of flatten_parameters=True it would actually segfault previously, since the FlattenParamsWrapper views would point to empty storage

myleott:
> The test failures might be related to the cast_buffer change?

It's very weird, tests only seem to fail on 1.6, but pass in 1.7.1 and 1.8 (and on my local machine)... will dig a bit

PR #546, opened 2021-03-24
[chore][fix] SDP: yet another unit test improvement + bugfixes

blefaudeux:
# Before submitting

## What does this PR do?
Another fallout from #517, thanks again @zhengwy888 for your persistence..
In particular:
- the unit test was only really running AMP=true, not AMP=false, unfortunate broken refactoring
- the change graph unit test cannot really well compare with DDP, because this feature is actually not supported in DDP (thanks @mrshenli for the heads up). In DDP's case, the wrapper needs to be rebuilt if the trainability changes
- the manual reduce was not being checked against a grad norm (mistake), enabling that revealed that it was broken, now fixed and locked (cc @prigoyal, in case that's useful)

## PR review

## Did you have fun?


fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -303,10 +303,12 @@ def reduce(self) -> None:
         ), "No grads waiting to be reduced, maybe that this was called twice or there was no BW pass ?"
 
         # Trigger all the current BW hooks
-        _ = map(lambda x: x(), self._grad_accs)
+        self._bucket_flush_callback_set = True  # no need to flush in the end, we own the callback execution
+        _ = list(map(lambda x: x(), self._manual_reduce))
```
blefaudeux, line 307: list() was missing to the left, so the execution was always skipped. The unit test was not checking for the grad norm in that case, so this slipped in. Both are fixed (the callbacks are properly executed, and this is properly tested)

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -54,6 +54,7 @@ def run_ddp_parity(
     fp16_reduction,
     clip_grad_norm,
     amp,
+    manual_reduction,
```
blefaudeux, line 57: manual reduction becomes a case like the others, having it as an insider case switch proved not to be foolproof, keeping up with the test matrix approach

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -64,132 +65,134 @@ def run_ddp_parity(
...
-                    grad_accumulation,
-                    _manual_reduction=manual_reduction,
-                )
-
-            # Step/scale both
-            for _scaler, _closure, _optimizer in (
-                (ddp_scaler, ddp_closure, ddp_optimizer),
-                (sharded_scaler, sharded_closure, sharded_optimizer),
-            ):
-                if _scaler is not None:
```
blefaudeux, line 150: this meant that the non AMP case was not run..

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -64,132 +65,134 @@ def run_ddp_parity(
...
-            if i == 0 and change_train_graph:
-                next(sharded_ddp_model.parameters()).requires_grad = not next(
-                    sharded_ddp_model.parameters()
-                ).requires_grad
-                next(ddp_model.parameters()).requires_grad = not next(ddp_model.parameters()).requires_grad
-                check_same_model_params(sharded_ddp_model, ddp_model, f"Rank: {rank} - Trainability refresh {i} broke")
+            model.reduce()
 
-    # Test all combinations: AMP, Accumulate, Change train graph, reduce buckets
-    manual_reductions = [False, True] if not grad_accumulation and not change_train_graph else [False]
```
blefaudeux, line 180: screw that, not foolproof, now filling in the full test matrix except for clearly "skip" labelled cases which don't make sense

tests/nn/data_parallel/test_sharded_ddp_pytorch_parity.py
```diff
@@ -64,132 +65,134 @@ def run_ddp_parity(
...
+        # NOTE: DDP does not handle parameters trainability being changed after the fact, see
+        # https://github.com/pytorch/pytorch/blob/5781aec74ef00284e0262817a649278c2e8072bf/torch/nn/parallel/distributed.py#L471
+        if clip_grad_norm and not change_train_graph:
+            total_norm = torch.nn.utils.clip_grad_norm_(ddp_model.parameters(), 0.3, norm_type=2.0)  # type: ignore
+            if not torch.isnan(total_norm):
+                oss_total_norm = sharded_optimizer.clip_grad_norm(0.3, norm_type=2.0)
+                allclose = torch.allclose(oss_total_norm, total_norm, atol=1e-2 if amp else 1e-8)
+
+                if not allclose:
+                    # Debug helper if this unit test does not pass, compare the gradients in between DDP and ShardedDDP
```
blefaudeux, line 177: adding a debug printout in case this test breaks in the future, makes it easier to see if and where the reduction broke

zhengwy888:
thanks for the quick fix! do you think this warrants a synchronize call at the end of the last flush_reduce_calls(), similar to the manual reduce()?

min-xu-ai (approved):
great catch and test improvements.

blefaudeux:
> thanks for the quick fix! do you think this warrants a synchronize call at the end of the last flush_reduce_calls(), similar to the manual reduce()? 

Good question, I'm not sure, to be honest this change slipped in I'll remove it to be consistent. I still plan to land the torch future PR, this one should nail the sync issues without too many explicit barriers like that I think

PR #549, opened 2021-03-25
[test] FSDP: check with ddp parity with conv + bn

min-xu-ai:
- added DDP equivalency test
- added rmf, state_dict_norm functions to testing utils
- added more debugging output to objects_are_equal

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


min-xu-ai:
cc: @prigoyal 

Unfortunately, this hasn't turn up a bug in FSDP yet. I initially thought there is a bug in sync_bn but turns out to be a test bug not FSDP bug.

- Next, I am going to add more model structure in this test, try to mimic the model of regnet (with relu, loss function, etc.)
- Then, I think mixed precision grad scalar will be added to ensure there is no problem with that as well.
- since we will need apex sync BN, I will add apex sync BN in the test and make it supported too.

any other ideas to debug the convergence?

prigoyal:
thank you @min-xu-ai and such an awesome debugging. I am also curious if we add the conv-bn stack 2 times i.e. conv-bn-conv-bn then what happens in the test. Further I think we should repeat the same after adding `conv-bn-relu-conv-bn-relu` . 

On Apex SyncBN, I think we can discuss the priority once. I want to test the speedup that pytorch syncBN reports and see if we can just move to pytorch syncBN :)

min-xu-ai:
> thank you @min-xu-ai and such an awesome debugging. I am also curious if we add the conv-bn stack 2 times i.e. conv-bn-conv-bn then what happens in the test. Further I think we should repeat the same after adding `conv-bn-relu-conv-bn-relu` .

Yes, exactly. I am doing that in a separate branch as this is being reviewed.
 
> Also another thought: what happens if we simulate the world_size = 8 or 128 etc in the test. Do the syncBN test still pass? maybe there's some overflow or random values that get injected into FSDP (slicing, etc) -> just speculating few things we could test. not saying that this is happening already :)

Great idea. I will keep this on the todo list. Hopefully don't need to go there before convergence is done.
 
> On Apex SyncBN, I think we can discuss the priority once. I want to test the speedup that pytorch syncBN reports and see if we can just move to pytorch syncBN :)

Cool. Let's chat.

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -9,86 +9,216 @@
...
+        self.conv = Conv2d(2, 2, (1, 1))
+        self.bn = BatchNorm2d(2)
+
+    def forward(self, x):
+        x = self.conv(x)
+        x = self.bn(x)
+        return x
+
+
+# We get a bit fancy here. Since the scope is `module`, this is run only
```
blefaudeux, line 50: nice, I did not know this, thanks for the explanations

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -9,86 +9,216 @@
...
+    # Get reference inputs per rank.
+    world_size = 2
+    iterations = 100
+    inputs = [[]] * world_size
+    for rank in range(world_size):
+        for i in range(iterations):
+            inputs[rank].append(torch.rand(2, 2, 2, 2))
+
+    # Run DDP training twice, fp and mp.
+    for precision in ["full", "mixed"]:
```
blefaudeux, line 78: parameteriz-able ?

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -9,86 +9,216 @@
...
+        assert isinstance(fsdp_config, dict), str(fsdp_config)
 
     model = Model()
-    # Note, different rank may wrap in different order due to different random
-    # seeds. But results should be the same.
-    if random.randint(0, 1) == 0:
-        print("auto_wrap_bn, then convert_sync_batchnorm")
-        model = auto_wrap_bn(model)
+    model.load_state_dict(state_before)
+    model = model.cuda()
```
blefaudeux, line 149: probably me but I'm not seeing where the unit test is guarded for cuda ?

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: yes, it is done below for the FSDP test. Here, we loop over both cases in DDP reference so that we compute the trained weights from the same initial weights and inputs. We ensure the two results are different with an assert below.

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: I was learning it the first time too.

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: it is on line 200, `skip_if_single_gpu`.  I think it really meant single or no GPU. :-)

blefaudeux (approved):
LGTM, minor questions, overall I find this super clean ! Thanks Min

PR #556, opened 2021-03-29
[fix] disable single rank process group for auto_wrap_bn

min-xu-ai:
- beefed up unit test with regnet-like model
- found that single-rank process group is causing problem
- disabled it to enable convergence tests on the vissl side
- use `raise e from None` to get a better assertion output
  in testing.py.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1517,7 +1517,7 @@ def _pre_load_state_dict_hook(
 ########################################################################################
 
 
-def auto_wrap_bn(module: nn.Module) -> nn.Module:
+def auto_wrap_bn(module: nn.Module, single_rank_pg: bool = False) -> nn.Module:
     """
     Auto wrap all BatchNorm (BN) instances with a safer FSDP, esp. when convert
     to sync BN is used and the outer FSDP is flattening.
@@ -1531,6 +1531,9 @@ def auto_wrap_bn(module: nn.Module) -> nn.Module:
     Args:
         module (nn.Module):
             The model (or part of the model) in which BN to be pre-wrapped.
+        single_rank_pg (bool):
+            If true, put BNs in a single-rank process group. Default False.
+            This might be needed for Apex sync BN support. Still under construction.
 
     Returns:
         Processed module, where BNs are wrapped with a special FSDP instance.
@@ -1543,10 +1546,15 @@ def auto_wrap_bn(module: nn.Module) -> nn.Module:
         else:
             return is_bn and not isinstance(module, tuple(default_auto_wrap_policy.EXCLUDE_WRAP_MODULES))  # type: ignore
 
-    my_rank = dist.get_rank()
+    pg = None
+    if single_rank_pg:
+        # No sharding with this single member group.
+        my_rank = dist.get_rank()
+        pg = dist.new_group(ranks=[my_rank])
+
     fsdp_config = {
         "wrapper_cls": FullyShardedDataParallel,
-        "process_group": dist.new_group(ranks=[my_rank]),  # No sharding with this single member group.
+        "process_group": pg,
         "mixed_precision": False,  # Keep the weights in FP32.
         "flatten_parameters": False,  # Do not flatten.
     }
--- a/fairscale/utils/testing.py
+++ b/fairscale/utils/testing.py
@@ -445,7 +445,7 @@ def objects_are_equal(a: Any, b: Any, raise_exception: bool = False, dict_key: O
                     # Add dict key to the assertion error.
                     msg = e.args[0]
                     new_msg = f"For dict key '{dict_key}': {msg}"
-                    raise AssertionError(new_msg)
+                    raise AssertionError(new_msg) from None
                 else:
                     raise e
             else:
--- a/tests/nn/data_parallel/test_fsdp_regnet.py
+++ b/tests/nn/data_parallel/test_fsdp_regnet.py
@@ -16,7 +16,7 @@ import tempfile
 import pytest
 import torch
 import torch.multiprocessing as mp
-from torch.nn import BatchNorm2d, Conv2d, Module, SyncBatchNorm
+from torch.nn import AdaptiveAvgPool2d, BatchNorm2d, Conv2d, Linear, Module, ReLU, Sequential, Sigmoid, SyncBatchNorm
 from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.optim import SGD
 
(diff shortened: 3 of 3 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1543,10 +1546,15 @@ def wrap_bn_only_policy(module: nn.Module, recurse: bool, unwrapped_params: int)
         else:
             return is_bn and not isinstance(module, tuple(default_auto_wrap_policy.EXCLUDE_WRAP_MODULES))  # type: ignore
 
-    my_rank = dist.get_rank()
+    pg = None
+    if single_rank_pg:
```
min-xu-ai, line 1550: this is the main change. For now, I put this in an option and may need it later with apex BN or performance. This optional seems to be buggy in FSDP. Need more debugging. Enabling it also has impact on state_dict saving/loading as well as the optimizer state.

fairscale/utils/testing.py
```diff
@@ -445,7 +445,7 @@ def objects_are_equal(a: Any, b: Any, raise_exception: bool = False, dict_key: O
                     # Add dict key to the assertion error.
                     msg = e.args[0]
                     new_msg = f"For dict key '{dict_key}': {msg}"
-                    raise AssertionError(new_msg)
+                    raise AssertionError(new_msg) from None
```
min-xu-ai, line 448: this makes assertion output better, it doesn't print the first exception and then print "while handling the first exception the second exception is raised".

min-xu-ai:
wow, this is passing locally on my machine but failing on CI. converting to draft for more debugging.

prigoyal:
this is such an awesome work @min-xu-ai . 

In case helpful , just adding a little more context on how we use syncBN: in SyncBatchNorm (Apex or Pytorch), the process groups are indeed very important. People frequently change process groups (for example, for 128 gpus, we can use process groups of size 64 which leads to 2groups, or we can use process groups of size 8 leading to 16 groups etc.).

min-xu-ai:
> 
> 
> this is such an awesome work @min-xu-ai .
> 
> In case helpful , just adding a little more context on how we use syncBN: in SyncBatchNorm (Apex or Pytorch), the process groups are indeed very important. People frequently change process groups (for example, for 128 gpus, we can use process groups of size 64 which leads to 2groups, or we can use process groups of size 8 leading to 16 groups etc.).

Thanks, @prigoyal . This is helpful info. I think we want to support apex sync bn as well but it currently has more issues than pytorch BN.

The process group here is actually NOT the group of the pytorch sync BN group. The group here is FSDP sharding group. I used to have each rank's BN in their own single rank groups, which avoids sharding them. But that seems to have bugs. So I am undoing that here.

min-xu-ai:
I think the test is going to pass and this change should be ready to review after that. High level updates:

- only on CUDA 11 and up, I can make mixed precision DDP and FSDP the same
- even there, Linear layer can't have bias=True (could be a FSDP bug to be fixed)
- have to use pytorch AMP context for both DDP and FSDP
- have to use FP32 grad reduction
- have to use single rank group for BNs for both full and mixed precision modes (this could be a FSDP bug to be fixed)

fairscale/utils/testing.py
```diff
@@ -116,6 +117,30 @@ def torch_version() -> Tuple[int, ...]:
     return tuple(int(n) for n in numbering)
 
 
+_smi_ver = None
+
+
+def torch_cuda_version(compiled: bool = False) -> Tuple[int, ...]:
```
min-xu-ai, line 123: added a new function to either get torch compiled cuda version or Nvidia-smi reported cuda version.

tests/ci_test_list_2.txt
```diff
@@ -33,4 +33,8 @@ tests/nn/pipe/test_deferred_batch_norm.py
 tests/nn/pipe/test_dependency.py
 tests/nn/pipe/test_stream.py
 tests/experimental/nn/test_multiprocess_pipe.py
+tests/nn/moe/test_moe_layer.py
```
min-xu-ai, line 36: move a few tests around to balance the CI runtime.

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -143,39 +227,65 @@ def _test_func(
...
         assert isinstance(fsdp_config, dict), str(fsdp_config)
+        if fsdp_config["mixed_precision"]:
+            # To match DDP in AMP -O1, we need fp32 reduce scatter.
+            fsdp_config["fp32_reduce_scatter"] = True
 
     model = Model()
     model.load_state_dict(state_before)
     model = model.cuda()
 
+    class DummyScaler:
```
min-xu-ai, line 238: @anj-s I used this DummyScaler when scaling is not needed in full precision mode. Could be a useful thing to do in general to unify full and mixed precision training loop.

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -215,7 +334,10 @@ def test1(temp_files, ddp_ref, precision, flatten):
     fsdp_config["mixed_precision"] = precision == "mixed"
     fsdp_config["flatten_parameters"] = flatten == "flatten"
 
-    world_size = 2
+    if fsdp_config["mixed_precision"] and torch_cuda_version() < (11, 0):
```
min-xu-ai, line 337: @msbaines It is great that CI has a CUDA 11 VM. Otherwise, this test would have no place to run. It seems that we should gradually move everything to CUDA 11.

prigoyal:
really awesome work @min-xu-ai and amazing insights!! :)

> * only on CUDA 11 and up, I can make mixed precision DDP and FSDP the same
I am curious why this is the case. Is there a particular layer that leads to non-match when we use different CUDA. NOTE that if we mandate CUDA 11, we need to confirm that the machines in Fblearner will have CUDA 11 as we will plan to train the model there. :) 

> * even there, Linear layer can't have bias=True (could be a FSDP bug to be fixed)

this is indeed interesting. In general, not having bias can introduce regularization issue and exploding training. I would encourage that we debug this. 

> * have to use pytorch AMP context for both DDP and FSDP

Just double checking: we use the context for the autocasting only? I think this is awesome! pytorchAMP is mandated also by SDP so we can make a consistent switch.

> * have to use FP32 grad reduction

interesting: this can slow down speed seems like. Curious question: why do we require fp32 grad reduction? if this something we require for all layers or simply because of BN (BN params should be all kept in fp32).

> * have to use single rank group for BNs for both full and mixed precision modes (this could be a FSDP bug to be fixed)

interesting. This is essentially plain BN and not syncBN if we have to use single rank BN. Definitely sounds like bug somewhere.

Overall, seems like BN is introducing a lot of weirdness.

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -15,35 +15,124 @@
...
+        print(f"Using relu inplace: {_relu_inplace}")
+        # trunk
+        self.trunk = Sequential()
+        self.trunk.need_fsdp_wrap = True  # Set a flag for later wrapping.
+        stem = Sequential(Conv2d(2, 4, (3, 3), (2, 2), (1, 1), bias=False), BatchNorm2d(4), ReLU(_relu_inplace))
+        any_stage_block1_0 = ResBlock(4, 8)
+        self.trunk.add_module("stem", stem)
+        self.trunk.add_module("any_stage_block1", Sequential(any_stage_block1_0))
+
+        # head
```
anj-s, line 123: nit: maybe remove since the variable is named as head?

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -15,35 +15,124 @@
...
+        self.head = Sequential(
+            # TODO (Min): FSDP-mixed_precision doesn't compute the same ways as DDP AMP when bias=True.
+            #             so, we use bias=False for now in the projection_head.
+            #             The Conv2d layers above does not used bias in regnet, but even if they use
```
anj-s, line 127: s/used/use

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -15,35 +15,124 @@
...
+        self.trunk = Sequential()
+        self.trunk.need_fsdp_wrap = True  # Set a flag for later wrapping.
+        stem = Sequential(Conv2d(2, 4, (3, 3), (2, 2), (1, 1), bias=False), BatchNorm2d(4), ReLU(_relu_inplace))
+        any_stage_block1_0 = ResBlock(4, 8)
+        self.trunk.add_module("stem", stem)
+        self.trunk.add_module("any_stage_block1", Sequential(any_stage_block1_0))
+
+        # head
+        self.head = Sequential(
+            # TODO (Min): FSDP-mixed_precision doesn't compute the same ways as DDP AMP when bias=True.
```
anj-s, line 125: Thank you for the context with the comments! very helpful to grok :)

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -143,39 +235,67 @@ def _test_func(
...
+    loss_func = CrossEntropyLoss()
 
     for in_data in inputs[rank]:
         in_data = in_data.cuda()
         context = contextlib.suppress()
         if ddp and ddp_mixed_precision:
             in_data = in_data.half()
             context = torch.cuda.amp.autocast(enabled=True)
+        if not ddp and fsdp_config["mixed_precision"]:
+            context = torch.cuda.amp.autocast(enabled=True)
```
anj-s, line 291: just curious, when will this be true?

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -190,6 +310,15 @@ def _test_func(
         # Move tensors to CPU to compare numerics.
         for k, v in fsdp_state.items():
             fsdp_state[k] = v.cpu()
+        # Enable for debugging.
```
anj-s, line 313: nit: Should we remove this for long running tests?  I guess you are still debugging the issue so it makes sense to have it. only concern here is that it adds to the execution time.

anj-s (approved):

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: good point. will do.

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -15,35 +15,124 @@
...
+        self.trunk.need_fsdp_wrap = True  # Set a flag for later wrapping.
+        stem = Sequential(Conv2d(2, 4, (3, 3), (2, 2), (1, 1), bias=False), BatchNorm2d(4), ReLU(_relu_inplace))
+        any_stage_block1_0 = ResBlock(4, 8)
+        self.trunk.add_module("stem", stem)
+        self.trunk.add_module("any_stage_block1", Sequential(any_stage_block1_0))
+
+        # head
+        self.head = Sequential(
+            # TODO (Min): FSDP-mixed_precision doesn't compute the same ways as DDP AMP when bias=True.
+            #             so, we use bias=False for now in the projection_head.
```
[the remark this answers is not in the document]
min-xu-ai: yeah, it is not a tab, I was aligning it but I should do a 4-space indent. will fix.

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: this is when we use FSDP in mixed precision mode. So we need to turn on pytorch AMP.

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: it is disabled with "if False". I will clarify it.

PR #557, opened 2021-03-29
[offload] Audit OffloadModel API, add error messages and remove redundant code path.

anj-s:
# Before submitting

## What does this PR do?
This PR is part of the audit for finalizing the OffloadModel API.
- Moves functions to being private
- Removes a redundant code path through the ShardSyncLayer
- other smaller nits/error messages.

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/experimental/nn/offload.py
+++ b/fairscale/experimental/nn/offload.py
@@ -319,8 +319,11 @@ class OffloadModel(nn.Module):
         num_microbatches: int = 1,
     ):
         super().__init__()
-        assert model
-        assert device
+        if not model:
+            raise TypeError("`model` argument to `OffloadModel` cannot be None.")
+
+        if not device:
+            raise TypeError("`device` argument to `OffloadModel` cannot be None.")
 
         if not isinstance(model, nn.Sequential):
             raise TypeError("Module must be nn.Sequential to be partitioned by OffloadModel.")
--- a/fairscale/experimental/nn/offload.py
+++ b/fairscale/experimental/nn/offload.py
@@ -13,7 +13,7 @@ import torch
 from torch import nn
 
 
-def conditional_amp_fwd_decorator(orig_func):  # type: ignore
+def _conditional_amp_fwd_decorator(orig_func):  # type: ignore
 
     if hasattr(torch.cuda.amp, "custom_fwd"):
         return torch.cuda.amp.custom_fwd(orig_func)  # type: ignore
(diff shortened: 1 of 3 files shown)


fairscale/experimental/nn/offload.py
```diff
@@ -388,21 +311,28 @@ class OffloadModel(nn.Module):
...
         offload_device: torch.device = torch.device("cpu"),
-        num_slices: int = 5,
+        num_slices: int = 3,
         checkpoint_activation: bool = False,
         num_microbatches: int = 1,
     ):
         super().__init__()
-        # TODO(anj-s): Add error checks for cuda and sequential model.
+        assert model
+        assert device
```
min-xu-ai, line 323: if these are error checking, it is ideal not to use assert since python -O can bypass asserts. It is better to raise errors.

fairscale/experimental/nn/offload.py
```diff
@@ -432,27 +362,6 @@ def __init__(
...
-
-        # We don't move the last activation/output since the target is present
-        # on the device.
-        # TODO(anj-s): It is now a requirement that the target tensors be placed on the
-        # device.
-        result = self._activations[-1]
-        return result[0] if len(result) == 1 else result
+        # We need the send param to be a dummy input to enable the
+        # backward pass to be triggered for integer inputs.
+        return ActivationCheckpointing.apply(*inputs, torch.tensor([], requires_grad=True), self)
```
min-xu-ai, line 367: I am not totally understanding the magic here. Doesn't this end up calling self.forward() again, which will end up doing ActCkpt.apply() again?

min-xu-ai (commented):
LG, some questions.

fairscale/experimental/nn/offload.py
anj-s: This calls self.forward of the ActivationCheckpointing class and correspondingly the self.backward of the same class.

anj-s:
[codecov](https://codecov.io/gh/facebookresearch/fairscale)

fairscale/experimental/nn/offload.py
min-xu-ai: but checkpoint class's forward needs another forward function to work and the inputs, right? I thought the forward function as a parameter to the checkpoint class forward is "self", which is a callable. But that would end up calling this apply again, right?

blefaudeux: (flip a bool to mark in which scope you are ? sorry if obvious)

anj-s: Sorry I don't think I get the issue. I am implementing a custom autograd.Function with static forward and backward functions. I am not calling self() but instead self.layer_shard() so it should trigger the forward pass of the shard/module object.

min-xu-ai: I don't know there is an "issue" or not. The only question here is asking what's this code doing? It is not clear to me what is it doing, that's all. Maybe very detailed comments?

anj-s: Added comments though the details are present in the documentation for autograd.Function which I have linked to. I don't think it makes sense to explain the details about the [apply hook](https://pytorch.org/docs/stable/_modules/torch/autograd/function.html#Function).

min-xu-ai (approved):
LGTM

PR #567, opened 2021-04-01
replacing multip-process pipe implementation with more flexible one

mrzzd:
Initial implementation of proposal https://github.com/pytorch/pytorch/issues/55256


anj-s:
@mrzzd  can you add a short description of what this PR is doing? Thanks!

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
-
-    """
-
-    #: The number of micro-batches.
-    chunks: int = 1
-
-    #: The checkpoint mode to determine when to enable checkpointing. It is one
-    #: of ``'always'``, ``'except_last'``, or ``'never'``.
-    checkpoint: str = "never"
+class PipelineModule(nn.Module):
```
msbaines, line 49: Please add a docstring that explains the parameters and how to use the API.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
+                dist_record.wait_for(i)
+            self.fence(dist_record, i)
+            self.compute(dist_record, i)
+            self.forward_results(i, dist_record)
+
+    def fence(self, dist_record: DistributedPipelineRecord, chunk: int) -> None:
+        """Copies micro-batches after computation for the previous
+        micro-batches.
+        """
+        # Ensure that batches[chunk-1] is executed after batches[chunk] in
```
msbaines, line 233: Why is this necessary?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
+        # if deferred_batch_norm:
+        #    module = DeferredBatchNorm.convert_deferred_batch_norm(module, chunks)
+
+        self.partitions = _split_module(g)
+        self.input_feeds = [
+            next((i, fj, feed_idx) for i, (p, m) in enumerate(self.partitions) if p[0] == fi)
+            for fi, fj, feed_idx in g.model_input_users
+        ]
+
+        self._copy_streams: List[List[AbstractStream]] = []
```
msbaines, line 430: This is not used.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
+            for b, remote_ph_list in zip(batch.tensors, dist_record.forwarded_phony[chunk - 1]):
+                r = b
+                for remote_ph in remote_ph_list:
+                    ph = remote_ph.to_here()
+                    r = join(r, ph)
+                t.append(r)
+            dist_record.batches[chunk] = Batch(tuple(t), chunk)
+
+    def compute(self, dist_record: DistributedPipelineRecord, chunk: int) -> None:
+        """Runs tasks with synchronization to copy streams."""
```
msbaines, line 248: This comment looks wrong. I don't see any use of copy streams.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
+        dist_record = dist_record_ref.local_value()
+        self.run(dist_record)
+
+        if not dist_record.users:
+            result = microbatch.gather(dist_record.batches)
+            assert len(result) == 1
+            result = result[0]
+            s0 = current_stream(result.device)
+            if is_cuda(s0):
+                as_cuda(s0).synchronize()
```
msbaines, line 333: Why is this synchronize necessary? It will not be good for performance.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
+            return result
+
+        return None
+
+
+class MultiInputSequential(nn.Module):
```
msbaines, line 339: Please add a docstring here explaining how MultiInputSequential should work.

This module should also have some unittests that demonstrate that it works as expected.

msbaines (commented):
Looks good but some more comments would be nice. The code is quite complex and I'm not sure I understand the design. Perhaps a comment at the top explaining the design and a comment for each class and function would be super helpful.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,541 @@
...
 
-def _rcheckpoint(rmodule: rpc.RRef, input_rref: rpc.RRef) -> TensorOrTensors:
-    module = rmodule.local_value()
-    input = module[0](input_rref)  # calls _ToHere.forward
-    return checkpoint_sequential(module[1:], 1, input)
 
-
-def _parameter_rrefs(module: rpc.RRef) -> List[rpc.RRef]:
-    return [rpc.RRef(p) for p in module.local_value().parameters()]
+if torch.__version__.split("+")[0].split(".")[:3] <= ["1", "8", "1"]:
```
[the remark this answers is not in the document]
mrzzd: I agree, specially that with BOUNCE_TENSORS we have significant performance drop.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: To prevent queuing in backward path. This makes sure microbatch no. chunk-1 is not copied over to this device in backward path before gradient of microbatch no. chunk is computed.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Updated the comment. It is now synchronization to tensorpipe steams (with wait_event calls).

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Adding a TODO to revisit that. I believe this was needed, but needs investigation why.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Multi input feature is not covered by unit tests yet. I'll add unit tests in followup PR. But basic functionality of MultiInputSequential is already covered, e.g. in create_multiple_layers test where multiple layers are placed on same device.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,546 @@
...
+
+        return super().to(*args, **kwargs)
 
     def parameter_rrefs(self) -> List[rpc.RRef]:
-        rrefs_list_of_lists = [rpc.rpc_sync(l.owner(), _parameter_rrefs, args=(l,)) for l in self.rmodule]
-        return list(itertools.chain(*rrefs_list_of_lists))
+        remote_params = []
+        for p in self.partition_handlers:
+            remote_params.extend(p.rpc_sync().local_parameter_rrefs())
+        print("params rrefs: ", len(remote_params))
```
msbaines, line 491: Please avoid use of print. Use logging instead.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,546 @@
...
+                # TODO. Investigate why this is needed and remove it if possible.
+                as_cuda(s0).synchronize()
+            return result
+
+        return None
+
+
+class MultiInputSequential(nn.Module):
+    """
+        A variation of nn.Sequential, that allows the first module in the sequence accepts
```
msbaines, line 350: Why only the first module?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,546 @@
...
+            input = module(input)
+        return input
+
+
+def RemoteSequential(rref_list: List[rpc.RRef]) -> MultiInputSequential:
+    return MultiInputSequential(*(r.local_value() for r in rref_list))
+
+
+def _split_module(g: PipelineModulesGraph) -> List[Tuple[List[int], rpc.RRef]]:
+    g.compute_output_users()
```
msbaines, line 370: This seems like too large a scope over which to use a single-letter variable name. Single-letter variable names should only be used in list comprehensions and short loops. Use `graph` instead?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,546 @@
...
+    def replicate_output(self, module: PipelineModule, outputs: List[PipelineModule]) -> None:
+        mi = self.modules_list.index(module)
+        for m in outputs:
+            self.inputs[self._find_or_add(m)] = [(mi, 0)]
+
+    def compute_output_users(self) -> None:
+        m = len(self.modules_list)
+        self.output_users: List[List[Tuple[int, int, int]]] = [[] for _ in range(m)]
+        self.model_input_users = []
+        for i, inp in enumerate(self.inputs):
```
msbaines, line 142: s/inp/input/

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,546 @@
...
+        return self
+
+    def get_module_rref(self) -> rpc.RRef:
+        return self.module_rref
 
-    #: The checkpoint mode to determine when to enable checkpointing. It is one
-    #: of ``'always'``, ``'except_last'``, or ``'never'``.
-    checkpoint: str = "never"
 
+class PipelineModulesGraph(nn.Module):
```
msbaines, line 94: This class needs a doc string.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,546 @@
...
+
+    index = 0
+    for num_layers, remote_device in zip(balance, devices):
+        next_index = index + num_layers
+        for li in range(index, next_index):
+            layers[li].instantiate(remote_device)
+        index = next_index
+
+    g.add_sequence(layers)
+    g.feed_model_input(layers[0])
```
msbaines, line 546: I think the PipelineModulesGraph code would be clearer if add_sequence and feed_model_input were called directly from the PipelineModulesGraph constructor.

msbaines (commented):
Still working my way through this but it would be nice if each class had a doc string. It's not totally obvious what the purpose of each class is.

fairscale/nn/pipe/phony.py
```diff
@@ -55,7 +55,9 @@ def forward(ctx, input):
         phony = _phonies[key]
     except KeyError:
         with use_stream(default_stream(device)):
-            phony = torch.empty(0, device=device, requires_grad=requires_grad)
+            # Creating phony with size 1 instead of zero, since currently
+            # tensorpipe does not work with tensors of size zero.
+            phony = torch.empty(1, device=device, requires_grad=requires_grad)
```
msbaines, line 60: Why do empty tensors need to be sent?

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Please note, this class is only for internal use in this file. Currently the _split_module function packs consecutive layers into a single partition if they are on the same device, and some conditions hold, including that only the first of such layers may be a multi-input layer , and only last of them can be multi-output (i.e. it breaks the partition each time it seems a multi-input or multi-output layer).  Thus, in current implementation only the first layer could be multi-input.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Not sure if I understand the design in your mind. PipelineModulesGraph is a utility to create arbitrary acyclic graph. It is very hard to inject all necessary information as constructor argument.

fairscale/nn/pipe/phony.py
mrzzd: These phony's are used to inject dependencies in auto-grad engine between the tensors for batch i in partition j and and tensors for batch i+1 in partition j-1. So they need to be sent over to the remote partition j-1. This is to make sure in backward path, first  partition j-1 finishes processing batch i+1, and then  grads for the phony are transmitted back, and then the grads for batch i in partition j is sent over to partition j-1 (to prevent queuing in partition j-1).

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,680 @@
...
+        module_cls (nn.Module): Class for the module to be created remotely.
+        args (Sequence): args to be passed to ``module_cls``.
+        kwargs (Dict, optional): kwargs to be passed to ``module_cls``.
+        num_input (int): number of inputs to the forward function.
+        num_outputs: (int, optional): If the forward function returns a tuple, number of elements
+            in the tuple, otherwise it should be None
+        remote_device: (str, optional): Device on the destination worker where we‘d like to place
+            this module. The format should be "<workername>/<device>", where the device field can be
+            parsed as torch.device type. E.g., "trainer0/cpu", "trainer0", "ps0/cuda:0".
+            If this parameter can be provided later by calling the method instantiate
```
anj-s, line 58: is this sentence cut midway? i feel like something should come after instantiate.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,680 @@
...
-    very efficient.
-    ::
+class PipelineModule(nn.Module):
+    """Constructs a module on a remote device, possibly at a later time (in case the device is not
+    specified when creating PipelineModule.
+    Args:
+        module_cls (nn.Module): Class for the module to be created remotely.
+        args (Sequence): args to be passed to ``module_cls``.
+        kwargs (Dict, optional): kwargs to be passed to ``module_cls``.
+        num_input (int): number of inputs to the forward function.
```
anj-s, line 52: Is this optional or required? If it is required should we specify 1 below?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,680 @@
...
+            self.instantiate(remote_device)
+
+    @staticmethod
+    def _create_module(module_cls: Callable, args: Tuple, kwargs: Dict, device: str) -> nn.Module:
+        result: nn.Module = module_cls(*args, **kwargs)
+        result.to(device)
+        return result
+
+    def instantiate(self, remote_device: str) -> "PipelineModule":
+        on, device = remote_device.split("/")
```
anj-s, line 84: nit: Can we name this as "worker" or "remote_worker"?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,680 @@
...
+    def _find_or_add(self, module: PipelineModule) -> int:
+        try:
+            return self.modules_list.index(module)
+        except ValueError:
+            self._add_new_module()
+            self.modules_list.append(module)
+            return len(self.modules_list) - 1
+
+    def add_sequence(self, modules: List[PipelineModule], first_input: Optional[PipelineModule] = None) -> None:
+        """Adds a list of modules to the graph, to be run sequencially.
```
anj-s, line 119: typo: sequentially

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,680 @@
...
+    """A collection of remote modules (of type PipelineModule) with connections showing how inputs
+    to the model or outputs of individual modules are use as inputs of subsequent modules.
+    The graph has a number of helper functions that add new modules to the graph and define inputs
+    to these module.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.modules_list: List = []
+        self.inputs: List[Optional[List[Tuple[int, int]]]] = []
```
anj-s, line 104: Can you add a comment about what this tuple should contain?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,680 @@
...
+            return len(self.modules_list) - 1
+
+    def add_sequence(self, modules: List[PipelineModule], first_input: Optional[PipelineModule] = None) -> None:
+        """Adds a list of modules to the graph, to be run sequencially.
+        The connection between these modules is as follows: the first output of each of these modules
+        (except the last one) is used as the first input of its next module in this sequence.
+        The user also may optionally specify the input to the first module in this sequence with
+        argument 'first_input'. In this case the module 'first_input' must be added to the graph previously.
+        """
+        old_m = len(self.modules_list)
```
anj-s, line 125: nit: can we rename to "old_module_list_len" and "module_list_len" or something along the lines that indicate that we are working with units of length?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+    def fan_out(self, module: PipelineModule, outputs: List[PipelineModule]) -> None:
+        """Feeds outputs of a previously added module to modules specified by argument 'outputs' (so
+        'module' should have at least 'len(outputs)' outputs.
+        Modules in the list 'outputs' are added to the graph if they have not been added previously.
+        """
+        mi = self.modules_list.index(module)
+        for i, m in enumerate(outputs):
+            self.inputs[self._find_or_add(m)] = [(mi, i)]
+
+    def replicate_output(self, module: PipelineModule, outputs: List[PipelineModule]) -> None:
```
msbaines, line 158: Not used.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        """Feeds the first output of a previously added module to multiple modules specified by
+        argument 'outputs'. Modules in the list 'outputs' are added to the graph if they have not
+        been added previously.
+        """
+        mi = self.modules_list.index(module)
+        for m in outputs:
+            self.inputs[self._find_or_add(m)] = [(mi, 0)]
+
+    def validate_graph(self) -> None:
```
msbaines, line 167: Maybe remove this for now until it has an implementation.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        """Makes sure graph satisfies necessary requirements"""
+        # TODO: implement following checks:
+        #   * the graph has a least one module, and is connected.
+        #   * num_inputs and num_outputs for modules matche list of connections defined in the graph.
+        #   * all inputs to a module should come from model input, or modules with smaller index in
+        #     the graph. This condition is used in implementaion of DistributedPipeline.forward. Even
+        #     if we relax this condition, still need to make sure the graph is acyclic.
+        pass
+
+    def compute_output_users(self) -> None:
```
msbaines, line 177: Maybe rename this to `compile`. It was confusing to see a method named compute which doesn't return a value.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
-    very efficient.
-    ::
+class PipelineModule(nn.Module):
+    """Constructs a module on a remote device, possibly at a later time (in case the device is not
+    specified when creating PipelineModule.
+    Args:
+        module_cls (nn.Module): Class for the module to be created remotely.
+        args (Sequence): args to be passed to ``module_cls``.
+        kwargs (Dict, optional): kwargs to be passed to ``module_cls``.
+        num_input (int, optional): number of inputs to the forward function.
```
anj-s, line 52: s/num_input/num_inputs

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        self, device: torch.device, rank: int, chunks: int, num_input: int, users: List[Tuple[rpc.RRef, int, int]]
+    ) -> None:
+        self.ready_cv = Condition()
+        # Each chunk consists of num_input tensors. self.tensors stores these individual tensors.
+        self.tensors: List[List[Optional[Tensor]]] = [[None] * num_input for _ in range(chunks)]
+        # For each tensor in self.tensors, we record a cuda event in corrsponding tensorpipe stream in self.recv_events,
+        # and later the stream that processes that tensor will wait on that event.
+        self.recv_events = [[None] * num_input for _ in range(chunks)]
+        # Once all num_input tensors of a given chunk are recieved, they are assembled as a batch and stored in
+        # self.batches
```
anj-s, line 219: redundant comment. maybe remove or replace?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        batch = self.batches[chunk]
+        assert batch is not None
+        return batch
+
+
+class PartitionHandler:
+    """This class processes a single partition of the pipeline.
+    Args:
+        module_rref: RRef to the nn.Module for this partition. It should be on the local rpc worker.
+        device: The device that hols the module.
```
anj-s, line 265: s/hol/holds

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        num_input: Numer of inputs to the module
+        num_output: Number of outputs of the module. If the module output is not a tuple (and it is a
+            single tensor), num_output should be None.
+        rank: The rank of the partition
+        chunks: Number of micor-batches in a mini-batch
+        checkpoint_stop::
```
anj-s, line 271: missing doc string?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        Create one RRef for each parameter in the given local module, and return a
+        list of RRefs.
+        """
+        return [rpc.RRef(p) for p in self.module.parameters()]
+
+    def make_pipeline_record(self, users: List[Tuple[rpc.RRef, int, int]]) -> DistributedPipelineRecord:
+        return DistributedPipelineRecord(self.device, self.rank, self.chunks, self.num_input, users)
+
+    def run(self, pipeline_record: DistributedPipelineRecord) -> None:
+        """Runs pipeline parallelism.
```
anj-s, line 304: this can be a single line doc string.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+    def fence(self, pipeline_record: DistributedPipelineRecord, chunk: int) -> None:
+        """Prepares micro-batches for computation."""
+        # Ensure that batches[chunk-1] is executed after batches[chunk] in
+        # backpropagation by an explicit dependency.
+        # TODO: This dependency injection causes deadlock if this partition
+        # gets its input from model input. 1) Figure out why 2) If we need to live
+        # with this constraint, replace the condition 'pipeline_record.rank > 0' below with
+        # a more accurate one.
+        if chunk != 0 and pipeline_record.users and pipeline_record.rank > 0:
+            t = []
```
anj-s, line 330: nit: can we rename variables to be a bit more descriptive? It would be easier to understand the logic.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        if first_input is not None:
+            self.inputs[old_modules_len] = [(self.modules_list.index(first_input), 0)]
+
+    def feed_model_input(self, module: PipelineModule, ind: int = 0) -> None:
+        """Declares the input to a module as the input to the model. In case the model has multiple
+        inputs, the argument 'ind' indicates the index of the model input that is fed to the module.
+        """
+        self.inputs[self._find_or_add(module)] = [(-1, ind)]
+
+    def add_multi_input_layer(self, module: PipelineModule, inputs: List[PipelineModule]) -> None:
```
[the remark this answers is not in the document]
mrzzd: Added new test that covers this

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+        """Adds a module with multiple inputs to the graph. The modules that provide inpurs to this module
+        must be added previously to the graph and are listed with argument inputs.
+        """
+        self.inputs[self._find_or_add(module)] = [(self.modules_list.index(m), 0) for m in inputs]
+
+    def fan_out(self, module: PipelineModule, outputs: List[PipelineModule]) -> None:
```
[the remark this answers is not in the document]
mrzzd: Added new test that covers this

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: I feel it needs some comment here, to explain why we need both self.tensors and self.batches.

fairscale/nn/pipe/phony.py
mrzzd: It should be a no-op change for single process pipe. The tensor is never moved between devices, and no real operation is run on them.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+            checkpoint_stop = 0
+
+        exc_info: Optional[ExcInfo] = None
+
+        batch = pipeline_record.get_batch(chunk)
+
+        if pipeline_record is not None and pipeline_record.rank >= 0:
+            for e in pipeline_record.recv_events[chunk]:
+                if e is not None and is_cuda(self.stream):
+                    self.stream.wait_event(e)
```
msbaines, line 343: I think the code would be easier to reason about and more modular if this logic was pulled into PipelineRecord.get_batch.

To understand this snippet of code I have to return back PipelineRecord. Modularity and layering would greatly aid readability.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
-        model = nn.Sequential(a, b, c, d)
-        model = Pipe(model, balance=[1, 1, 1, 1], chunks=8)
-        output = model(input)
+def RemoteSequential(rref_list: List[rpc.RRef]) -> MultiInputSequential:
+    return MultiInputSequential(*(r.local_value() for r in rref_list))
 
-    .. _Pipe: https://arxiv.org/abs/1811.06965
 
-    Pipe combines pipeline parallelism with checkpointing to reduce peak
+def _split_module(graph: PipelineModulesGraph) -> List[Tuple[List[int], rpc.RRef]]:
```
msbaines, line 444: There are a lot of layering issues with this function. It is accessing many internal properties of PipelineModulesGraph. To understand requires going back to the PipelineModulesGraph. Should this be a method of PipelineModulesGraph. Or maybe some additional methods should be added to PipelineModulesGraph to that hide some of the details exposed here.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+
+    def run(self, pipeline_record: DistributedPipelineRecord) -> None:
+        """Runs pipeline parallelism. It modifies the given batches in place."""
+        m = len(pipeline_record.batches)
+
+        self.stream = current_stream(self.device)
+
+        for i in range(m):
+            with record_function("feed"):
+                pipeline_record.wait_for(i)
```
msbaines, line 303: It would be easier to understand the `wait_for`/`feed logic` if `feed()` was also called from within this function. For readability it is nice to have paired APIs balanced.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+        # with this constraint, replace the condition 'pipeline_record.rank > 0' below with
+        # a more accurate one.
+        if chunk != 0 and pipeline_record.users and pipeline_record.rank > 0:
+            dependant_tensors = []
+            batch = pipeline_record.batches[chunk]
+            assert batch is not None
+            for tensor, remote_ph_list in zip(batch.tensors, pipeline_record.forwarded_phony[chunk - 1]):
+                dependant = tensor
+                for remote_ph in remote_ph_list:
+                    ph = remote_ph.to_here()
```
msbaines, line 323: s/ph/phony/

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+
+            with use_device(self.device):
+                task.finalize(batch)
+
+            pipeline_record.batches[chunk] = batch
+
+        if exc_info is not None:
+            raise exc_info[0].with_traceback(exc_info[1], exc_info[2])
+
+    def forward_results(self, chunk: int, pipeline_record: DistributedPipelineRecord) -> None:
```
msbaines, line 396: The code would be easier to reason about if this were inlined in run. That way the `wait_for` and `feed` are balanced. Also, since this is a private method it should be prefixed with `_`.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+        """Declares the input to a module as the input to the model. In case the model has multiple
+        inputs, the argument 'ind' indicates the index of the model input that is fed to the module.
+        """
+        self.inputs[self._find_or_add(module)] = [(-1, ind)]
+
+    def add_multi_input_layer(self, module: PipelineModule, inputs: List[PipelineModule]) -> None:
+        """Adds a module with multiple inputs to the graph. The modules that provide inpurs to this module
+        must have been added previously to the graph and are listed with argument inputs.
+        """
+        self.inputs[self._find_or_add(module)] = [(self.modules_list.index(m), 0) for m in inputs]
```
msbaines, line 142: self.inputs[self._find_or_add(module)] is used in many places. Create a helper?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+            # stop adding more modules to the partition.
+            if len(graph.output_users[current_module_idx]) != 1:
+                break
+            if graph.modules_list[current_module_idx].num_outputs is not None:
+                break
+            # Next module to add is the only consumer of the ouput of the current module
+            next_module_idx = graph.output_users[current_module_idx][0][0]
+            next_module = graph.modules_list[next_module_idx]
+            # If the next module has multiple inputs, do not add it to the current partition and stop.
+            if graph.inputs[next_module_idx] != [(current_module_idx, 0)]:
```
msbaines, line 473: It would be nice for knowledge of graph.inputs to not be exposed outside the graph class.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+                # TODO: Debug why we need this special handling
+                if pipeline_record.owner().name == rpc.get_worker_info().name:  # type: ignore
+                    pipeline_record.local_value().feed(i, fj, b[feed_idx].value)
+                else:
+                    pipeline_record.rpc_async().feed(i, fj, b[feed_idx].value)  # type: ignore
+
+        return result
+
+
+def create_sequence_pipeline(
```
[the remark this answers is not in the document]
mrzzd: I added it as a helper function, not a new API. But what you say makes sense, removing it.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+    """
+
+    def __init__(
+        self,
+        module_cls: nn.Module,
+        args: Tuple,
+        kwargs: Optional[Dict] = None,
+        num_inputs: int = 1,
+        num_outputs: Optional[int] = None,
+        remote_device: str = None,
```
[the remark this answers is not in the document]
mrzzd: My feeling was that it's quite useful to be able to defer instantiating remote-modules. That allows to first define remote-modules, and then use helper utilities to find best placements. But if you feel it's not that useful we can drop this ability.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: I refactored this part to a separate method of PipelineRecord. get_batch is a light weight getter and is used elsewhere also.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+    to these module.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.modules_list: List = []
+        # self.inputs specifies inputs to each module in modules_list. Each input to each module
+        # is represented by a tuple (i, j). If i>=0, then the input is the j'th output of the i'th
+        # module. If i<0, the input is the j'th input to the model.
+        self.inputs: List[Optional[List[Tuple[int, int]]]] = []
```
anj-s, line 107: My suggestion is to use a python structure that holds 1) the module index 2) output of module index in 1). I think it will be more clear instead of have tuples with indices. WDYT?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+        """
+        old_modules_len = len(self.modules_list)
+        new_modules_len = len(modules)
+        self.modules_list.extend(modules)
+        # update inputs array
+        self.inputs.append([(self.modules_list.index(first_input), 0)] if first_input is not None else None)
+        for i in range(old_modules_len + 1, old_modules_len + new_modules_len):
+            self.inputs.append([(i - 1, 0)])
+
+    def feed_model_input(self, module: PipelineModule, ind: int = 0) -> None:
```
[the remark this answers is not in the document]
anj-s: +1 I thought something similar.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+        """Declares the input to a module as the input to the model. In case the model has multiple
+        inputs, the argument 'ind' indicates the index of the model input that is fed to the module.
+        """
+        self.inputs[self._find_or_add(module)] = [(-1, ind)]
+
+    def add_multi_input_layer(self, module: PipelineModule, inputs: List[PipelineModule]) -> None:
+        """Adds a module with multiple inputs to the graph. The modules that provide inpurs to this module
```
anj-s, line 139: s/inpurs/inputs

anj-s, line 139: s/provide/provides

fairscale/experimental/nn/multiprocess_pipe.py
anj-s, line 142: why is it always the 0th output?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+        """Feeds outputs of a previously added module to modules specified by argument 'outputs' (so
+        'module' should have at least 'len(outputs)' outputs.
+        Modules in the list 'outputs' are added to the graph if they have not been added previously.
+        """
+        mi = self.modules_list.index(module)
+        for i, m in enumerate(outputs):
+            self.inputs[self._find_or_add(m)] = [(mi, i)]
+
+    def compile_output_users(self) -> None:
+        """Precomputs self.model_input_users and self.output_users for internal use by the pipleine
```
anj-s, line 154: s/Precomuts/Precomputes

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,629 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+
+        # TODO: We need to make sure following conditions hold before preparing the graph for the pipeline:
+        #   * the graph has a least one module, and is connected.
+        #   * num_inputs and num_outputs for modules matche list of connections defined in the graph.
+        #   * all inputs to a module should come from model input, or modules with smaller index in
+        #     the graph. This condition is used in implementaion of DistributedPipeline.forward. Even
+        #     if we relax this condition, still need to make sure the graph is acyclic.
+
+        m = len(self.modules_list)
+        self.output_users: List[List[Tuple[int, int, int]]] = [[] for _ in range(m)]
```
anj-s, line 168: I would suggest that we create a Python structure to represent this. It's going to be hard to figure out what the three indices stand for down the line.

tests/experimental/nn/test_multiprocess_pipe.py
```diff
@@ -154,8 +156,59 @@ def update(devices):
     torch.random.manual_seed(3)
     criterion = DistributedLoss(torch.nn.MSELoss)
     x = torch.randn(8, 4).to(device)
-    model = [("linear1", nn.Linear, (4, 4), {}), ("relu", nn.ReLU, (), {})]
-    pipe = MultiProcessPipe(model, balance=[1, 1], chunks=4, devices=devices[:2])
+    model = [PipelineModule(nn.Linear, (4, 4), {}), PipelineModule(nn.ReLU, (), {})]
```
anj-s, line 159: Are we eventually thinking of creating helpers that can extract this info from a user's model? A user will have to rewrite their model to use this API

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Note, feed is called from remote worker, where as wait_for is called locally to wait until feed is called. Not sure if I understand your suggestion.
Are you suggesting I get rid of forward_results and move its logic to here?

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Yes, I have a PR coming next that has this (#591). Can it wait?

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: The API for now does not support other outputs. In next iterations we can extend the signature to accept something like this:
```py 
inputs: List[Union[PipelineModule, Tuple[PipelineModule, int]]]
```
to enable user to set output index if needed.

tests/experimental/nn/test_multiprocess_pipe.py
mrzzd: I am thinking of this API as a relatively low level API. We can have some sort of FX-tracing to extract the graph for the user. Or we can have a sharder utility that shards a layer and inserts them in the graph. But these ideas need a bit more work.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+    to these module.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.modules_list: List = []
+        # self.inputs specifies inputs to each module in modules_list. Each input to each module
+        # is represented by a tuple (i, j). If i>=0, then the input is the j'th output of the i'th
+        # module. If i<0, the input is the j'th input to the model.
+        self.inputs: List[Optional[List[Tuple[int, int]]]] = []
```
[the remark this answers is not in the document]
mrzzd: The current API allows to add modules and later specify inputs (e.g. in feed_model_input), so it's a bit hard to get rid of Optional. I'll get rid of Optional in a fllow up PR (#591)

fairscale/experimental/nn/multiprocess_pipe.py
anj-s: Sure! Are you planning to add a python structure for other places that use tuples of ints to connect modules?

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -3,204 +3,683 @@
...
+            return len(self.modules_list) - 1
+
+    def add_sequence(self, modules: List[PipelineModule], first_input: Optional[PipelineModule] = None) -> None:
+        """Adds a list of modules to the graph, to be run sequentially.
+        The connection between these modules is as follows: the first output of each of these modules
+        (except the last one) is used as the first input of its next module in this sequence.
+        The user also may optionally specify the input to the first module in this sequence with
+        argument 'first_input'. In this case the module 'first_input' must be added to the graph previously.
+        """
+        old_modules_len = len(self.modules_list)
```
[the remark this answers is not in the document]
mrzzd: You are right, we need to improve api here, so the user can specify first layer input directly here. Let's do that separately if you do not mind.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: Replaced other tuples (with more than two inputs) with NamedTuple.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,646 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+                        consumer.remote().feed(chunk, input_idx, v)
+                    )
+
+    def fence(self, pipeline_record: DistributedPipelineRecord, chunk: int) -> None:
+        """Prepares micro-batches for computation."""
+        # Ensure that batches[chunk-1] is executed after batches[chunk] in
+        # backpropagation by an explicit dependency.
+        # TODO: This dependency injection causes deadlock if this partition
+        # gets its input from model input. 1) Figure out why 2) If we need to live
+        # with this constraint, replace the condition 'pipeline_record.rank > 0' below with
```
msbaines, line 432: I believe the condition below is correct and matches the single-process code https://github.com/facebookresearch/fairscale/blob/cae9b638d6494fe916d0bdd308215b45473157f0/fairscale/nn/pipe/pipeline.py#L144

You do need a dependency for the first chunk or for the first partition since there is no queueing in either of these cases. The first partition becomes a sync in the backward pass and hence cannot queue.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,646 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+                pipeline_record.wait_for(chunk)
+            self.fence(pipeline_record, chunk)
+            self.compute(pipeline_record, chunk)
+            # Forward outputs of processing the chunk in this parition for processing by next partition.
+            with use_stream(self.stream):
+                for consumer, input_idx, output_idx in pipeline_record.consumers:
+                    v = pipeline_record.get_batch(chunk).value[output_idx]
+                    pipeline_record.forwarded_phony[chunk][output_idx].append(
+                        consumer.remote().feed(chunk, input_idx, v)
+                    )
```
msbaines, line 424: Should all of this logic maybe go into pipeline_record.

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +42,646 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+    def run(self, pipeline_record: DistributedPipelineRecord) -> None:
+        """Runs pipeline parallelism. It modifies the given batches in place."""
+        m = len(pipeline_record.batches)
+
+        self.stream = current_stream(self.device)
+
+        for chunk in range(m):
+            with record_function("feed"):
+                pipeline_record.wait_for(chunk)
+            self.fence(pipeline_record, chunk)
```
msbaines, line 416: Should fence be a method on pipeline record. It feels like there are too many internals of pipeline_record being exposed here. Should PartitionHandler and PipelineRecord be one single class? Or should more of the logic in this class be pushed to PipelineRecord so that it is hiding more details and has a simpler interface.

fairscale/experimental/nn/multiprocess_pipe.py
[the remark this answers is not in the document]
mrzzd: It's gone in followup PR (#591 )

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: IIUC this dependency is to prevent queuing grads coming from second partition to the first partition. So if backward path in the first partition is slow, it is possible to we get OOM. To be honest I also had problem understanding the condition in single-process code.
But even if that's the case, it resolves "comment 1" in my comment. The second item in the comment is still valid. the condition here is not simply "rank != 0". Since with the graph some of other ranks could also have the same situation (if their input comes from model input).

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +43,658 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+        #   * the graph has a least one module, and is connected.
+        #   * num_inputs and num_outputs for modules matche list of connections defined in the graph.
+        #   * all inputs to a module should come from model input, or modules with smaller index in
+        #     the graph. This condition is used in implementaion of DistributedPipeline.forward. Even
+        #     if we relax this condition, still need to make sure the graph is acyclic.
+
+        m = len(self.modules_list)
+        self.output_consumers: List[List[PipelineModulesGraph.OutputConsumerIndex]] = [[] for _ in range(m)]
+        self.model_input_consumers = []
+        for i, input in enumerate(self.inputs):
```
msbaines, line 207:
```suggestion
        for model_index, module_inputs in enumerate(self.inputs):
```

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +43,658 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+    def partition_graph(self) -> List[Tuple[List[int], rpc.RRef]]:
+        """Splits the graph into pipeline partitions and for each parition returns a tuple (indices, module_rref),
+        where indices is indices of modules of the partition in the graph, and module_rref is an RRef to an nn.Module:
+        Each partition is a list of modules on the same device that are executed sequentially (output of each module is
+        the input to the next module).
+        If there is only one module in the partition, module_rref is reference to that module; otherwise those modules
+        are wrapped by a MultiInputSequential and module_rref referes to that.
+        """
+        self._compile()
+        module_used = [False] * len(self.modules_list)
```
msbaines, line 259: I think `set` would be more pythonic and easier to read. It would also remove the need for module_idx entirely.
```suggestion
        modules_used = set()
```

fairscale/experimental/nn/multiprocess_pipe.py
```diff
@@ -85,122 +43,658 @@ def dloss(input_rref: rpc.RRef, target_rref: rpc.RRef) -> rpc.RRef:
...
+            assert input is not None
+            for j, input_item in enumerate(input):
+                if not self._is_model_input(input_item):
+                    self.output_consumers[input_item[0]].append(
+                        PipelineModulesGraph.OutputConsumerIndex(i, j, input_item[1])
+                    )
+                else:
+                    self.model_input_consumers.append((i, j, input_item[1]))
+
+    def _trace_modules(self, module_idx: int) -> List[int]:
```
msbaines, line 217: It feels like the code would be much simpler if we just used the modules directly instead of adding the indirection of index tables.

msbaines (commented):
Code is looking much better. I think the code would be a lot easier to read if the index tables were removed entirely and the modules were used directly and sets, lists and dictionaries of modules used where appropriate.

fairscale/experimental/nn/multiprocess_pipe.py
mrzzd: It is not so easy to get rid of indexes. For example self.output_consumers is indexed by module index, and we store indexes of modules in self.inputs and in output_consumers. We change all that to module object ids, but let's do that as a separate PR, since it is not going to be a small change (if we think that kind of change helps).

msbaines: My primary concern is readability. I still find it very challenging to read through the code. For me, the i and j variables and arrays of objects are what make the code most challenging. If they could be hidden behind a lower-level API or removed entirely, it would make the code much easier to read and understand. Each iteration of review is still very challenging. I would like to see the code get a stage where it flows naturally when reading.

mrzzd: I see, so I merge other changes in #591 to this branch, and do a major revision to reduce from complexity

mrzzd: I went over all data structures used in the PR changed them to help readability. Also splitted the code to smaller modules. Please have another look.
I suggest you first have a look at #591 (which is based on changes that you have already reviewed), because in this PR the files are now splitted and it's hard to understand how they have changed.

msbaines (approved):
Love it! This is so much easier to read.

fairscale/experimental/nn/distributed_pipeline/graph.py
```diff
@@ -0,0 +1,202 @@
...
+                else:
+                    self.model_input_consumers.append(data_consumer)
+
+    def _trace_modules(self, node: Node) -> List[Node]:
+        """Compiles a list of modules (starting from module number module_idx), where each module in the list
+        gets the output of previous module in the list as its input. So every module in the list, except the
+        first one should have only one input, and similarly, every module in the list, except the last one
+        should have only one output.
+        """
+        partition = []
```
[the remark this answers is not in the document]
mrzzd: That's because the way each partition is going to work. It waits untill all required inputs are ready and then it runs the partition. This may endup in a deadlock if partitions can have arbitrary connections to each other. 
For example, consider modules m1, m2, m3, where m1 and m3 are on device0 and m2 is  on device1, and iinput connections are as fllow:
m1 --> m2
m1 --> m3
m2 --> m3
 If we create have m1 and m3 on the same partition and m2 on another partition, there will be a deadlock.

PR #579, opened 2021-04-05
[FSDP] [feat] Add state_dict_device option

myleott:
This option is helpful for fairseq to avoid duplicating the full model params on GPU when calling state_dict. It typically makes sense to do `state_dict_device=torch.device("cpu")`.

I also moved the state_dict related tests to a new test_fsdp_state_dict.py file. This is self-contained in the first comment.


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1501,6 +1501,15 @@ def get_shard_from_optim_state_dict(self, full_optim_state_dict: Dict[str, Any])
         return full_optim_state_dict
 
 
+def _get_default_compute_device(module: nn.Module) -> torch.device:
```
sshleifer, line 1504: By my reading, this will never return `cpu`, so should be called `_get_default_cuda_device`

sshleifer (approved):

tests/nn/data_parallel/test_fsdp_state_dict.py
```diff
@@ -0,0 +1,241 @@
...
+            if config["cpu_offload"] or (sd_device is not None and sd_device.type == "cpu"):
+                assert v.device.type == "cpu", v.device.type
+            else:
+                assert v.device.type == "cuda", v.device.type
+
+        expected_dtype = torch.float16 if pure_fp16 else torch.float32
+        for k, v in sd.items():
+            if not torch.is_floating_point(v):
+                continue
+            assert v.dtype == expected_dtype, f"{v.dtype} != {expected_dtype}"
```
myleott, line 237: for future reference, this is the new test

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1534,12 +1543,14 @@ def alloc_storage_(data: torch.Tensor, size: torch.Size) -> None:
...
-        # We copy the state_dict since full param will be freed after
-        # we exit the summon_full_params() context.
-        for key in state_dict.keys():
+    for key in state_dict.keys():
+        if state_dict[key].device.type != module.state_dict_device.type:
+            state_dict[key] = state_dict[key].to(device=module.state_dict_device)
+        elif module.training_state == TrainingState.SUMMON_FULL_PARAMS:
+            # We copy the state_dict since full param will be freed after
+            # we exit the summon_full_params() context.
             state_dict[key] = state_dict[key].clone()
```
shruti-bh, line 1554: I am a bit hazy on when after calling model.state_dict(), is the state dict is moved to CPU?
What happens when the full state dict is summoned on rank 0 (for example), the accumulation of the full state dict still happens on the GPU associated with rank 0, right?

myleott: This is a hook that's called after each FSDP instance's state_dict method is called. So if we have a nested FSDP wrapper around an expert layer, we'll call state_dict for that layer, which will summon the params on GPU, then this hook will copy it to CPU, and then finally the params should be freed.

This happens recursively for all FSDP instances, but there's actually a bug here (which I'm working on fixing), since this state_dict is actually built up incrementally so we have something like `n**2` calls to `clone()`, for `n` nested FSDP instances.

myleott: > there's actually a bug here (which I'm working on fixing)

This is fixed now

min-xu-ai (approved):
this is awesome!

PR #583, opened 2021-04-05
[fix][OSS] Mixed types & proper buckets + no grad_fn is fine

blefaudeux:
# Before submitting

- [~x] Did you write any new necessary tests?

## What does this PR do?
Fixes https://fb.workplace.com/groups/pytorchLightning/permalink/826203544631599/. I tried for some time but could not find a case in the unit test which would trigger this (inc. with pytorch 1.9 nightly), I've asked the users for repro examples and will update if something comes

Fixes another bug pre-emptively (no report yet), where the buckets could be of the wrong type if these were mixed and bad luck. Adds a unit test for this one

## PR review

## Did you have fun?

cc @ananthsub

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -494,18 +494,24 @@ class ShardedDataParallel(nn.Module):
             if param.grad is not None and param.grad.requires_grad:
                 raise RuntimeError("ShardedDataParallel only works with gradients that don't require grad")
 
-            # Register the hook to the next function in line,
-            # so that the hook is fired when this grad has properly been computed
             p_tmp = param.expand_as(param)
-            assert p_tmp.grad_fn is not None
-            grad_acc = p_tmp.grad_fn.next_functions[0][0]
-            dst_rank = self._trainable_param_to_rank[param]
 
-            reduce_function = self._get_reduce_fn(index, param, dst_rank)
+            # See https://pytorch.org/docs/stable/tensors.html?highlight=grad_fn
+            # We're interested in the tensors which will be tracked by Autograd
+            # Some tensors can have gradients independent of the inputs (ie. pooling layer for instance),
+            # these do not need to be sync'ed
+            if p_tmp.grad_fn is not None:
+                # Register the hook to the next function in line,
+                # so that the hook is fired when this grad has properly been computed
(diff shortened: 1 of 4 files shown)


fairscale/optim/oss.py
```diff
@@ -591,13 +591,14 @@ def _setup_flat_buffers(self) -> None:
 
                     # Merge all the trainable params in a single bucket
                     trainable_params = list(filter(lambda x: x.requires_grad, params))
-                    buffer_size = sum(map(lambda x: x.numel(), trainable_params))
-                    bucket = ParamBucket(size=buffer_size, dtype=params[0].dtype, device=device)
+                    if trainable_params:
+                        buffer_size = sum(map(lambda x: x.numel(), trainable_params))
+                        bucket = ParamBucket(size=buffer_size, dtype=trainable_params[0].dtype, device=device)
```
blefaudeux, line 596: there was a typo here, which could change everything: pick the first param type vs. first trainable param. The trainable params are the ones which go in a bucket, so you could have a config (see added unit test) which could break that, by adding another param of another type which was not trainable, and have it come first in the list

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -494,18 +494,24 @@ def _setup_backward_hooks(self) -> None:
...
-            reduce_function = self._get_reduce_fn(index, param, dst_rank)
 
-            self._grad_hooks.append(grad_acc.register_hook(reduce_function))
-            self._grad_accs.append(grad_acc)  # keep this hook in scope
-            self._manual_reduce.append(reduce_function)
+            # See https://pytorch.org/docs/stable/tensors.html?highlight=grad_fn
+            # We're interested in the tensors which will be tracked by Autograd
+            # Some tensors can have gradients independent of the inputs (ie. pooling layer for instance),
+            # these do not need to be sync'ed
+            if p_tmp.grad_fn is not None:
```
blefaudeux, line 503: the assert here was too strong, we can just ignore the params whose gradients don't depend on the inputs (see link), by definition they don't need to be synced since all the ranks will have the same. I could not repro an issue here though (even with 1.9 nightly), but several users faced the issue with a recent pytorch, so I'm  guessing that pytorch is getting more aggressive at pruning .grad_fn when not required

ananthsub (approved):

anj-s (approved):

PR #587, opened 2021-04-07
[fix][ShardedDDP] Properly handle .eval() mode

### LINKED ISSUE #586: [ShardedDDP] Handle transition to eval + parameter change
## 🐛 Bug
Reported by @SeanNaren, after a successful training the switch to eval() is not properly taken into account, and the grads are marked as "waiting to be reduced" while they should not (we're in eval..)

blefaudeux:
# Before submitting

## What does this PR do?
Fixes eval mode not being properly handled by ShardedDDP, at best it kept all the grad hooks in place, at worst it could crash when there was a trainability change which should have been ignored. Add a unit test to catch that

Closes #586 
cc @SeanNaren  @ananthsub

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -635,7 +635,9 @@ class ShardedDataParallel(nn.Module):
         trainability_changed |= not self.training and len(self._grad_hooks) > 0
 
         if trainability_changed:
-            logging.warning("ShardedDDP detected that the trainable params changed, updating the partitioning")
+            logging.warning(
+                "ShardedDDP detected that the trainable params changed, either because of eval/train mode or parameter freezing/unfreeze."
+            )
             self._reference_trainable_mask = trainable_mask
 
         return trainability_changed
--- a/fairscale/nn/data_parallel/sharded_ddp.py
+++ b/fairscale/nn/data_parallel/sharded_ddp.py
@@ -198,15 +198,10 @@ class ShardedDataParallel(nn.Module):
         """
 
         # Deferred initialization, or change detection
-        needs_setup = len(self._grad_hooks) == 0
+        needs_setup = len(self._grad_hooks) == 0 and self.training
 
         if self.auto_refresh_trainable:
-            # Optionally check whether the trainable parameters have changed
-            trainable_mask = list(map(_trainable, self._all_params))
-            if trainable_mask != self._reference_trainable_mask:
-                logging.warning("ShardedDDP detected that the trainable params changed, updating the partitioning")
-                needs_setup = True
-                self._reference_trainable_mask = trainable_mask
+            needs_setup |= self._detect_train_change()
 
         if needs_setup:
             self.refresh_trainable()
@@ -278,7 +273,6 @@ class ShardedDataParallel(nn.Module):
 
         self._trainable_params = list(filter(lambda x: x.requires_grad, self._all_params))
         self._trainable_params.sort(key=lambda x: x.numel())
-        self._grad_to_be_reduced = [True for _ in self._trainable_params]
 
         self._trainable_param_to_rank = {}
         for optim in self.sharded_optimizers:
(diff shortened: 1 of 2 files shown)


tests/nn/data_parallel/test_sharded_ddp_features.py
```diff
@@ -262,6 +262,29 @@ def test_mixed_types():
...
+
+    model = _get_mlp()
+    model.train()
+    optimizer = OSS(params=model.parameters(), optim=torch.optim.SGD, lr=1e-3, momentum=0.99)
+    model = ShardedDataParallel(model, optimizer)
+    input_tensor = torch.rand((2, 2))
+    loss = model(input_tensor).sum()
+    loss.backward()  # make sure that the gradients are reduced
+
+    model = model.eval()
```
anj-s, line 277: Is there something else we should check? e.g that the weights haven't changed. This only checks that we don't throw an error.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -490,6 +486,9 @@ def _setup_backward_hooks(self) -> None:
         # Go through the parameters, attach the hook
         self._grad_accs = []
         self._manual_reduce = []
+        if not self.training:
```
anj-s, line 489: do we need to remove existing hooks in eval model? Just curious, otherwise we could move this to the top of the function.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -624,3 +623,19 @@ def _flush_reduce_calls(self) -> None:
                 bucket.sent = True
 
         self._consume_work_handles()
+
+    def _detect_train_change(self) -> bool:
```
anj-s, line 627: Nice!

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -624,3 +623,19 @@ def _flush_reduce_calls(self) -> None:
...
+        # Optionally check whether the trainable parameters have changed
+        trainable_mask = list(map(_trainable, self._all_params))
+
+        # - one or more parameters trainability changed
+        trainability_changed = trainable_mask != self._reference_trainable_mask
+
+        # - the whole model is not trainable but we still have grad hooks
+        trainability_changed |= not self.training and len(self._grad_hooks) > 0
```
anj-s, line 635: does this mean that grad_hooks should be greater than 0 in eval model? Not sure I understand why this should be the case.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -624,3 +623,19 @@ def _flush_reduce_calls(self) -> None:
...
+
+        if trainability_changed:
+            logging.warning("ShardedDDP detected that the trainable params changed, updating the partitioning")
```
anj-s, line 638: update logging statement since "partitioning" is not clear in this function.

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: I thought that it was better for correctness, in that if there's a .backward() left somewhere it still respects the eval() setting ? The documentation is not super clear, to me at least https://pytorch.org/docs/stable/generated/torch.nn.Module.html?highlight=eval#torch.nn.Module.train

fairscale/nn/data_parallel/sharded_ddp.py
blefaudeux: it was meant to detect that the trainability changed, ie. we're in eval() mode but there are grad_hooks in place so we should refresh ? it's tied to the question above, I'm not sure of the reference behavior here

anj-s: From my offline conversation with @blefaudeux to understand this better:
* We can't detect when a module switches from train->eval unless we use the presence of hooks as an indicator. 
* We refresh trainable 1) at the beginning 2) when params changes their requires_grad property 3) train<->eval switch. 

Thanks for the explanation @blefaudeux !

anj-s (approved):

SeanNaren:
Fixes the issue upstream in Lightning, thanks so much for the quick fix @blefaudeux :)

PR #589, opened 2021-04-07
preliminary loss scaler

tmarkstrum:
This is a start point for the loss scaler. The testrun.py is a script to verify the DynamicLossScaler.  I will move the file under fairscale/tests/experimental/ later.  

python testrun.py 
tensor([1.0235], requires_grad=True) tensor([1.9690], requires_grad=True)
[tmarkstrum@devvm1106.frc0 ~/fairscale_tmark/fairscale/experimental/optim] 

The DynamicLossScaler currently has a 1-1 mapping with PyTorch’s torch.cuda.amp.GradScaler. Support for scaling up and down. 
Todo:
Support for saving/loading loss scale during training.
Support for skipping updates if we hit NaNs.

### DIFF
--- a/fairscale/experimental/optim/dynamic_loss_scaler.py
+++ b/fairscale/experimental/optim/dynamic_loss_scaler.py
@@ -1,0 +1,139 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from collections import defaultdict, abc
+import warnings
+from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple
+
+
+class OptState(Enum):
+    READY = 0
+    UNSCALED = 1
+    STEPPED = 2
+
+
+def _refresh_per_optimizer_state():
+    return {"stage": OptState.READY}
+
+class DynamicLossScaler(object):
+    def __init__(
+        self,
+        init_scale=2.0 ** 15,
+        scale_factor=2.0,
+        scale_window=2000,
+        tolerance=0.0,
+        threshold=None,
+        min_loss_scale=1e-4,
+    ):
+        self.loss_scale = init_scale
+        self.scale_factor = scale_factor
+        self.scale_window = scale_window
+        self.tolerance = tolerance
+        self.threshold = threshold
+        self._iter = 0
+        self._last_overflow_iter = -1
+        self._last_rescale_iter = -1
+        self._overflows_since_rescale = 0
+        self.min_loss_scale = min_loss_scale
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+        self._scale = None
+
+
+    def scale(self, outputs):
+        print(outputs)
+        return self.loss_scale * outputs
+
+
+    def update(self):
+        if (self._iter - self._last_overflow_iter) % self.scale_window == 0:
+            self.loss_scale *= self.scale_factor
+            self._last_rescale_iter = self._iter
+        self._iter += 1
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+
+
+    def step(self, optimizer, *args, **kwargs):
+        if "closure" in kwargs:
+            raise RuntimeError("Closure use is not currently supported if GradScaler is enabled.")
+
+        optimizer_state = self._per_optimizer_states[id(optimizer)]
+
+        if optimizer_state["stage"] is OptState.STEPPED:
+            raise RuntimeError("step() has already been called since the last update().")
+
+        retval = None
+
+        if optimizer_state["stage"] is OptState.READY:
+            self.unscale_(optimizer)
+
+        retval = optimizer.step(*args, **kwargs)
+
+        optimizer_state["stage"] = OptState.STEPPED
+
+        return retval
+
+
+
+
+    def unscale_(self, optimizer):
+
+        optimizer_state = self._per_optimizer_states[id(optimizer)]
+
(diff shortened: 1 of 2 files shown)


fairscale/experimental/optim/testrun.py
```diff
@@ -0,0 +1,63 @@
...
+
+for epoch in range(n_epochs):
+    yhat = a + b * x_train_tensor
+
+    loss = loss_fn(y_train_tensor, yhat)
+
+    scaler.scale(loss).backward()
+    scaler.step(optimizer)
+    scaler.update()
+    optimizer.zero_grad()
```
anj-s, line 61: Zero the grad right at the beginning of training.

fairscale/experimental/optim/testrun.py
```diff
@@ -0,0 +1,63 @@
...
+
+print(a, b)
```
anj-s, line 63: To verify this the values will need to be deterministic. Are you thinking of any other ways?

fairscale/experimental/optim/testrun.py
```diff
@@ -0,0 +1,63 @@
...
+
+lr = 1e-1
+n_epochs = 1000
+
+loss_fn = nn.MSELoss(reduction='mean')
+
+optimizer = optim.SGD([a, b], lr=lr)
+
+for epoch in range(n_epochs):
+    yhat = a + b * x_train_tensor
```
anj-s, line 54: Shall we use a nn.Linear? It's the same thing but used in Models.

fairscale/experimental/optim/testrun.py
```diff
@@ -0,0 +1,63 @@
...
+x_val, y_val = x[val_idx], y[val_idx]
+
+device = 'cuda' if torch.cuda.is_available() else 'cpu'
+
+torch.manual_seed(42)
+a = torch.randn(1, requires_grad=True, dtype=torch.float, device=device)
+b = torch.randn(1, requires_grad=True, dtype=torch.float, device=device)
+print(a, b)
+
+x_train_tensor = torch.from_numpy(x_train).float().to(device)
```
anj-s, line 43: you can create a tensor of a given size instead of using numpy. Some of our unit tests in fairscale/tests are good examples.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,139 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from collections import defaultdict, abc
+import warnings
+from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple
+
+
+class OptState(Enum):
```
anj-s, line 13: +1 This is great!

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,139 @@
...
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+
+
+    def step(self, optimizer, *args, **kwargs):
+        if "closure" in kwargs:
+            raise RuntimeError("Closure use is not currently supported if GradScaler is enabled.")
+
+        optimizer_state = self._per_optimizer_states[id(optimizer)]
+
+        if optimizer_state["stage"] is OptState.STEPPED:
```
anj-s, line 65: I think we want to check if we have already called `unscale` in which case we just call optimizer.step()

anj-s: I think you are using READY for this. LGTM.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,139 @@
...
+        self.scale_factor = scale_factor
+        self.scale_window = scale_window
+        self.tolerance = tolerance
+        self.threshold = threshold
+        self._iter = 0
+        self._last_overflow_iter = -1
+        self._last_rescale_iter = -1
+        self._overflows_since_rescale = 0
+        self.min_loss_scale = min_loss_scale
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
```
anj-s, line 42: Does this need to be a dict of int->dict? Maybe I am missing how you get the right dict for a given optimizer.

fairscale/experimental/optim/testrun.py
tmarkstrum: Originally, I was thinking as long as the outputs are within a small range around (1.0, 2.0), the test is considered as pass. Strangely, the outputs consistently land at (1.0235, 1.9690), though the data set is random generated.

fairscale/experimental/optim/testrun.py
tmarkstrum: yes. I will use nn.Linear.

fairscale/experimental/optim/testrun.py
tmarkstrum: ok. I will look into it.

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: ok.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,139 @@
...
+class OptState(Enum):
+    READY = 0
+    UNSCALED = 1
+    STEPPED = 2
+
+
+def _refresh_per_optimizer_state():
+    return {"stage": OptState.READY}
+
+class DynamicLossScaler(object):
```
min-xu-ai, line 22: can you please add some documentation of this class? It is purpose, features, usage, etc? Thanks!

PR #590, opened 2021-04-08
[fix] [FSDP] optim state dict should be completely on CPU

sshleifer:
### Problem
non broadcasted params were on GPU, but the rest of the OSD was on CPU. This wastes memory and is undesirable on the client side.

### Solution
Move `no_broadcast_optim_state` params to cpu before unflattening them.

### Regression Tests
- test that all OSD state on CPU
- test that no cuda memory is used
- both of these cases fail without moving the params to GPU.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -8,6 +8,7 @@ import copy
 from enum import Enum, auto
 import functools
 from math import inf
+import time
 import traceback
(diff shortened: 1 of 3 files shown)


sshleifer:
I think the CI failure here is flaky @min-xu-ai:

```bash
pytest tests/nn/data_parallel/test_fsdp_regnet.py 
```
passes for me on devfair and learnfair.

sshleifer:
I'm merging because this shouldn't be controversial at all. Comments will be addressed in the next PR!

PR #598, opened 2021-04-09
[feat][OSS] Expose different possible partitioning strategies for OSS

blefaudeux:
# Before submitting

## What does this PR do?
Make it possible to choose an ordering-aware partitioning for OSS. In a follow up PR, this is used by ShardedDDP to let the optimizer step and subsequent broadcast step early on, as soon as the rank has all the gradients it needs. This was an old idea which should benefit speed on multinode, given the users already investing in OSS I figured that could be useful

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -3,12 +3,11 @@
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
 
-from collections import OrderedDict
-import copy
+from collections import OrderedDict, deque
 from itertools import chain
 import logging
 from math import inf
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union
+from typing import TYPE_CHECKING, Any, Callable, Deque, Dict, List, Optional, Type, Union
 
 import torch
 import torch.distributed as dist
@@ -17,7 +16,8 @@ from torch.optim import SGD, Optimizer
 
 from fairscale.nn.misc import ParamBucket
 
-from .utils import broadcast_object, calc_grad_norm, get_global_rank, recursive_copy_to_device
+from .oss_utils import partition_keep_ordering, partition_memory
+from .utils import Workhandle, broadcast_object, calc_grad_norm, get_global_rank, recursive_copy_to_device
 
 __all__ = ["OSS"]
 
@@ -25,6 +25,8 @@ if TYPE_CHECKING:  # pragma: no cover
     from torch.optim.optimizer import _params_t
 else:
     _params_t = Any
+
+import enum
 
 
 class OSS(Optimizer):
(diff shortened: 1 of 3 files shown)


fairscale/optim/oss.py
```diff
@@ -125,29 +137,12 @@ def partition_parameters(self) -> List[List[dict]]:
...
-
-                for rank, params in enumerate(param_lists):
-                    param_group_rank = copy.copy(param_group)
-                    param_group_rank["params"] = params
-                    self._partition_parameters[rank].append(param_group_rank)
+            if self._partition_type == OSS.PartitionType.MEMORY:
+                self._partition_parameters = partition_memory(self.param_groups, self.world_size)
+
+            elif self._partition_type == OSS.PartitionType.ORDERING:
+                # TODO: update the partitioning and ordering given the backward pass information
```
blefaudeux, line 144: a follow up PR does the non-blocking-backward change from shardedDDP, and dynamic partitioning given the ordering learnt during the backward would be the final nail in the coffin

fairscale/optim/oss.py
```diff
@@ -555,20 +547,34 @@ def _broadcast_params(self) -> None:
...
+                            src=self._local_to_global_rank[dst_rank],
+                            group=self.group,
+                            async_op=True,
+                        ),
+                        None,
                     )
                 )
 
-        _ = list(filter(lambda x: x.wait(), work_handles))
+        self._consume_work_handles()
```
blefaudeux, line 562: I know that I went back and forth on that, but exposing these handles now so that ShardedDPP can use them in a subsequent PR. The idea is that ShardedDDP can fire all the reduce calls, but does not have to block for all of them to be done before handing it over to OSS, it just has to wait for the gradients this rank owns. It can then pass the handles to OSS, which has to wait for them to be done + subsequent broadcasts before the next FW. This way some of the step+broadcast could overlap with the BW

fairscale/optim/oss_utils.py
```diff
@@ -0,0 +1,107 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
```
blefaudeux, line 5: moving the different partitioning schemes to a dedicated file, to simplify a bit on the OSS side

tests/optim/test_oss.py
```diff
@@ -857,6 +863,8 @@ def check_optimizer_equivalence(optimizer: Type[torch.optim.Optimizer], change_t
             optim=optimizer,
             group=None,
             broadcast_buffer_size=2 ** 10,
+            broadcast_fp16=broadcast_fp16,
```
blefaudeux, line 866: note that broadcastfp16 was missing here, my mistake :/

blefaudeux:
Not sure that it's worth it, would be a significant investment to get the whole suite across the finish line and FSDP makes it probably moot. This particular PR and the next one work fine, tested on FB cluster with PySpeech, no speed up though as expected (would need parameter re-ordering + dedicated cuda streams)

PR #606, opened 2021-04-14
[fix] [FSDP] Make _get_default_cuda_device more robust to modules without params

myleott:
Make `_get_default_cuda_device` fall back to the current device when the input module doesn't have any parameters.

Also fix a small doc issue in auto_wrap.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1540,11 +1540,14 @@ class FullyShardedDataParallel(nn.Module):
 
 def _get_default_cuda_device(module: nn.Module) -> torch.device:
     """Try to infer CUDA device from module parameters."""
-    compute_device = next(module.parameters()).device
-    if compute_device.type != "cuda":
-        # Fall back to current CUDA device.
-        compute_device = torch.device("cuda")
-    return compute_device
+    try:
+        compute_device = next(module.parameters()).device
+        if compute_device.type == "cuda":
+            return compute_device
+    except StopIteration:
+        pass
+    # Fall back to current CUDA device
+    return torch.device("cuda")
 
 
 @torch.no_grad()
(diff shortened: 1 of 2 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1540,11 +1540,14 @@ def _print_r0(self, msg: str, restart: bool = False) -> None:
...
-        compute_device = torch.device("cuda")
-    return compute_device
+    try:
+        compute_device = next(module.parameters()).device
+        if compute_device.type == "cuda":
+            return compute_device
+    except StopIteration:
+        pass
+    # Fall back to current CUDA device
+    return torch.device("cuda")
```
min-xu-ai, line 1550: when if compute_device is cpu on line 1544, the old code will return cpu, but now we return cuda? Is that the intent?

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -88,9 +88,11 @@ def enable_wrap(auto_wrap_policy: Optional[Callable] = None, **wrapper_kwargs: A
         with enable_wrap(**params):
             # Wraps layer in FSDP by default if within context
             self.l1 = wrap(torch.nn.Linear(5, 5))
-            # Wraps children modules based on a different min_num_params
-            my_auto_wrap_policy = functools.partial(auto_wrap_policy, min_num_params=1e7)
-            self.l2 = auto_wrap(TransformerBlock(), shuold_wrap=my_auto_wrap_policy)
+            self.l2 = auto_wrap(
```
min-xu-ai, line 91: Thanks!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
myleott: I think in the old code it should also return "cuda", since in that case (`compute.device.type == "cpu"`) we'd go to line 1546

min-xu-ai: you are right, the old code is `!=`. I misread it.

min-xu-ai (approved):
LGTM!

PR #607, opened 2021-04-14
[FSDP] Consolidate cpu_adam optimizer state dict

sshleifer:
### Problem
- If an optimizer, like `CPUAdam`, keeps state on CPU, `all_gather` will break.

### Approach
- move tensor entries to cuda before `all_gather`


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1411,6 +1411,9 @@ def _gather_optim_state(
             buffer = None  # for sharded tensors
             singleton_buffer = None  # for singleton tensors
             for buffer_name, t in v.items():
+                if torch.is_tensor(t):
+                    t = t.cuda()
```
myleott, line 1415: use `.to(self.compute_device)` instead, since it's possible the user hasn't properly called `torch.cuda.set_device`

myleott (approved):

PR #612, opened 2021-04-15
draft change

min-xu-ai:


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -176,49 +186,62 @@ def __init__(
...
         flatten_parameters: bool = True,
         cpu_offload: bool = False,
-        compute_dtype: Optional[torch.dtype] = None,
-        buffer_dtype: Optional[torch.dtype] = None,
         move_grads_to_cpu: Optional[bool] = None,
         bucket_cap_mb: int = 25,
         compute_device: Optional[torch.device] = None,
-        no_broadcast_optim_state: Optional[bool] = False,
         state_dict_device: Optional[torch.device] = None,
+        no_broadcast_optim_state: Optional[bool] = False,
```
min-xu-ai, line 198: this reordering just make two device related options more together.

tests/ci_test_list_1.txt
```diff
@@ -1,2 +1,3 @@
 tests/nn/misc/test_flatten_params_wrapper.py
 tests/nn/data_parallel/test_fsdp.py
+tests/nn/data_parallel/test_fsdp_conv.py
```
min-xu-ai, line 3: I added this new test to catch known issues with pytorch version. I have seen py1.9 core dump on it.

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -343,8 +339,29 @@ def test1(temp_files, ddp_ref, precision, flatten):
     fsdp_config["mixed_precision"] = precision == "mixed"
     fsdp_config["flatten_parameters"] = flatten == "flatten"
 
-    if fsdp_config["mixed_precision"] and torch_cuda_version() < (11, 0):
-        pytest.skip("Only CUDA 11 is supported with AMP equivalency")
+    if fsdp_config["mixed_precision"]:
+        # For convolution layers, we need fp32_compute_dtype. Otherwise, we run
```
min-xu-ai, line 343: this documents the findings and best practices with conv layers. Later, can turning into some kind guide or doc.

min-xu-ai:
tests are still running, but I think this is ready.

cc: @QuentinDuval @prigoyal

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -126,26 +126,37 @@ class FullyShardedDataParallel(nn.Module):
...
+            if ``True``, use FP32 for full parameters for computation.
+            Default: None, which means default to FP16 or FP32 based on the
+            *``mixed_precision``* argument.
+        fp32_buffer_dtype (bool, Optional):
+            if ``True``, use FP32 for buffers for computation.
+            Default: None, which means using the value of the
+            *``fp32_compute_dtype``* argument.
+        fp32_reduce_scatter (bool):
+            if ``True``, then reduce-scatter gradients in FP32.
+            This is only relevant when *``mixed_precision``* is ``True``.
```
min-xu-ai, line 146: I filed an issue to later perhaps decouple this with mixed precision. after ddp has a fp16 grad hook even without mixed precision.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -126,26 +126,37 @@ class FullyShardedDataParallel(nn.Module):
...
+            Default: False
         flatten_parameters (bool, Optional):
             if ``True``, flatten parameters into a single contiguous tensor,
             which improves training speed.
+            Default: True
         cpu_offload (bool, Optional):
-            if ``True``, offload FP32 params to CPU. This is only relevant when
-            *``mixed_precision``* is ``True``.
-        compute_dtype (torch.dtype, Optional):
```
myleott, line 139: reflecting on this, I think we should still support compute_dtype, and potentially have fp32_compute_dtype/buffer_dtype as convenience flags instead. The reason is that the latest A100 GPUs support torch.bfloat16 and torch.tfloat32, which are actually compatible with FSDP if you do `FSDP(mixed_precision=True, compute_dtype=torch.bfloat16)`

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -227,16 +224,18 @@ def _test_func(
     rank_0_output,
     state_after,
 ):
+    # Make convs more deterministic.
+    torch.backends.cudnn.benchmark = True
+    torch_set_deterministic(True)
+    os.putenv("CUBLAS_WORKSPACE_CONFIG", ":4096:8")
```
myleott, line 230: wow, didn't know about this env variable :)

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -343,8 +342,36 @@ def test1(temp_files, ddp_ref, precision, flatten):
...
+    if fsdp_config["mixed_precision"]:
+        # For convolution layers, we need fp32_compute_dtype. Otherwise, we run
+        # into conv layer errors like below with mixed precision with larger
+        # regnet128Gf models when flatten_parameter is true:
+        #
+        #   RuntimeError: Unable to find a valid cuDNN algorithm to run convolution
+        #   or
+        #   RuntimeError: cuDNN error: CUDNN_STATUS_BAD_PARAM
+        #
+        # If you know the solutions to the above issue, please let me (min-xu-ai) know. Thanks!
```
myleott, line 354: Can we post in some internal forum about this? Maybe https://fb.workplace.com/groups/1405155842844877/

tests/nn/data_parallel/test_fsdp_regnet.py
```diff
@@ -343,8 +342,36 @@ def test1(temp_files, ddp_ref, precision, flatten):
...
+        #
+        # Also, without fp32_compute_dtype, Linear layer with bias=True won't match DDP
+        # numerically, which may not be necessary for convergence depending on the model and
+        # the optimizer and a number of other factors.
```
myleott, line 358: Can you clarify what this is referring to, perhaps with a small repro?

For some context: the recommended mixed precision recipe for PyTorch is to use an FP32 model and autocast weights to FP16 on-the-fly.

Contrast that with `FSDP(mixed_precision=True)`, where FSDP will move everything to FP16 internally, and will rely on autocast to up-cast things to FP32 when needed (e.g., for Softmax).

When comparing these two recipes, they are not expected to give identical results. Is there some specific concern for Linear with bias, or is this just a general observation about differences between the two mixed precision recipes?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: Good point. Let me add them back.

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: yeah, it was printed out by the pytorch error msg when I don't have it. determinism seems to help avoid some small numerical mismatch on the CI machines (but not locally).

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: Good idea! Will do.

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: I will actually preserve this part in this test. Will be added in a new commit. So the repro will be in this test.

Here is what I think is happening:

without bias: the operation is sum(x * y for x, y in zip(vector1, vector2)) for each resulting weight.
with bias: the operation is sum(x*y for x, y in zip(vector1, vector2)) + bias for each weight.

The vector1/vector2 are in fp16 whether or not they were originally in fp32 or fp16, because of auto-cast.

However, with bias, pytorch's mode will be adding a fp32 term but in FSDP, we will be adding a fp16 term. I suspect that it is the issue causing the difference. What do you think?

tests/nn/data_parallel/test_fsdp_regnet.py
min-xu-ai: done: https://fb.workplace.com/groups/1405155842844877/permalink/4614727588554337/

tests/nn/data_parallel/test_fsdp_regnet.py
myleott: > However, with bias, pytorch's mode will be adding a fp32 term but in FSDP, we will be adding a fp16 term.

Are you sure? That could be possible, but I'm not able to repro locally (PyTorch 1.8.1):
```python
>>> import torch
>>>
>>> m = torch.nn.Linear(1024, 4096, bias=True).cuda()
>>> m2 = torch.nn.Linear(1024, 4096, bias=True).cuda()
>>> m2.load_state_dict(m.state_dict())
>>>
>>> m2 = m2.half()
>>> m.bias.dtype
torch.float32
>>> m2.bias.dtype
torch.float16
>>>
>>> with torch.cuda.amp.autocast():
...   y = m(x)
...   y2 = m2(x)
...
>>> y.dtype
torch.float16
>>> y2.dtype
torch.float16
>>> torch.testing.assert_allclose(y.float(), y2.float())
```

min-xu-ai: interesting, with my 1.6:
```
  File "/private/home/m1n/e/py38_fs/lib/python3.8/site-packages/torch/testing/__init__.py", line 223, in assert_allclose
    raise AssertionError(msg)
AssertionError: With rtol=0.0001 and atol=1e-05, found 202 element(s) (out of 4096) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.00048828125 (0.56494140625 vs. 0.564453125), which occurred at index 1915.
```

same code as yours.

myleott: 😮

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: I thought about this more, should we do something like the following?

compute_dtype, buffer_dtype, reduce_scatter_dtype all take the same data type input, i.e. Union[torch.dtype, str]. The reason for str is that for easier config where users can supply string to be converted to the right dtype (like "fp16" -> torch.float16, etc.).

What do you think?

myleott: Yeah, seems cleaner too, that way we don't have redundant kwargs.

There seems to have been some discussion about this in a PyTorch issue: https://github.com/pytorch/pytorch/issues/40471, but sadly no public interface. We can easy enumerate the dtypes we support: float16, bfloat16, float32. I was wrong about tensorfloat32, it's not actually a separate dtype, but we should support at least the other three.

min-xu-ai:
Closing this since Yanli is upstreaming FSDP to pytorch and do an API change. For fairscale, we will keep the current APIs and let it evolve more.

Issue #610, opened 2021-04-15
Support for frozen parameters (Linear classification benchmarks)


QuentinDuval:
## ❓ Support for frozen parameters

In the context of VISSL, we are training a SWAV model with FSDP and then want to evaluate this model using linear evaluation, again using FSDP to wrap our model.

In the context of linear evaluation, only the linear layer on top of the "trunk" will be learned: the representations of the trunk, fed to the linear layer, will stay constant, and so we set `requires_grad=False` on the parameters of the trunk.

While this works with DDP, this leads to an assert in the case of FSDP:

```
expected to be in states [<TrainingState.BACKWARD_POST: 4>] but current state is TrainingState.IDLE
```

To help you dig into this issue, I have reproduced a small test case which demonstrate this issue:
- run this script with `-f` to enable FSDP (otherwise, we use DDP)
- run this script with `-l` (for linear) to freeze the trunk

You will get the assert if you enable both `-f` and `-l` and everything will work fine otherwise.

```
import argparse

import torch
import torch.distributed as dist
import torch.multiprocessing as mp
import torch.nn as nn
import torch.optim as optim
from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP
from torch.nn.parallel import DistributedDataParallel


class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.trunk = nn.Sequential(
            nn.Conv2d(3, 64, kernel_size=3),
            nn.ReLU(inplace=True),
            nn.AdaptiveAvgPool2d(output_size=(1, 1)),
            nn.Flatten(),
        )
        self.head = nn.Linear(64, 10)

    def forward(self, x):
        return self.head(self.trunk(x))


def create_model(with_fsdp: bool):
    model = Model()
    if with_fsdp:
        model.trunk = FSDP(model.trunk)
        model.head = FSDP(model.head)
    return model


def _distributed_worker(gpu_id: int, with_fsdp: bool, with_linear: bool):
    torch.cuda.set_device(gpu_id)
    dist.init_process_group(
        backend="nccl", init_method="tcp://127.0.0.1:9099", world_size=2, rank=gpu_id
    )

    torch.manual_seed(0)
    torch.backends.cudnn.deterministic = True
    batch = torch.randn(size=(2, 3, 224, 224)).cuda()

    model = create_model(with_fsdp)
    model = model.cuda()

    if with_linear:
        for name, param in model.named_parameters():
            if "trunk" in name:
[...]

PR #614, opened 2021-04-16
FSDP: fixing training with freezing weights

min-xu-ai:
- an assert is changed to catch this case correctly
- unit test added (based on Quentin's test code) for this case and
  compare DDP and FSDP

fixes: #610

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1187,11 +1187,23 @@ class FullyShardedDataParallel(nn.Module):
         for m in self.modules():  # includes self
             if isinstance(m, FullyShardedDataParallel):
                 _remove_shard_bwd_hook(m)
-                if m._has_params:
-                    m.assert_state(TrainingState.BACKWARD_POST)
+                if m.training_state == TrainingState.IDLE:
+                    # Unlikely case, only happens if has params but none of the param
+                    # has requires_grad.
+                    if not m._has_params:
+                        # using print in backward for better debugging.
+                        print("ERROR: IDLE instance should have params: " + str(self))
+                        raise RuntimeError()
+                    for p in m.params:
+                        if p.requires_grad:
+                            print("ERROR: IDLE instance not have params with grad: " + str(p.shape))
+                            raise RuntimeError()
                 else:
-                    m.assert_state(TrainingState.BACKWARD_PRE)
-                m.training_state = TrainingState.IDLE
+                    if m._has_params:
+                        m.assert_state(TrainingState.BACKWARD_POST)
+                    else:
+                        m.assert_state(TrainingState.BACKWARD_PRE)
+                    m.training_state = TrainingState.IDLE
 
     @torch.no_grad()
     def _rebuild_full_params(self, force_full_precision: bool = False) -> Optional[List[Tuple[torch.Tensor, bool]]]:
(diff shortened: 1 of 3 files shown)


tests/nn/data_parallel/test_fsdp_freezing_weights.py
```diff
@@ -0,0 +1,125 @@
...
+    torch.backends.cudnn.deterministic = True
+    batch = torch.randn(size=(2, 3, 224, 224)).cuda()
+
+    model = _create_model(with_fsdp)
+    model = model.cuda()
+
+    # freezing the trunk.
+    for name, param in model.named_parameters():
+        if "trunk" in name:
+            param.requires_grad = False
```
blefaudeux, line 68: From what I understand, VISSL is doing something different, the grads are removed after the BW pass. requires_grad being changed makes more sense to me since you actually save time and space during the BW instead of computing then wiping it. This being said DDP does not support changing .requires_grad over time, the model needs a re-wrap in that case, may be the reason why this was done like that to begin with

min-xu-ai: Thanks for taking a look Ben and great insights!!!

Indeed, setting grad to None is for a different type of freezing used in vissl. And we have tested that in vissl and seems to work fine. This is for "linear evaluation" type of training, where only the head is allowed to learn. I.e. vissl uses both methods.

In FSDP, change requires_grad seems to be fine but I haven't tested all cases exhaustively. It might not work with different flattening and nested wrap boundaries, etc.

tests/nn/data_parallel/test_fsdp_freezing_weights.py
```diff
@@ -0,0 +1,125 @@
...
+    torch.manual_seed(0)
+    torch.backends.cudnn.deterministic = True
+    batch = torch.randn(size=(2, 3, 224, 224)).cuda()
+
+    model = _create_model(with_fsdp)
+    model = model.cuda()
+
+    # freezing the trunk.
+    for name, param in model.named_parameters():
+        if "trunk" in name:
```
myleott, line 67: nit: you could also do:
```python
for param in model.trunk.parameters():
    param.requires_grad = False
```

myleott (approved):
Nice fix!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1187,11 +1187,23 @@ def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
...
+                            print("ERROR: IDLE instance not have params with grad: " + str(p.shape))
+                            raise RuntimeError()
                 else:
-                    m.assert_state(TrainingState.BACKWARD_PRE)
-                m.training_state = TrainingState.IDLE
+                    if m._has_params:
+                        m.assert_state(TrainingState.BACKWARD_POST)
+                    else:
+                        m.assert_state(TrainingState.BACKWARD_PRE)
+                    m.training_state = TrainingState.IDLE
```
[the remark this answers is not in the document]
min-xu-ai: Love this simplification! Thanks so much for suggesting it.

I added two commits, one for addressing the comments and another to test both requires_grad=False and p.grad.None so that both methods are covered.

min-xu-ai:
will merge tonight or tomorrow morning and make a release.

PR #621, opened 2021-04-21
[fix]:  let FSDP handle model with multiple forward pass and checkpoint

### LINKED ISSUE #617: FSDP and checkpoint_wrapper with support multiple forward passes in sub-modules
## 🐛 Bug

In the context of VISSL, we are training a SWAV model with FSDP and activation checkpointing. The SWAV algorithm requires two forward passes to be performed to compute the loss (each pass deals with images in different resolutions).

When enabling activation checkpointing (using the pattern `FSDP(checkpoint_wrapper(module))`), this leads to one of the following assert during the `run_backward` execution:

**Assert 1**

```
expected to be in states [<TrainingState.BACKWARD_POST: 4>] but current state is TrainingState.IDLE
```

**Assert 2**

```
RuntimeError: expected to be in states [<TrainingState.IDLE: 1>, <TrainingState.BACKWARD_PRE: 3>] but current state is TrainingState.BACKWARD_POST
```

The test cases for both use cases follows.

## Scenario 1 (independant of VISSL)

Use this script to reproduce the **Assert 1**:
- option `-d` enables a double forward pass to compute the loss
- option `-f` enables FSDP in place of DDP to handle the distributed training
- option `-c` enables activation checkpointing `checkpoint_wrapper` 

All combination of options work except `-fdc` pointing to the fact that FSDP and `checkpoint_wrapper` do not work well together when a module trigger 2 forward pass of an underlying sub-module.

```
import argparse
from typing import List, Union

import torch
import torch.nn as nn
import torch.optim as optim
from fairscale.nn import checkpoint_wrapper
from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP, auto_wrap_bn
from torch.nn.parallel import DistributedDataParallel


class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.block1 = nn.Sequential(
            nn.Conv2d(3, 64, kernel_size=3),
            nn.BatchNorm2d(64),
            nn.ReLU(inplace=True),
        )
        self.block2 = nn.Sequential(
            nn.Conv2d(64, 128, kernel_size=3),
            nn.BatchNorm2d(128),
            nn.ReLU(inplace=True),
            nn.AdaptiveAvgPool2d(output_size=(1, 1)),
            nn.Flatten(),
        )
        self.head = nn.Linear(128, 10)

    def forward(self, x: Union[torch.Tensor, List[torch.Tensor]]):
        if isinstance(x, torch.Tensor):
            return self.head(self.block2(self.block1(x)))
        elif isinstance(x, list):
            ys = [self.head(self.block2(self.block1(e))) for e in x]
            return torch.cat(ys, dim=0)


def create_model(with_fsdp: bool, with_checkpoint: bool):
    model = Model()
    if with_fsdp:
        model.block1 = auto_wrap_bn(model.block1, single_rank_pg=False)
        model.block2 = auto_wrap_bn(model.block2, single_rank_pg=False)
        if with_checkpoint:
            model.block2 = checkpoint_wrapper(model.block2)
        model.block1 = FSDP(model.block1)
        model.block2 = FSDP(model.block2)
        model.head = FSDP(model.head)
    else:
        if with_checkpoint:
            model.block2 = checkpoint_wrapper(model.block2)
    return model


def _distributed_worker(gpu_id: int, with_fsdp: bool, double_forward: bool, with_checkpoint: bool):
    torch.cuda.set_device(gpu_id)
    dist.init_process_group(
        backend="nccl", init_method="tcp://127.0.0.1:9099", world_size=2, rank=gpu_id
    )

    torch.manual_seed(0)
    torch.backends.cudnn.deterministic = True
    if double_forward:
        batch = [
            torch.randn(size=(2, 3, 224, 224)).cuda(),
            torch.randn(size=(2, 3, 96, 96)).cuda(),
        ]
    else:
        batch = torch.randn(size=(4, 3, 224, 224)).cuda()

    model = create_model(with_fsdp, with_checkpoint)
    model = model.cuda()

    if with_fsdp:
        model = FSDP(model)
    else:
        model = DistributedDataParallel(model, device_ids=[gpu_id])

    if gpu_id == 0:
        print(model)

    target = torch.LongTensor([0, 1, 2, 3]).cuda()
    criterion = nn.CrossEntropyLoss()
    optimizer = optim.SGD(model.parameters(), lr=0.1, momentum=0.9)

    for iteration in range(3):
        out = model(batch)
        loss = criterion(out, target)
        print("Loss", iteration, ":", loss.item())
        optimizer.zero_grad()
        loss.backward()
        optimizer.step()


if __name__ == "__main__":
    parser.add_argument("-f", "--fsdp", action="store_const", const=True, default=False)
    parser.add_argument("-d", "--double", action="store_const", const=True, default=False)
    parser.add_argument("-c", "--checkpoint", action="store_const", const=True, default=False)
    args = parser.parse_args()
    mp.spawn(_distributed_worker, (args.fsdp, args.double, args.checkpoint), nprocs=2)
```

## Scenario 2 (VISSL use case)

Use this script to reproduce the **Assert 2**:
- option `-d` enables a double forward pass to compute the loss
- option `-f` enables FSDP in place of DDP to handle the distributed training
- option `-c` enables activation checkpointing `checkpoint_wrapper` 

The main difference with the previous use case is that the model is much bigger (we use a RegNet16). The assert is different but also occurs in `run_backward`.

Again this error occurs only if FSDP is used in conjunction with `checkpoint_wrapper` and if a module triggers 2 forward passes on an underlying sub-module:
- FSDP works fine with `checkpoint_wrapper` in case of single forward pass
- FSDP works fine with double forward passes if not coupled with `checkpoint_wrapper`

```
import argparse

import torch
import torch.nn as nn
import torch.optim as optim
from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP
from hydra.experimental import initialize_config_module, compose
from torch.nn.parallel import DistributedDataParallel

from vissl.models import build_model
from vissl.utils.hydra_config import convert_to_attrdict


def create_config():
    with initialize_config_module(config_module="vissl.config"):
        cfg = compose(
            "defaults",
            overrides=[
                "config=pretrain/swav/swav_8node_resnet",
                "+config/pretrain/swav/models=regnet16Gf",
                "config.SEED_VALUE=2",
                "config.MODEL.AMP_PARAMS.USE_AMP=True",
                "config.MODEL.AMP_PARAMS.AMP_TYPE=pytorch",
                "config.MODEL.SYNC_BN_CONFIG.CONVERT_BN_TO_SYNC_BN=True",
                "config.MODEL.SYNC_BN_CONFIG.SYNC_BN_TYPE=pytorch",
                "config.OPTIMIZER.num_epochs=1",
                "config.OPTIMIZER.param_schedulers.lr.lengths=[0.1, 0.9]",
                "config.OPTIMIZER.use_larc=False",
                "config.LOSS.swav_loss.epsilon=0.03",
                "config.DATA.TRAIN.DATASET_NAMES=[imagenet1k_folder]",
                "config.DATA.TRAIN.DATA_SOURCES=[disk_folder]",
                "config.DATA.TRAIN.BATCHSIZE_PER_REPLICA=16",
                "config.DISTRIBUTED.NCCL_DEBUG=False",
                "config.SLURM.LOG_FOLDER=/checkpoint/qduval/log_runs",
                "config.SLURM.PARTITION=learnfair",
                "config.SLURM.USE_SLURM=true",
                "config.DISTRIBUTED.NUM_NODES=8",
                "hydra.verbose=true",
                "config.MODEL.FSDP_CONFIG.flatten_parameters=True",
                "config.MODEL.FSDP_CONFIG.mixed_precision=False",
                "config.MODEL.FSDP_CONFIG.fp32_reduce_scatter=False",
            ],
        )
    args, config = convert_to_attrdict(cfg)
    return config


def create_model(with_fsdp: bool, with_checkpoint: bool):
    config = create_config()
    if with_checkpoint:
        config["MODEL"]["ACTIVATION_CHECKPOINTING"]["USE_ACTIVATION_CHECKPOINTING"] = True
    if with_fsdp:
        config["MODEL"]["TRUNK"]["NAME"] = "regnet_fsdp"
        config["MODEL"]["HEAD"]["PARAMS"][0][0] = "swav_head_fsdp"
    else:
        config["MODEL"]["TRUNK"]["NAME"] = "regnet_v2"
    return build_model(config["MODEL"], config["OPTIMIZER"])


def create_batch(double_forward: bool):
    torch.manual_seed(0)
    if double_forward:
        return [
            torch.randn(size=(2, 3, 224, 224)).cuda(),
            torch.randn(size=(2, 3, 96, 96)).cuda(),
        ]
    else:
        return torch.randn(size=(4, 3, 224, 224)).cuda()


def _distributed_worker(gpu_id: int, with_fsdp: bool, double_forward: bool, with_checkpoint: bool):
    torch.cuda.set_device(gpu_id)
    dist.init_process_group(
        backend="nccl", init_method="tcp://127.0.0.1:9099", world_size=2, rank=gpu_id
    )

    torch.manual_seed(0)
    torch.backends.cudnn.deterministic = True

    model = create_model(with_fsdp, with_checkpoint)
    model = model.cuda()

    if with_fsdp:
        model = FSDP(model)
    else:
        model = DistributedDataParallel(model, device_ids=[gpu_id])

    if gpu_id == 0:
        print(model)

    criterion = nn.MSELoss()
    optimizer = optim.SGD(model.parameters(), lr=0.001, momentum=0.9)

    for iteration in range(3):
        batch = create_batch(double_forward)
        out = sum(o.sum() for o in model(batch)[0])
        loss = criterion(out, torch.tensor(0.).cuda(gpu_id))
        print("Loss", iteration, ":", loss.item())
        optimizer.zero_grad()
        loss.backward()
        optimizer.step()


if __name__ == "__main__":
    parser.add_argument("-f", "--fsdp", action="store_const", const=True, default=False)
    parser.add_argument("-d", "--double", action="store_const", const=True, default=False)
    parser.add_argument("-c", "--checkpoint", action="store_const", const=True, default=False)
    args = parser.parse_args()
    mp.spawn(_distributed_worker, (args.fsdp, args.double, args.checkpoint), nprocs=2)
```

## Environment

- Pytorch 1.6 + cu102
- Fairscale 0.3.5
- VISSL master@3a3934b3cfc1802079da85f7ae703497798f85b6
- `module load cuda/11.0 cudnn/v8.0.3.33-cuda.11.0 NCCL/2.8.3-1-cuda.11.0`

## Interested parties

CC: @min-xu-ai @myleott @prigoyal

min-xu-ai:
This PR adds support of multiple passes of a module in a forward pass and activation checkpoint on the module at the same time.

Right now, the tested cases are:

1. FSDP(ckpt(module))
2. FSDP(ckpt(module, FSDP(BN), module))

Next commit will test and enable FSDP(ckpt(), ..., ckpt()) type of cases.

The fix is to add an option to `checkpoint_wrapper` so that it can keep a counter on the checkpointed modules so that FSDP can check the counter to determine if the backward callback fired from the module is the last one in the bigger, outer backward pass.

A new API to override the pre/post gradient divide factors. It is currently used in the test to make sure numerical match with DDP.

# Before submitting

## What does this PR do?
Fixes part of  #617 .

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -115,9 +115,15 @@ class FullyShardedDataParallel(nn.Module):
         an assert on the backward pass. The solution is to leave some parameters
         to the outer FSDP.
 
+    .. warning::
+
+        If activation checkpointing is used with FSDP, it is strongly encouraged
+        to use ``checkpoint_wrapper`` function from FairScale instead of the
+        ``checkpoint`` function from PyTorch.
+
     Args:
         module (nn.Module):
-            module to checkpoint
+            module to be wrapped with FullyShardedDataParallel.
         process_group (Optional):
             process group for sharding
         reshard_after_forward (bool, Optional):
@@ -207,7 +213,7 @@ class FullyShardedDataParallel(nn.Module):
         self.no_broadcast_optim_state = no_broadcast_optim_state
         self.state_dict_device = state_dict_device or self.compute_device
 
-        self.gradient_predivide_factor: int = self._get_gradient_predivide_factor(self.world_size)
+        self.gradient_predivide_factor: float = self._get_gradient_predivide_factor(self.world_size)
         self.gradient_postdivide_factor: float = self.world_size / self.gradient_predivide_factor
 
         self.numel_padded_per_param: List[int] = []
@@ -280,11 +286,11 @@ class FullyShardedDataParallel(nn.Module):
         self._pre_backward_hook_has_run = False
         # self._b2_counter = 0
 
-    def _get_gradient_predivide_factor(self, world_size: int) -> int:
-        factor = 1
+    def _get_gradient_predivide_factor(self, world_size: int) -> float:
+        factor: int = 1
         while world_size % factor == 0 and world_size / factor > factor:
-            factor = factor * 2
-        return factor
+            factor *= 2
+        return float(factor)
 
     def set_gradient_divide_factors(self, pre: float, post: float, recursive: bool) -> None:
         """Allowing user to override the pre and post divide factors.
@@ -996,7 +1002,7 @@ class FullyShardedDataParallel(nn.Module):
 
             # Start of a backward pass.
             # if self.rank == 0:
-            #    print("XXX bwd pre", self._id)
+            #    print("XXX bwd pre", self._id if hasattr(self, "_id") else "bn")
             self.assert_state([TrainingState.IDLE, TrainingState.BACKWARD_PRE])
             self.training_state = TrainingState.BACKWARD_PRE
 
@@ -1094,8 +1100,14 @@ class FullyShardedDataParallel(nn.Module):
         the local optimizer only sees the relevant parameter shard.
         """
         # First hook callback will see PRE state. If we have multiple params,
-        # then subsequent hook callbacks will see POST state.
-        self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
+        # then subsequent hook callbacks will see POST state. When checkpoint
+        # fwd counter is used, IDLE is also possible since the pre-backward hook
+        # is not triggered (see ``auto_wrap_bn`` below, we have to use
+        # FSDP(checkpoint(conv, FSDP(bn), ...)), with reshard_after_forward=False).
+        if hasattr(self, "_checkpoint_fwd_counter"):
+            self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST, TrainingState.IDLE])
+        else:
+            self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
         # if self.rank == 0:
         #    print("XXX bwd post", self._id)
         self.training_state = TrainingState.BACKWARD_POST
@@ -1109,14 +1121,16 @@ class FullyShardedDataParallel(nn.Module):
         #        print("XXX skip")
         #        return
 
+        if param.grad.requires_grad:
+            raise RuntimeError("FullyShardedDataParallel only works with gradients that "
+                               "don't require gradients")
+
         # If this is a checkpointed module, we check if the following
         # counter reaches 0. If not, it is not the final backward call
         # for this module yet. Therefore, we early return in that case.
         if hasattr(self._fsdp_wrapped_module, "_checkpoint_fwd_counter"):
             if self._fsdp_wrapped_module._checkpoint_fwd_counter != 0:
                 return
-        if param.grad.requires_grad:
-            raise RuntimeError("FullyShardedDataParallel only works with gradients that don't require grad")
 
         if self._require_backward_grad_sync or self.reshard_after_forward:
             # Free full params. As a special case, we don't free the full params
@@ -1139,10 +1153,7 @@ class FullyShardedDataParallel(nn.Module):
         # of FSDP (root and children) make this attempt here to queue to ensure it is queued
         # no matter which instance(s) has(have) params.
         assert self._queue_wait_for_post_backward_closure is not None or not self._is_root
-        # if self._queue_wait_for_post_backward_closure is not None and (self._id != "block2":
         if self._queue_wait_for_post_backward_closure is not None:
-            # if self.rank == 0:
-            #    print("XXX queued by", self._id)
             self._queue_wait_for_post_backward_closure()
 
         if not self._require_backward_grad_sync:
@@ -1739,6 +1750,10 @@ def auto_wrap_bn(module: nn.Module, single_rank_pg: bool = False, process_group:
         "process_group": pg,
         "mixed_precision": False,  # Keep the weights in FP32.
         "flatten_parameters": False,  # Do not flatten.
+        # Reshard==False is good for performance. When FSDP(checkpoint(FSDP(bn))) is used, this
+        # **must** be False because BN's FSDP wrapper's pre-backward callback isn't called
+        # within the checkpoint's outer backward when multiple forward passes are used.
+        "reshard_after_forward": False,
     }
 
     with enable_wrap(wrap_bn_only_policy, **fsdp_config):
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -207,8 +207,8 @@ class FullyShardedDataParallel(nn.Module):
 
         self.gradient_predivide_factor: int = self.get_gradient_predivide_factor(self.world_size)
         self.gradient_postdivide_factor: float = self.world_size / self.gradient_predivide_factor
-        self.gradient_predivide_factor = 1
-        self.gradient_postdivide_factor = 2.0
+        # self.gradient_predivide_factor = 1
+        # self.gradient_postdivide_factor = 2.0
         # print("XXX", self.gradient_predivide_factor, self.gradient_postdivide_factor)
 
         self.numel_padded_per_param: List[int] = []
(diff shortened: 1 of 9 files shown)


tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def create_model(with_fsdp, with_checkpoint, mixed_precision, flatten, wrap_bn, fp32_reduce_scatter):
+    model = Model()
+    if with_fsdp:
+        if wrap_bn:
+            model.block1 = auto_wrap_bn(model.block1, single_rank_pg=False)
+            model.block2 = auto_wrap_bn(model.block2, single_rank_pg=False)
+        if with_checkpoint:
+            model.block2 = checkpoint_wrapper(model.block2, maintain_forward_counter=True)
```
min-xu-ai, line 65: @QuentinDuval, once this PR is merged, you need to use mintain_forward_counter=True in vissl when forward-multiple-times and checkpoint are both used.

sshleifer (approved):
LGTM, although I don't have context on why `pre_backward_hook_has_run` used to be a list.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1062,13 +1096,27 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
+            self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST, TrainingState.IDLE])
+        else:
+            self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
         self.training_state = TrainingState.BACKWARD_POST
         if param.grad is None:
             return
+
         if param.grad.requires_grad:
-            raise RuntimeError("FullyShardedDataParallel only works with gradients that don't require grad")
+            raise RuntimeError("FullyShardedDataParallel only works with gradients that don't require gradients")
```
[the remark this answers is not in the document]
min-xu-ai: Great question! I think it came from @myleott originally. I would love to know more too. :-)

min-xu-ai:
> 
> 
> LGTM. Note that I don't have context on why `pre_backward_hook_has_run` used to be a list.

Thanks for the review @sshleifer! The `pre_backward_hook_has_run` being a list isn't originally from me, but I think I know why it is that way.

Say for this example:

```
def func():
   x = 0
   def nest():
       x = 1
```

The second assignment to `x` will be to a new local variable inside the `nest` function. However, if you turn it into:

```
def func():
   x = [ 0 ]
   def nest():
      x[0] = 1
```

Then, the second assignment will be a read of the outer `x`, which ensures the right variable is updated. It is a very clever trick to turn a write into a read first so that python knows which variable you are referring too. Using python's keyword `nonlocal` would get the same effect, like:

```
def func():
  x = 0
  def nest():
     nonlocal x
     x = 1
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -953,16 +985,18 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
-        backward. Hooks should be attached to all outputs from the forward."""
+        backward. Hooks should be attached to all outputs from the forward.
+
+        Returns:
+            outputs: new outputs with hooks registered if they requires gradient.
+        """
         if not torch.is_grad_enabled():
             return outputs  # don't register hooks if grad isn't enabled
 
-        pre_backward_hook_has_run = [False]
```
min-xu-ai, line 960: originally, the variable is local to `_register_pre_backward_hooks` function and it only prevents callbacks registered within this function. When there are multiple calls to this function, this local flag won't prevent multiple callbacks. Moving this flag to `self` solves this problem.

min-xu-ai:
I have merged this for now @myleott. Definitely happy to address more comments separately.

PR #622, opened 2021-04-22
Fixing logging, changing info to debug to avoid clutter

girifb:
Fixing logging, changing info to debug to avoid clutter


min-xu-ai (approved):
I guess for big models, this init is called many many times. :-)

PR #624, opened 2021-04-22
[fix] mypy and flaky test

min-xu-ai:
- CI didn't seem to catch this or maybe I merged incorrectly yesterday
- this should fix the mypy error on master
- also updated a test that seems to be flaky due to tcp port conflict

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/tests/nn/data_parallel/test_fsdp_input.py
+++ b/tests/nn/data_parallel/test_fsdp_input.py
@@ -34,7 +34,7 @@ def temp_files():
         rmf(name)
 
 
-# We only test on GPU since mix-precision only really works on GPU.
+# We only test on GPU since mix-precision only works on GPU.
 @skip_if_no_cuda
 @pytest.mark.parametrize(
     "fsdp_config", [{}, {"mixed_precision": True}],
--- a/tests/nn/data_parallel/test_fsdp_input.py
+++ b/tests/nn/data_parallel/test_fsdp_input.py
@@ -9,8 +9,7 @@
 
 """ Test FSDP with different input types. """
 
-import os
-import random
+import tempfile
 
 import pytest
 import torch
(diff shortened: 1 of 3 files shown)


tests/nn/data_parallel/test_fsdp_input.py
```diff
@@ -28,53 +40,45 @@
...
-            else:
-                assert input_cls is dict
-                in_data = {"in": in_data}
 
-            out = model(in_data)
-            out.sum().backward()
-            optim.step()
-            optim.zero_grad()
+    world_size = 1
+    result = dist_init(0, world_size, temp_files[0], temp_files[1])
```
anj-s, line 47: nit: inline world_size? since it is not used anywhere else

tests/nn/data_parallel/test_fsdp_input.py
```diff
@@ -28,53 +40,45 @@
...
+                input = input[0]
+            else:
+                assert isinstance(input, dict), input
+                input = input["in"]
+            return self.layer(input)
+
+    model = FSDP(Model(), **fsdp_config).cuda()
+    optim = SGD(model.parameters(), lr=0.1)
+
+    for _ in range(5):
```
anj-s, line 68: why do we need to do this 5 times? why not just once?

anj-s (approved):

tests/nn/data_parallel/test_fsdp_input.py
min-xu-ai: Thanks for the quick review! done.

tests/nn/data_parallel/test_fsdp_input.py
min-xu-ai: main reason is to (1) cover the difference between 1st iter and 2nd iter and (2) run a few more so let error accumulate a bit to trigger numerical mismatch error -- too few iters may not be enough to detect errors. It is ad hoc.

anj-s: Ah I see. Maybe I missed it but I did not see us asserting on anything other than input type so I thought a single iteration would be sufficient.

min-xu-ai: You are right, for this particular test, 5 is likely not needed. Must be a copy paste from somewhere else. Since it is very fast already, mind I just leave as is?

PR #625, opened 2021-04-22
[SDP][Minor] Removing an assert which does not seem always accurate

blefaudeux:
# Before submitting

## What does this PR do?
Fixes https://fb.workplace.com/groups/pytorchLightning/permalink/1432835657052968/
I'm not super clear on why this happens, but the general thinking is that this was a hard guard on not loosing any gradient, and actually the user may want to do to whatever (like, but may not be all of it: no_sync()fw/bw, then .eval()).
Keep the warning so that it's somewhat visible

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1664,9 +1664,7 @@ def auto_wrap_bn(module: nn.Module, single_rank_pg: bool = False, process_group:
         if recurse:
             return not isinstance(module, tuple(default_auto_wrap_policy.FORCE_LEAF_MODULES))  # type: ignore
         else:
-            return is_bn and not isinstance(
-                module, tuple(default_auto_wrap_policy.EXCLUDE_WRAP_MODULES)
-            )  # type: ignore
+            return is_bn and not isinstance(module, tuple(default_auto_wrap_policy.EXCLUDE_WRAP_MODULES))  # type: ignore
 
     pg = None
     if single_rank_pg:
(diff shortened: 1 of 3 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1664,9 +1664,7 @@ def wrap_bn_only_policy(module: nn.Module, recurse: bool, unwrapped_params: int)
         if recurse:
             return not isinstance(module, tuple(default_auto_wrap_policy.FORCE_LEAF_MODULES))  # type: ignore
         else:
-            return is_bn and not isinstance(
-                module, tuple(default_auto_wrap_policy.EXCLUDE_WRAP_MODULES)
-            )  # type: ignore
+            return is_bn and not isinstance(module, tuple(default_auto_wrap_policy.EXCLUDE_WRAP_MODULES))  # type: ignore
```
blefaudeux, line 1667: unrelated, but master linting was broken because of that

min-xu-ai (approved):
warning makes sense.

tests/nn/data_parallel/test_sharded_ddp_features.py
```diff
@@ -288,6 +288,14 @@ def test_train_eval_change():
     dist.destroy_process_group()
 
 
+def test_train_eval_change():
+    world_size = 4
```
blefaudeux, line 292: making the test case truly distributed, but still not seeing the issue which was reported in a somewhat similar situation. My guess is that there are some changes on the lightning side which make it a little more than just .train()/.eval(), but crashing on users is probably not a good idea in that case. Worst case is a lost step() worth of gradients

PR #629, opened 2021-04-26
[chore] Adding some profiling labels to OSS

blefaudeux:
# Before submitting

## What does this PR do?
Adds labels to the different steps within OSS

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/optim/oss.py
+++ b/fairscale/optim/oss.py
@@ -11,6 +11,7 @@ from math import inf
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union
 
 import torch
+from torch.autograd import profiler
 import torch.distributed as dist
 from torch.nn import Parameter
 from torch.optim import SGD, Optimizer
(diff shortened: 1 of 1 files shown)


min-xu-ai (approved):
Looks great assuming no logical changes.

fairscale/optim/oss.py
```diff
@@ -166,19 +167,22 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
         OSS._sync_param_groups(self.param_groups, self.optim.param_groups)
 
         # Catch a possible change of devices in between OSS construction and step()
-        if self._default_device.type != self.param_groups[0]["params"][0].device.type:
-            logging.info("OSS detected that the parameter changed devices, re-allocating buffers")
-            self._clear_cache()
-            self.refresh_trainable()
+        with profiler.record_function("oss_refresh_trainable"):
```
anj-s, line 170: Do you want to add the fairscale namespace similar to [offload](https://github.com/facebookresearch/fairscale/blob/5cddaea485e9d8aa46f4e5ed440d2677939ae55f/fairscale/experimental/nn/offload.py#L168)? I think it will be clearer to the user that this is an internal record_function call.

blefaudeux: that was the idea behind "oss", kind of a namespace. I actually think that a long name is not a good idea in a trace, it will probably be unreadable

blefaudeux: This is a typical trace, I would use the smallest possible label to try to keep it readable

Screen Shot 2021-04-22 at 5 10 24 PM

anj-s: we probably want see the entire label anyway unless you expand the trace and once you do you will see the entire label(tested this with offload). My suggestion is to  standardize how we want to add these labels within FairScale in way that the user is able to differentiate between their own record_function labels and the ones we add.

blefaudeux: I'm personally not sold on the added value of `fairscale.experimental.nn.`, people looking into a trace probably already have an idea of what they're looking at. It does make sure that the blocks are not identifiable from a distance, same in the above trace (the`xformers` labels on some blocks are not that useful in my view), requiring people to dig in to get a readable label does not look like a usability win ?

blefaudeux: For instance, the aten labels above just have the "aten::name" format, and I think that works well

fairscale/optim/oss.py
```diff
@@ -166,19 +167,22 @@ def step(self, closure: Optional[Callable[[], float]] = None, **kwargs: Any) ->
         OSS._sync_param_groups(self.param_groups, self.optim.param_groups)
 
         # Catch a possible change of devices in between OSS construction and step()
-        if self._default_device.type != self.param_groups[0]["params"][0].device.type:
-            logging.info("OSS detected that the parameter changed devices, re-allocating buffers")
-            self._clear_cache()
-            self.refresh_trainable()
+        with profiler.record_function("fairscale::oss::refresh_trainable"):
```
blefaudeux, line 170: @anj-s would that be ok ?

anj-s: I am going to modify offload to match this since it is definitely more succinct. Thanks!

anj-s (approved):

fairscale/optim/oss.py
blefaudeux: thanks to you @anj-s, it was a good call I think, some middle ground and following what others are doing, feels like a good outcome

blefaudeux:
oops, not sure why this closed

Issue #627, opened 2021-04-23
FSDP memory utilization issue 


prigoyal:
Reporting the issue here:

We observe weird memory usage when using FSDP. 

**Setup:** 1.5B params model, no activation checkpointing used

**Observations are:**
- **DDP, 1gpu, 1img/gpu , Pytorch AMP , memory looks like:**
    - 5480 (before fwd) -> 9547 (after fwd) -> 10960 (after bwd) -> 16384 (after optimizer step and params update)
    - verified with PyTorch 1.6 and PyTorch 1.9.0 both have same

- **FSDP (newest integration), 2gpu, 1img/gpu, compute_dtype=fp32 and equivalent to O1 settings, memory looks like**
    - 2739 (before fwd) -> 7416 (after fwd) -> **11190 (after bwd)** -> 12599 (after optimizer step and params update)
    - verified with Pytorch 1.6 and Pytorch 1.9.0 both
    - **What's wrong**: On backward we would expect the memory to be 2739 (params sharded) + 2739 (gradients sharded) but instead we see 11190

- **FSDP (integration from March beginning that was studied for perf), 2gpu, 1img/gpu, compute_dtype=fp32 and equivalent to O1 settings, memory looks like**
    - 274 (before fwd) -> 7412 (after fwd) -> **5488 (after bwd)** -> 5774 (after optimizer step and params update)
    - Pytorch 1.9.0 from March

**_Something has changed in the newest integration and after backward pass with FSDP, we don't see the expected memory_**.

**Things we have ruled out:**

- [x] Pytorch memory fragmentation: 
- [x] Pytorch version issue
- [x] Issue with FSDP -> **CONFIRMED**
- [x] Issue in our integration

# Some more details on FSDP vs DDP memory consumption

## In the context of VISSL

We have been noticing that for RegNet16, the memory consumption with FSDP is significantly higher than DDP. The key differences happen in iteration 0. All the data below is obtains on a RegNet16, trained on 8 GPUs for a few epochs with a batch size of 2 (small batch size to limit the contribution of activation memory).

### At iteration 0

Here is a graph of the memory consumption during forward/backward **in DPP**:

<img width="523" alt="Screenshot 2021-04-23 at 11 44 01" src="https://user-images.githubusercontent.com/7412790/115896278-400adc00-a429-11eb-835f-80308a58fb47.png">

Here is a graph of the memory consumption during forward/backward **in FSDP**:

[...]

PR #633, opened 2021-04-27
[feat] save memory by using bucket buffer only in backward

min-xu-ai:
- this fixes bug #627
- added documentation to clarify the buffer's cost and speed/memory
  tradeoff
- added setup/teardown calls so that the buffer is only allocated
  during the backward pass, saving more memory for forward and stepping
  so that they can be used for things like activations.
- added a unit test that assert the memory is in range.

Comparing with DDP:

  1. buffer size scales with # of FSDP not model size
  2. buffer is only allocated during backward
  3. buffer is used for small tensors only to reduce overhead
  4. overlapping of compute-reduction is very different

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## NEXT - TBD
+### Added
+- FSDP: better memory usage for reduce bucket
 
 ## [0.3.6] - 2021-04-26
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## NEXT - TBD
 ### Added
-- FSDP: better memory usage for reduce bucket
+- FSDP: better memory usage for reduce bucket ([#633](https://github.com/facebookresearch/fairscale/pull/633))
 
 ## [0.3.6] - 2021-04-26
 ### Added
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -155,10 +155,30 @@ class FullyShardedDataParallel(nn.Module):
             *``cpu_offload``*.
         bucket_cap_mb (int, Optional):
             FSDP will bucket parameters so that gradient reduction can
-            potentially overlap with backward computation. bucket_cap_mb
-            controls the bucket size in MegaBytes (MB). Buckets are sub-divided
-            based on world_size, so the max shard size is roughly
-            ``bucket_cap_mb / world_size``. Values <= 0 disable bucketing.
+            be more efficient for small tensors.
+            ``bucket_cap_mb`` controls the bucket size in MegaBytes (MB). Buckets
+            are sub-divided based on world_size, so the max shard size is roughly
+            ``bucket_cap_mb / world_size``. When nested FSDP is used,
+            each FSDP instance will have a separate set of buckets (1 per bucket
+            per tensor <type, device, process group> tuple). Large gradient tensors
+            are directly reduced without using the buffer. The buffer is there to reduce
+            communication overhead for small tensors. Overlapping with computation
+            happens due to use of a different CUDA stream than the computation CUDA
+            stream. The total memory overhead per buffer is around
+            ``bucket_cap_mb / world_size * (world_size + 1)``.
+            The buffers are allocated during the backward pass and freed at the end
+            of the backward pass to save more memory for other phases of the
+            training process.
+            Note, the memory vs. speed tradeoff of bucket size is very different
+            from that of the DDP engine. In DDP, the buffer size ``1MB + n*cap_mb``,
+            until n is big enough to cover the entire model size. The order
+            of which buffer is ready there is more rigid and DDP requires all
+            gradients to be computed in the backward. In FSDP, the buffer
+            does not change with model size (it scales per # of FSDP instances)
+            and gradient ready order matters little and we have a final flush
+            call that ensures everything is reduced and not all gradients need
+            to be upfront known. Overlapping with compute is done differently too.
+            Values <= 0 disable bucketing.
             Default: 25.
         compute_device (torch.device, Optional):
             device for computation. If not given and module params are on a CUDA
@@ -1739,6 +1759,8 @@ def auto_wrap_bn(module: nn.Module, single_rank_pg: bool = False, process_group:
         # **must** be False because BN's FSDP wrapper's pre-backward callback isn't called
         # within the checkpoint's outer backward when multiple forward passes are used.
         "reshard_after_forward": False,
+        # No bucketing or small bucketing should be enough for BNs.
+        "bucket_cap_mb": 0,
     }
 
     with enable_wrap(wrap_bn_only_policy, **fsdp_config):
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1246,15 +1246,6 @@ class FullyShardedDataParallel(nn.Module):
         else:
             self.assert_state(TrainingState.BACKWARD_PRE)
 
-        def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
-            """Helper used below on all fsdp modules."""
-            for p in fsdp_module.params:
-                if p.requires_grad:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## NEXT - TBD
+### Added
+- FSDP: better memory usage for reduce bucket ([#633](https://github.com/facebookresearch/fairscale/pull/633))
```
min-xu-ai, line 9: @blefaudeux, one thing tricky here is that when I was modifying the changelog file, the PR number wasn't available. I had to come back and edit this file again. :-)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -155,10 +155,30 @@ class FullyShardedDataParallel(nn.Module):
             *``cpu_offload``*.
         bucket_cap_mb (int, Optional):
             FSDP will bucket parameters so that gradient reduction can
-            potentially overlap with backward computation. bucket_cap_mb
-            controls the bucket size in MegaBytes (MB). Buckets are sub-divided
-            based on world_size, so the max shard size is roughly
-            ``bucket_cap_mb / world_size``. Values <= 0 disable bucketing.
+            be more efficient for small tensors.
```
min-xu-ai, line 158: I try to document completely here. Please let me know if this is too much and/or inaccurate.

fairscale/utils/reduce_scatter_bucketer.py
```diff
@@ -37,6 +38,24 @@ def flush(self) -> None:
         self.callbacks.clear()
         self.output_shard = torch.zeros_like(self.data[0])
 
+    def setup(self) -> None:
```
min-xu-ai, line 41: This two functions enables some memory saving outside of backward pass.

tests/nn/data_parallel/test_fsdp_memory.py
```diff
@@ -0,0 +1,243 @@
...
+        fake_loss = criterion(out, torch.tensor(0.0).cuda())
+        get_cur_mem(gpu_id, results, f"iter {iteration}: after loss")
+
+        fake_loss.backward()
+        get_cur_mem(gpu_id, results, f"iter {iteration}: after bwd")
+
+        optimizer.step()
+        get_cur_mem(gpu_id, results, f"iter {iteration}: after step")
+
+        # It is important to use the loop below, not optimizer.zero_grad() to reclaim memory.
```
min-xu-ai, line 141: This seems to be very important for FSDP. Without it FSDP end up with >60MB after the first iter and DDP ends up with around 12MB. I think it is good for everyone to do it this way anyway, so I don't think we need to fix this in FSDP.

prigoyal: just a minor comment: we usually do `optimizer.zero_grad() -> optimizer.step()` in this order not the other way around.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -155,10 +155,30 @@ class FullyShardedDataParallel(nn.Module):
...
+            ``bucket_cap_mb`` controls the bucket size in MegaBytes (MB). Buckets
+            are sub-divided based on world_size, so the max shard size is roughly
+            ``bucket_cap_mb / world_size``. When nested FSDP is used,
+            each FSDP instance will have a separate set of buckets (1 per bucket
+            per tensor <type, device, process group> tuple). Large gradient tensors
```
myleott, line 163: The setup/teardown functionality looks good.

But I thought we share the ReduceScatterBucketer instance across all nested FSDP instances (similar to how we share streams), so I thought there shouldn't be separate buckets for separate FSDP instances: https://github.com/facebookresearch/fairscale/blob/4dce2d62df4b8aeee35ff156cb83bfa6e6d1c970/fairscale/nn/data_parallel/fully_sharded_data_parallel.py#L938-L940

tests/nn/data_parallel/test_fsdp_memory.py
```diff
@@ -0,0 +1,243 @@
...
+        for p in model.parameters():
+            p.grad = None
```
myleott, line 143: nit: you can also do `model.zero_grad(set_to_none=True)`

myleott (approved):
Looks good, and nice test!

fairscale/utils/reduce_scatter_bucketer.py
```diff
@@ -131,6 +150,7 @@ def flush(self) -> None:
         """Reduce-scatter any partial buckets."""
         for bucket in self.buckets.values():
             bucket.flush()
+            bucket.teardown()
```
myleott, line 153: also, maybe we don't want to teardown on every flush? We might have lots of intermediate flush's during the backward pass, but we probably only want to teardown at the end. Perhaps expose teardown as a separate API and call it in the final backward hook in FSDP?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: Thanks for catching this! My comments are certainly incorrect.

This is very subtle. I commented out the the teardown and the memory for the unit test became:
```
{
    "iter 0: start": 3,
    "iter 0: after fwd": 340,
    "iter 0: after loss": 340,
    "iter 0: after bwd": 258,
    "iter 0: after step": 258,
    "iter 0: done": 196,
    "iter 1: start": 196,
    "iter 1: after fwd": 533,
    "iter 1: after loss": 533,
    "iter 1: after bwd": 258,
    "iter 1: after step": 258,
    "iter 1: done": 196,
    "iter 2: start": 196,
    "iter 2: after fwd": 533,
    "iter 2: after loss": 533,
    "iter 2: after bwd": 258,
    "iter 2: after step": 258,
    "iter 2: done": 196,
}
```
vs. with `teardown`
```
{
    "iter 0: start": 3,
    "iter 0: after fwd": 340,
    "iter 0: after loss": 340,
    "iter 0: after bwd": 66,
    "iter 0: after step": 66,
    "iter 0: done": 3,
    "iter 1: start": 3,
    "iter 1: after fwd": 340,
    "iter 1: after loss": 340,
    "iter 1: after bwd": 66,
    "iter 1: after step": 66,
    "iter 1: done": 3,
    "iter 2: start": 3,
    "iter 2: after fwd": 340,
    "iter 2: after loss": 340,
    "iter 2: after bwd": 66,
    "iter 2: after step": 66,
    "iter 2: done": 3,
}
```

It increased by 193MB, which close to 5 * (25 + 12.5) = 187.5. The `5` is actually number of different process groups instances used by the model.

The test is already using a global PG. But `auto_wrap_bn` was not using the right global PG.

It seems we should index the buckets dict using a key that represent the process group with its members in this line:

https://github.com/facebookresearch/fairscale/blob/9b79cc020981a9b6e0b13570ede36a080a69c7b5/fairscale/utils/reduce_scatter_bucketer.py#L144

But I can't seem to find an API of the process group that can turn it into its members. Ideas? Maybe leave it as is?

tests/nn/data_parallel/test_fsdp_memory.py
min-xu-ai: done.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: I updated the comment but hasn't change the `key = ...` line.

min-xu-ai: I will merge this as it with a TODO here once the CI passes so that vissl can use this.

tests/nn/data_parallel/test_fsdp_memory.py
min-xu-ai: see below. we can use model.zero_grad(set_to_none=True) as well. I recommended `optimizer.step() -> model.zero_grad()` since this will free the memory sooner than the other way around.

fairscale/utils/reduce_scatter_bucketer.py
```diff
@@ -166,6 +166,10 @@ def _get_shard_size(self, element_size: int, num_shards: int) -> int:
         return int(bucket_size // num_shards)
 
     def _get_bucket(self, tensor: Tensor, group: ProcessGroup) -> Bucket:
+        # TODO (Min): the `group` used here in the key is the object hash, not the content
+        #     hash. That means if FSDP instances are initialized with different process groups,
+        #     even when the group members are in fact the same, we end up creating different
+        #     buckets here.
```
myleott, line 172: oh, I see, this is a good point!

sshleifer:
Just read this, nice catch and very educational docs!

PR #634, opened 2021-04-27
[chore] do not build cuda extensions by default

msbaines:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
     "setuptools >= 40.6.2",
-    "torch >= 1.4.0",
     "wheel >= 0.30.0"
 ]
 build-backend = "setuptools.build_meta:__legacy__"
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,6 @@ import re
 import warnings
 
 import setuptools
-import torch
-from torch.utils.cpp_extension import CUDA_HOME, BuildExtension, CUDAExtension
 
 this_dir = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -34,8 +32,10 @@ def find_version(version_file_path):
 extensions = []
 cmdclass = {}
 
-force_cuda = os.getenv("FORCE_CUDA", "0") == "1"
-if (torch.cuda.is_available() and CUDA_HOME is not None) or force_cuda:
+build_cuda_extensions = os.getenv("BUILD_CUDA_EXTENSIONS", "0") == "1"
```
min-xu-ai, line 35: This is very nice! Can you also check whether it will download the latest version of torch during the pip install? Last time I checked, it always download and disregard the torch version I already have in my env.

setup.py
```diff
@@ -34,8 +32,10 @@ def find_version(version_file_path):
...
+if build_cuda_extensions:
+    from torch.utils.cpp_extension import BuildExtension, CUDAExtension
+
```
anj-s, line 38: Thank you for the change! Should we modify the warning statement in the else: clause?

does it make sense to still check for cuda and CUDA_HOME?  Looks like we will require users to set the environment var: BUILD_CUDA_EXTENSIONS if they want to use fused_adam. It's not being used right now so maybe its ok to leave it as is?

anj-s (approved):

setup.py
anj-s: Also, do we need to modify the changelog?

setup.py
min-xu-ai: I wasn't clear, but I thought always downloading is not good for the users since some may have slow internet and downloading 1GB will take a long time for no good reason.

setup.py
msbaines: > Thank you for the change! Should we modify the warning statement in the else: clause?
> 
> does it make sense to still check for cuda and CUDA_HOME? Looks like we will require users to set the environment var: BUILD_CUDA_EXTENSIONS if they want to use fused_adam. It's not being used right now so maybe its ok to leave it as is?

It feels simpler just to fail if the user sets BUILD_CUDA_EXTENSIONS but then does not have CUDA_HOME set.

pyproject.toml
```diff
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
     "setuptools >= 40.6.2",
-    "torch >= 1.4.0",
```
min-xu-ai, line 4: awesome!

min-xu-ai (approved):
Thanks again!

PR #635, opened 2021-04-27
[feat]Adding DynamicLossScaler class for supporting optimizer updates on the CPU

tmarkstrum:
# Before submitting

## What does this PR do?
Added dynamic loss scaler to avoid gradient overflow/underflow by scaling the loss.
The DynamicLossScaler can be used in the same manner with GradScaler except that it can be applied to cup devices. https://pytorch.org/docs/stable/notes/amp_examples.html#amp-examples

## PR review
cc @blefaudeux @msbaines Please feel free to review the code.

## Did you have fun?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     rev: 19.10b0
     hooks:
     - id: black
-      language_version: python3.8
+      language_version: python3.7
 
 -   repo: https://gitlab.com/pycqa/flake8
     rev: 3.7.9
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     rev: 19.10b0
     hooks:
     - id: black
-      language_version: python3.7
+      language_version: python3.8
 
 -   repo: https://gitlab.com/pycqa/flake8
     rev: 3.7.9
--- a/fairscale/experimental/optim/dynamic_loss_scaler.py
+++ b/fairscale/experimental/optim/dynamic_loss_scaler.py
@@ -11,6 +11,7 @@ dynamically scale up and down gradients by scaling the loss.
 
 from collections import defaultdict
 from enum import Enum
+from typing import Dict, List, Optional
 
 import torch
 
@@ -21,7 +22,7 @@ class OptState(Enum):
     STEPPED = 2
 
 
-def _refresh_per_optimizer_state():
+def _refresh_per_optimizer_state() -> OptState:
     return OptState.READY
 
 
@@ -36,12 +37,12 @@ class DynamicLossScaler(object):
 
     def __init__(
         self,
-        init_scale=2.0 ** 15,
-        scale_factor=2.0,
-        scale_window=2000,
-        tolerance=0.0,
-        threshold=None,
-        min_loss_scale=1e-4,
+        init_scale: float = 2.0 ** 15,
+        scale_factor: float = 2.0,
+        scale_window: int = 2000,
+        tolerance: float = 0.0,
+        threshold: float = None,
+        min_loss_scale: float = 1e-4,
     ):
         self.loss_scale = init_scale
         self.scale_factor = scale_factor
@@ -53,10 +54,10 @@ class DynamicLossScaler(object):
         self._last_overflow_iter = -1
         self._last_rescale_iter = -1
         self._overflows_since_rescale = 0
-        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+        self._per_optimizer_states: Dict[int, OptState] = defaultdict(_refresh_per_optimizer_state)
         self._scale = None
 
-    def scale(self, outputs):
+    def scale(self, outputs):  # type: ignore
         """
         Multiplies ('scales') a tensor or list of tensors by the scale factor.
 
@@ -71,7 +72,7 @@ class DynamicLossScaler(object):
         return self.loss_scale * outputs
 
     @torch.no_grad()
-    def _get_gradients_norm(self, params) -> float:
+    def _get_gradients_norm(self, params: List[torch.nn.Parameter]) -> float:
         grads = []
         for p in params:
             if p.grad is None:
@@ -80,20 +81,20 @@ class DynamicLossScaler(object):
                 grads.append(p.grad.detach())
 
         if len(grads) == 0:
-            return torch.tensor(0.0)
+            return 0.0
 
         if len(grads) == 1:
-            total_norm = torch.norm(grads[0], p=2, dtype=torch.float32)
+            total_norm = torch.norm(grads[0], p=2, dtype=torch.float32)  # type: ignore
         else:
-            total_norm = torch.norm(torch.stack([torch.norm(g, p=2, dtype=torch.float32) for g in grads]))
+            total_norm = torch.norm(torch.stack([torch.norm(g, p=2, dtype=torch.float32) for g in grads]))  # type: ignore
         return total_norm.item()
 
-    def _decrease_loss_scale(self):
+    def _decrease_loss_scale(self) -> None:
         self.loss_scale /= self.scale_factor
         if self.threshold is not None:
             self.loss_scale = max(self.loss_scale, self.threshold)
 
-    def _check_overflow(self, grad_norm):
+    def _check_overflow(self, grad_norm: float) -> None:
         # detect inf and nan
         if grad_norm == float("inf") or grad_norm != grad_norm:
             # overflow has occured
@@ -123,7 +124,7 @@ class DynamicLossScaler(object):
             self._iter += 1
             raise OverflowError("setting loss scale to: " + str(self.loss_scale))
 
-    def update(self):
+    def update(self) -> None:
         """
         Updates the scale factor.
         """
@@ -134,7 +135,7 @@ class DynamicLossScaler(object):
         self._iter += 1
         self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
 
-    def step(self, optimizer, *args, **kwargs):
+    def step(self, optimizer, *args, **kwargs):  # type: ignore
         """
         :meth:`step` unscale the gradients and step the optimizer.
 
@@ -175,7 +176,7 @@ class DynamicLossScaler(object):
 
         return retval
 
-    def unscale_(self, optimizer):
+    def unscale_(self, optimizer: torch.optim.Optimizer) -> None:
 
         optimizer_state = self._per_optimizer_states[id(optimizer)]
 
@@ -197,10 +198,10 @@ class DynamicLossScaler(object):
 
         optimizer_state = OptState.UNSCALED
 
-    def state_dict(self):
-        if self.scaler is not None:
+    def state_dict(self) -> Optional[Dict[str, float]]:
+        if self.loss_scale is not None:
             return {"loss_scale": self.loss_scale}
 
-    def load_state_dict(self, state_dict):
+    def load_state_dict(self, state_dict: Dict[str, float]) -> None:
         if "loss_scale" in state_dict:
             self.loss_scale = state_dict["loss_scale"]
--- a/fairscale/experimental/optim/dynamic_loss_scaler.py
+++ b/fairscale/experimental/optim/dynamic_loss_scaler.py
@@ -1,0 +1,169 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+from collections import abc, defaultdict
+from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple
+import warnings
+
+import torch
+
+
+class OptState(Enum):
+    READY = 0
+    UNSCALED = 1
+    STEPPED = 2
+
+
+def _refresh_per_optimizer_state():
+    return OptState.READY
+
+
+class DynamicLossScaler(object):
+    def __init__(
+        self,
+        init_scale=2.0 ** 15,
+        scale_factor=2.0,
+        scale_window=2000,
+        tolerance=0.0,
+        threshold=None,
+        min_loss_scale=1e-4,
+    ):
+        self.loss_scale = init_scale
+        self.scale_factor = scale_factor
+        self.scale_window = scale_window
+        self.tolerance = tolerance
+        self.threshold = threshold
+        self._iter = 0
+        self._last_overflow_iter = -1
+        self._last_rescale_iter = -1
+        self._overflows_since_rescale = 0
+        self.min_loss_scale = min_loss_scale
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+        self._scale = None
+
+    def scale(self, outputs):
+        return self.loss_scale * outputs
+
+    @torch.no_grad()
+    def _get_gradients_norm(self, params) -> float:
+        grads = []
+        for p in params:
+            if p.grad is None:
+                continue
+            else:   
+                grads.append(p.grad.detach())
+        
+        if len(grads) == 0:
+            return torch.tensor(0.0)
+
+        if len(grads) == 1:
+            total_norm = torch.norm(grads[0], p=2, dtype=torch.float32)
+        else:
+            total_norm = torch.norm(
+                torch.stack([torch.norm(g, p=2, dtype=torch.float32) for g in grads])
+            )
+        return total_norm.item()
+
+    def _decrease_loss_scale(self):
+        self.loss_scale /= self.scale_factor
+        if self.threshold is not None:
+            self.loss_scale = max(self.loss_scale, self.threshold)
+
+    def _check_overflow(self, grad_norm):
+        # detect inf and nan
+        if grad_norm == float("inf") or grad_norm != grad_norm:
+            # overflow has occured
+            prev_scale = self.loss_scale
+            iter_since_rescale = self._iter - self._last_rescale_iter
+
+            self._last_overflow_iter = self._iter
+            self._overflows_since_rescale += 1
+            pct_overflow = self._overflows_since_rescale / float(iter_since_rescale)
+            if pct_overflow >= self.tolerance:
+                self._decrease_loss_scale()
+                self._last_rescale_iter = self._iter
+                self._overflows_since_rescale = 0
+
+            if self.loss_scale <= self.min_loss_scale:
+                # Use FloatingPointError as an uncommon error that parent
+                # functions can safely catch to stop training.
+                self.loss_scale = prev_scale
+                raise FloatingPointError(
+                    (
+                        "Minimum loss scale reached ({}). Your loss is probably exploding. "
+                        "Try lowering the learning rate, using gradient clipping or "
+                        "increasing the batch size."
+                    ).format(self.min_loss_scale)
+                )
+
+            self._iter += 1
+            raise OverflowError("setting loss scale to: " + str(self.loss_scale))
+
+    def update(self):
+
+        if (self._iter - self._last_overflow_iter) % self.scale_window == 0:
+            self.loss_scale *= self.scale_factor
+            self._last_rescale_iter = self._iter
+        self._iter += 1
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+
+    def step(self, optimizer, *args, **kwargs):
+        if "closure" in kwargs:
+            raise RuntimeError("Closure use is not currently supported if GradScaler is enabled.")
+
+        optimizer_state = self._per_optimizer_states[id(optimizer)]
+
+        if optimizer_state is OptState.STEPPED:
+            raise RuntimeError("step() has already been called since the last update().")
+
+        # check gradient norm. If gradient norm is nan or inf, adjust scale here.
+        # clip_grads_norm can happen before this step
+        for group in optimizer.param_groups:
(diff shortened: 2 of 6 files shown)


fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+To prevent underflow or overflow of gradients, Dynamic Loss Scaler is used to
```
anj-s, line 7: s/ Dynamic Loss Scaler/DynamicLossScaler?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+    READY = 0
+    UNSCALED = 1
+    STEPPED = 2
+
+
+def _refresh_per_optimizer_state():
+    return OptState.READY
+
+
+class DynamicLossScaler(object):
```
anj-s, line 29: Shall we add a link to the file that was used as a reference for this implementation?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+        self.loss_scale = init_scale
+        self.scale_factor = scale_factor
+        self.scale_window = scale_window
+        self.tolerance = tolerance
+        self.threshold = threshold
+        self._iter = 0
+        self._last_overflow_iter = -1
+        self._last_rescale_iter = -1
+        self._overflows_since_rescale = 0
+        self.min_loss_scale = min_loss_scale
```
anj-s, line 56: should this be moved above under `self.threshold`?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+        self._scale = None
+
+    def scale(self, outputs):
+        """
+        Multiplies ('scales') a tensor or list of tensors by the scale factor.
+
+        Returns scaled outputs.
```
anj-s, line 64: There should be a Returns section for this. Take a look at python doc strings and modify it slightly.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+                raise FloatingPointError(
+                    (
+                        "Minimum loss scale reached ({}). Your loss is probably exploding. "
+                        "Try lowering the learning rate, using gradient clipping or "
+                        "increasing the batch size."
+                    ).format(self.min_loss_scale)
+                )
+
+            self._iter += 1
+            raise OverflowError("setting loss scale to: " + str(self.loss_scale))
```
anj-s, line 122: Should we still raise an Error?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+
+        Returns the return value of ``optimizer.step(*args, **kwargs)``.
+
+        Args:
+            optimizer (torch.optim.Optimizer):  Optimizer that applies the gradients.
+            args:  Any arguments.
+            kwargs:  Any keyword arguments.
+        """
+        if "closure" in kwargs:
+            raise RuntimeError("Closure use is not currently supported if GradScaler is enabled.")
```
anj-s, line 149: s/GradScaler/DynamicLossScaler

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+        if optimizer_state is OptState.STEPPED:
+            raise RuntimeError("step() has already been called since the last update().")
+
+        # check gradient norm. If gradient norm is nan or inf, adjust scale here.
+        # clip_grads_norm can happen before this step
+        for group in optimizer.param_groups:
+            grad_norm = self._get_gradients_norm(group["params"])
+            self._check_overflow(grad_norm)
+
+        retval = None
```
anj-s, line 162: do you need this?

anj-s:
@tmarkstrum I think you have to add your test file to one these .txt [files](https://github.com/facebookresearch/fairscale/blob/master/tests/ci_test_list_3.txt). There are three of them and are fairly even in terms of runtime. Once you add them, they will run on circleCI.

min-xu-ai:
@tmarkstrum Thanks for this PR! Before I review it in details, I have several high level questions:

1. What's the relationship of this scalar with the existing one from Ben called sharded scalar?
2. Perhaps we should have some discussion first in an issue about the purpose and design for a new class like this?
3. Why is this in experimental? If something is likely going to be used without much unknown of the algorithm's usefulness, perhaps it is not very experimental. Do you see this introduce a new algorithm that needs to be validated by researchers?
4. I see the CPU tests fails. Perhaps you can use this alias `alias fs_test='black . && isort . && flake8 && mypy --ignore-missing-imports --scripts-are-modules --pretty . && echo ALL GOOD'` before make a PR non-draft and ready to be reviewed? It will be good to review PR only after CI is passing or going to pass.
5. Finally, perhaps update the changelog file in the PR too?

tests/experimental/optim/test_dynamic_loss_scaler.py
```diff
@@ -0,0 +1,73 @@
...
+    idx = np.arange(100)
+    np.random.shuffle(idx)
+    # Generates train sets
+    x_train, y_train = x[idx], y[idx]
+    x_train_tensor = torch.tensor([x_train]).float().to(device)
+    y_train_tensor = torch.tensor([y_train]).float().to(device)
+    return x_train_tensor, y_train_tensor
+
+
+def _train_with_dls(x, y):
```
anj-s, line 45: How do we know if we are hitting Nans and scaling the loss accordingly?

blefaudeux:
> @tmarkstrum Thanks for this PR! Before I review it in details, I have several high level questions:
> 
> 1. What's the relationship of this scalar with the existing one from Ben called sharded scalar?
The sharded grad scaler is a wrap around pytorch's, so that it's shard aware. It does not change its logic, just adds a reduction on the world so that all of the grad scalers have the same picture. 

Pytorch's grad scaler is meant to complement AMP, which is CUDA specific. I'm just starting on this PR but to my understanding this is a more generic take ?

> 4. I see the CPU tests fails. Perhaps you can use this alias `alias fs_test='black . && isort . && flake8 && mypy --ignore-missing-imports --scripts-are-modules --pretty . && echo ALL GOOD'` before make a PR non-draft and ready to be reviewed? It will be good to review PR only after CI is passing or going to pass.

Just for general awareness, there's a tool called [pre-commit](https://pre-commit.com/) which does that for you, and this repo is compatible with it, it has been a personal life saver when combined with a dedicated env for this repo (pytorch is also using this)

fairscale/experimental/optim/dynamic_loss_scaler.py
blefaudeux: +1, easier for the future us and for readers to get some context

min-xu-ai:
Yeah, pre-commit is even better. I personally just more used to do it by hand when I wanted it to run. By all means use pre-commit to automate, @tmarkstrum.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+        self.threshold = threshold
+        self._iter = 0
+        self._last_overflow_iter = -1
+        self._last_rescale_iter = -1
+        self._overflows_since_rescale = 0
+        self.min_loss_scale = min_loss_scale
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+        self._scale = None
+
+    def scale(self, outputs):
```
blefaudeux, line 60: nit, but can we try to type everything @tmarkstrum ? I know it clumsifies a bit the python syntax, but in my experience it does help when using the codebase from a distance, and modern IDEs are making good use of this

blefaudeux: incidentally I think that it removes the need for some details in the documentation, since it's already there in the code, so two birds & one stone so to say

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,203 @@
...
+            total_norm = torch.norm(torch.stack([torch.norm(g, p=2, dtype=torch.float32) for g in grads]))
+        return total_norm.item()
+
+    def _decrease_loss_scale(self):
+        self.loss_scale /= self.scale_factor
+        if self.threshold is not None:
+            self.loss_scale = max(self.loss_scale, self.threshold)
+
+    def _check_overflow(self, grad_norm):
+        # detect inf and nan
```
blefaudeux, line 95: could you elaborate a little in the docstring about the plan here ? it looks like this can raise exceptions, but seeing where it's called it does not look like it's completely the plan, I was wondering whether this was on purpose ? The [torch AMP scaler ](https://github.com/pytorch/pytorch/blob/master/torch/cuda/amp/grad_scaler.py)does not raises too aggressively, so I'm wondering whether there's a philosophical difference that I missed ?

blefaudeux:
Thanks @tmarkstrum for the PR ! I'm probably missing some context, but I would love a master plan somewhere (in the main docstring ?) to explain the expected behaviour, right now it's a bit hard to know if every path does what it's supposed to do, at least for me ? else generally looks good to me !

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: Do you mean a link about how to use the scaler? I can comment as "Use the DynamicLossScaler similar with GradScaler except it can be applied to cup device. https://pytorch.org/docs/stable/notes/amp_examples.html#amp-examples

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: I need to catch the Error in the other function. Let me fix it.

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: not really.

msbaines:
The commit message is lacking detail. For ease of review, please describe what this change does and why it is needed.

Also please follow the format documented here:

https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: I added the line to catch the exception in the step function in the 6th commit. When the exception is raised, it means there is INF or NAN in gradients, the optimizer step will be skipped.

tmarkstrum:
The pre-commit need python3.7. My  conda vm python is 3.8. Does anyone have similar issue?

anj-s:
> The pre-commit need python3.7. My conda vm python is 3.8. I tried fs_test, Min's command. It also caught errors from other files subs/torch/mm/modules/adaptive.pyi. benchmarks/experimental/experiemntal_async_approaches.py. Does anyone notice the similar issues?

some of the errors that are caught by mypy (using fs_test) don't cause the test to fail. However it is good practice to make sure you did not add to the list :) 
What is the error that you are seeing with pre-commit? The [yaml config](https://github.com/facebookresearch/fairscale/blob/master/.pre-commit-config.yaml) specified python 3.7 for black. I wonder if that is what is causing the error. Can you try bumping it up to 3.9 and seeing if that fixes it for you?

tests/experimental/optim/test_dynamic_loss_scaler.py
tmarkstrum: The test cases did not cover it. Do you have any examples of model training that generates the Nans/Inf gradients in the process?

fairscale/experimental/optim/dynamic_loss_scaler.py
anj-s: I meant the actual GradScaler class whose implementation mirrors this except for how we detect Nans.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,207 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+To prevent underflow or overflow of gradients, DynamicLossScaler is used to
+dynamically scale up and down gradients by scaling the loss.
```
anj-s, line 8: Will be good to mention what is different about this from the [torch.cuda.amp.GradScaler](https://pytorch.org/docs/stable/_modules/torch/cuda/amp/grad_scaler.html#GradScaler) class - i.e it needs to be used for updates on the CPU device

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,207 @@
...
+
+def _refresh_per_optimizer_state() -> OptState:
+    return OptState.READY
+
+
+class DynamicLossScaler(object):
+    """An instance ``scaler`` helps perform the steps of gradient scaling
+    conveniently.
+
+    * ``scaler.scale(loss)`` multiplies a given loss by ``scaler``'s current scale factor.
```
anj-s, line 33: We generally don't specify the methods in the doc string.  The  python doc string usually has the following format:
```
single line comment

multiple line comments

Args:

Returns:

Raises:
```

fairscale/experimental/optim/dynamic_loss_scaler.py
anj-s: Remove the Returns line?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,207 @@
...
+        self._iter += 1
+        self._per_optimizer_states = defaultdict(_refresh_per_optimizer_state)
+
+    def step(self, optimizer, *args, **kwargs):  # type: ignore
+        """
+        :meth:`step` unscale the gradients and step the optimizer.
+
+        ``*args`` and ``**kwargs`` are forwarded to ``optimizer.step()``.
+
+        Returns the return value of ``optimizer.step(*args, **kwargs)``.
```
anj-s, line 144: Fix Returns section here. Also we end up returning None. is that right?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,207 @@
...
+            state_dict["loss_scale"] = self.loss_scale
+
+        retval = optimizer.step(*args, **kwargs)
+
+        optimizer_state = OptState.STEPPED
+
+        return retval
+
+    def unscale_(self, optimizer: torch.optim.Optimizer) -> None:
+
```
anj-s, line 180: missing doc string?

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,207 @@
...
+            for group in optimizer.param_groups:
+                for param in group["params"]:
+                    if param.grad is None:
+                        continue
+                    else:
+                        param.grad.data.mul_(inv_scale)
+
+        optimizer_state = OptState.UNSCALED
+
+    def state_dict(self) -> Optional[Dict[str, float]]:
```
anj-s, line 201: Lets add a test for this?

tests/experimental/optim/test_dynamic_loss_scaler.py
```diff
@@ -0,0 +1,75 @@
...
+    for name, param in model.named_parameters():
+        if param.requires_grad:
+            print(name, param.data)
+            if name == "linear.weight":
+                assert (param.data.item() - 2) <= 0.05
+            if name == "linear.bias":
+                assert (param.data.item() - 1) <= 0.03
+
+
+# TODO: add test case covering check_overflow function
```
anj-s, line 75: s/TODO/TODO(<your github id>):

tests/experimental/optim/test_dynamic_loss_scaler.py
```diff
@@ -0,0 +1,75 @@
...
+        model.train()
+        yhat = model(x)
+        loss = loss_fn(y, yhat)
+        scaler.scale(loss).backward()
+        scaler.step(optimizer)
+        scaler.update()
+    return model
+
+
+def test_dls():
```
anj-s, line 63: Both the test names seem very similar. can we rename one to reflect what it is testing. We also need to test FP16 so perhaps add a TODO to tackle that in a followup PR?

anj-s:
Thanks you for the PR! Looks great and is almost there.

High level comment about the PR description: Title should follow the format as mentioned in the contributing guidelines that @msb pointed to. For example, the title here should be 
"[feat]Adding DynamicLossScaler class for supporting optimizer updates on the CPU."
Also populate the PR description box as needed. I know you've made some changes but it needs more details for someone who does not have context.

.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
     rev: 19.10b0
     hooks:
     - id: black
-      language_version: python3.7
+      language_version: python3.8
```
msbaines, line 23: This should not change.

tests/experimental/optim/test_dynamic_loss_scaler.py
msbaines: I usually add my name inside the TODO. That way I can easily grep for my TODOs by searching for `TODO(msb)`

tests/experimental/optim/test_dynamic_loss_scaler.py
```diff
@@ -0,0 +1,75 @@
...
+    return x_train_tensor, y_train_tensor
+
+
+def _train_with_dls(x, y):
+    scaler = DynamicLossScaler()
+    torch.manual_seed(42)
+    lr = 1e-1
+    n_epochs = 1000
+    loss_fn = nn.MSELoss(reduction="mean")
+    model = ManualLinearRegression().to(device)
```
msbaines, line 50: This doesn't appear to being using float16. To test the loss scaler, shouldn't the model be float16?

msbaines (commented):
Might be easier for the user to call this GradScaler and implement the same interface so that it is a drop-in replacement.

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: I referred to https://pytorch.org/docs/stable/_modules/torch/cuda/amp/grad_scaler.html#GradScaler as a template. Is there any  wiki or doc that I can follow for this codebase?

fairscale/experimental/optim/dynamic_loss_scaler.py
tmarkstrum: This is internal method. Doc string is not necessary. I can add some comments here.

tests/experimental/optim/test_dynamic_loss_scaler.py
tmarkstrum: will add f16 model in the next pr.

.pre-commit-config.yaml
tmarkstrum: I changed the version back to python3.7. Do we have to pin the version to python3.7 as the default python version has be specified as python3 on the top of the pre-commit-config.yaml file?

anj-s: cc @blefaudeux who uses pre-commit. I don't use pre-commit so not sure about this.

tmarkstrum:
> Might be easier for the user to call this GradScaler and implement the same interface so that it is a drop-in replacement.

The scaling logic is not exactly same with GradScaler. The logic is same with the dynamic_loss_scaler in fairseq. I am concerned that users will assume this scaler having the exact same behavior as GradScaler if we are using the same name.

fairscale/experimental/optim/dynamic_loss_scaler.py
```diff
@@ -0,0 +1,209 @@
...
+                        "Try lowering the learning rate, using gradient clipping or "
+                        "increasing the batch size."
+                    ).format(self.min_loss_scale)
+                )
+
+            self._iter += 1
+            raise OverflowError("setting loss scale to: " + str(self.loss_scale))
+
+    def update(self) -> None:
+        """
```
anj-s, line 127: 127-129 can be on the same line.

fairscale/experimental/optim/dynamic_loss_scaler.py
anj-s: In that case I would remove this line or add more details.

fairscale/experimental/optim/dynamic_loss_scaler.py
anj-s: Can we add a TODO to follow up here?

anj-s: Just saw that you added a TODO.

anj-s (approved):

.pre-commit-config.yaml
blefaudeux: we should use the same as what the CI is using, that's the only rule I know. CI seems to be using black directly, not specifying the language version, so I don't think that it changes much in practice (for this particular repo)

PR #657, opened 2021-05-05
[fix] better assert and better test for frozen weights

### LINKED ISSUE #643: [FSDP] Frozen weights support
## 🐛 Bug: Frozen weight support in FSDP

This bug is a follow up to issue https://github.com/facebookresearch/fairscale/issues/610. I encountered a new test case in which a scenario of linear evaluation fails.

### Context (same as previous issue)

In the context of VISSL, we are training a SWAV model with FSDP and then want to evaluate this model using linear evaluation, again using FSDP to wrap our model.

In the context of linear evaluation, only the linear layer on top of the "trunk" will be learned: the representations of the trunk, fed to the linear layer, will stay constant, and so we set `requires_grad=False` on the parameters of the trunk.

### New symptom

The initial test case raised in https://github.com/facebookresearch/fairscale/issues/610 is green, but I discovered that nesting FSDP blocks in the trunk (the part that is frozen) is leading to a new assertion being raised:

`ERROR: expected to be in states [<TrainingState.BACKWARD_PRE: 3>] but current state is TrainingState.IDLE`

## To Reproduce

This script should help you reproduce the issue: simply run the following script with the options `-f -l` where:
- `-f` means "enable FSDP"
- `-l` means "freeze the trunk"

**Important notes**:
- The issue only appears if both `-f` and `-l` are used together
- The issue disappear if I comment the line `self.trunk = FSDP(self.trunk)` and so it seems that the issue is linked to the nesting of FSDP blocks when layers are frozen

```
import argparse

import torch
import torch.nn as nn
import torch.optim as optim
from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP
from torch.nn.parallel import DistributedDataParallel


class Model(nn.Module):
    def __init__(self, with_fsdp: bool):
        super().__init__()
        self.trunk = nn.Sequential(
            self._create_block(3, 64, with_fsdp),
            self._create_block(64, 64, with_fsdp),
        )
        self.head = nn.Sequential(
            nn.AdaptiveAvgPool2d(output_size=(1, 1)),
            nn.Flatten(),
            nn.Linear(64, 10),
        )
        if with_fsdp:
            self.trunk = FSDP(self.trunk)
            self.head = FSDP(self.head)

    def forward(self, x):
        return self.head(self.trunk(x))

    def _create_block(self, in_channels: int, out_channels: int, with_fsdp: bool):
        block = nn.Sequential(
            nn.Conv2d(in_channels, out_channels, kernel_size=3),
            nn.ReLU(inplace=True),
        )
        if with_fsdp:
            block = FSDP(block)
        return block


def create_model(with_fsdp: bool):
    return Model(with_fsdp)


def _distributed_worker(gpu_id: int, with_fsdp: bool, with_linear: bool):
    torch.cuda.set_device(gpu_id)
    dist.init_process_group(
        backend="nccl", init_method="tcp://127.0.0.1:9099", world_size=2, rank=gpu_id
    )

    torch.manual_seed(0)
    torch.backends.cudnn.deterministic = True
    batch = torch.randn(size=(2, 3, 224, 224)).cuda()

    model = create_model(with_fsdp)
    model = model.cuda()

    if with_linear:
        for name, param in model.named_parameters():
            if "trunk" in name:
                param.requires_grad = False

    if with_fsdp:
        model = FSDP(model)
    else:
        model = DistributedDataParallel(model, device_ids=[gpu_id])

    if gpu_id == 0:
        print(model)

    target = torch.LongTensor([0, 1]).cuda()
    criterion = nn.CrossEntropyLoss()
    optimizer = optim.SGD(model.parameters(), lr=0.1, momentum=0.9)

    for iteration in range(3):
        out = model(batch)
        fake_loss = criterion(out, target)
        if gpu_id == 0:
            print("Loss", iteration, ":", fake_loss.item())
        optimizer.zero_grad()
        fake_loss.backward()
        optimizer.step()


if __name__ == "__main__":
    parser.add_argument("-f", "--fsdp", action="store_const", const=True, default=False)
    parser.add_argument(
        "-l", "--linear", action="store_const", const=True, default=False
    )
    args = parser.parse_args()
    mp.spawn(_distributed_worker, (args.fsdp, args.linear), nprocs=2)
```

## Environment

I use pytorch 1.6.0+cu102, with the following modules enabled: `module load cuda/11.0 cudnn/v8.0.3.33-cuda.11.0 NCCL/2.8.3-1-cuda.11.0`.

My version of fairscale is 0.3.6.

## Additional context

CC: @prigoyal @min-xu-ai

min-xu-ai:
- the precise condition should have been check m.parameters(), not
  m.params.
- fixes #643

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1301,15 +1301,19 @@ class FullyShardedDataParallel(nn.Module):
             if isinstance(m, FullyShardedDataParallel):
                 _remove_shard_bwd_hook(m)
                 m._pre_backward_hook_has_run = False
-                if m._has_params:
-                    if any(p.requires_grad for p in m.params):
+                # Note: m.parameters() should not be an empty list. FSDP
+                # wrapping modules without weights is not tested at the moment.
+                if any(p.requires_grad for p in m.parameters()):
+                    if m._has_params:
                         m.assert_state(TrainingState.BACKWARD_POST)
                     else:
-                        # Unlikely case, should only happens if `m` has params but none of the
-                        # params has `requires_grad==True`.
-                        m.assert_state(TrainingState.IDLE)
+                        m.assert_state(TrainingState.BACKWARD_PRE)
                 else:
-                    m.assert_state(TrainingState.BACKWARD_PRE)
+                    # Unlikely case. When m and its children has no params
+                    # with `requires_grad==True`, then m's pre-backward and
+                    # post-backward hooks aren't called by autograd. Therefore,
+                    # it is in IDLE state.
+                    m.assert_state(TrainingState.IDLE)
                 m.training_state = TrainingState.IDLE
 
     @torch.no_grad()
--- a/tests/nn/data_parallel/test_fsdp_freezing_weights.py
+++ b/tests/nn/data_parallel/test_fsdp_freezing_weights.py
@@ -10,6 +10,7 @@
 """ Test FSDP with some params frozen. """
 
 
+from enum import Enum
 import tempfile
 
 import pytest
@@ -68,6 +69,11 @@ def _create_model(with_fsdp, with_nested_trunk):
     return model
 
 
+class FreezingMethod(str, Enum):
+    GradToNone = "grad_to_none"
+    RequiresGrad = "requires_grad"
+
+
 def _distributed_worker(
     gpu_id,
     world_size,
@@ -93,8 +99,7 @@ def _distributed_worker(
     model = model.cuda()
 
     # freezing the trunk using requires_grad.
-    assert freezing_method in ["requires_grad", "grad_to_none"]
-    if freezing_method == "requires_grad":
+    if freezing_method == FreezingMethod.RequiresGrad:
         for param in model.trunk.parameters():
             param.requires_grad = False
 
@@ -116,7 +121,7 @@ def _distributed_worker(
         print("Loss", iteration, ":", fake_loss.item())
         optimizer.zero_grad()
         fake_loss.backward()
-        if freezing_method == "grad_to_none":
+        if freezing_method == FreezingMethod.GradToNone:
             for param in model.trunk.parameters():
                 param.grad = None
         optimizer.step()
@@ -155,7 +160,7 @@ def test_freezing_weights(temp_files, nested_trunk):
     world_size = 2
     # DDP
     with_fsdp = False
-    freezing_method = "requires_grad"
+    freezing_method = FreezingMethod.RequiresGrad
     mp.spawn(
         _distributed_worker,
         (world_size, with_fsdp, with_nested_trunk, freezing_method) + temp_files[0:3] + (None,),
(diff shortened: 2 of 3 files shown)


min-xu-ai:
@QuentinDuval @prigoyal FYI.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1301,15 +1301,19 @@ def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
             if isinstance(m, FullyShardedDataParallel):
                 _remove_shard_bwd_hook(m)
                 m._pre_backward_hook_has_run = False
-                if m._has_params:
-                    if any(p.requires_grad for p in m.params):
+                # Note: m.parameters() should not be an empty list. FSDP
+                # wrapping modules without weights is not tested at the moment.
```
anj-s, line 1305: Should we add an error check here for this case then?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1301,15 +1301,19 @@ def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
...
+                if any(p.requires_grad for p in m.parameters()):
+                    if m._has_params:
```
anj-s, line 1307: what does `_has_params`  bool signify?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: Thanks for the review! I prefer to do that separately since it is a different use case and the check might be better in the init. I will file an issue.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: It indicates if this instance has any “owned” params or not. It is possible that all params within this module is owned by children FSDP instances, leaving the parent empty without any owned params

tests/nn/data_parallel/test_fsdp_freezing_weights.py
```diff
@@ -118,21 +148,30 @@ def temp_files():
...
-def test_freezing_weights(temp_files):
+@pytest.mark.parametrize("nested_trunk", ["nested_trunk", "simple_trunk"])
+def test_freezing_weights(temp_files, nested_trunk):
+    with_nested_trunk = nested_trunk == "nested_trunk"
+
     world_size = 2
     # DDP
-    fsdp = False
+    with_fsdp = False
     freezing_method = "requires_grad"
```
blefaudeux, line 158: nit and pet peeve of mine, but 
```
class FreezingMethod(str, Enum):
   ToNone = "to_none"
   RequiresGrad = "requires_grad"
```
removes the need for your asserts, adds type checking in different places, and makes it easy for a newcomer to know what are the supported methods without looking for asserts

blefaudeux: (it systematically prevents silent fall through for mismatches, even if you don't think of adding an assert everywhere, can be useful..)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: issue filed.

tests/nn/data_parallel/test_fsdp_freezing_weights.py
min-xu-ai: done. it is so much better! thanks for the suggestion!

QuentinDuval (approved):

PR #662, opened 2021-05-06
[feat] experimental.nn.SyncBatchNorm: initial commit

msbaines:
Fast/simple re-implementation of SyncBatchNorm.

When profiling SSL Vision, I was seeing a majority of cycles spent in
SyncBatchNorm. With this change, I see a 10% to 20% speedup on the
model I was profiling.

When running benchmarks/experimental/sync_batchnorm.py on 8 x V100,
I get a 6x speedup:

<class 'torch.nn.modules.batchnorm.BatchNorm2d'>
Elapsed time is  0.08709120750427246
Elapsed time is  0.12632274627685547
Elapsed time is  0.14095258712768555
Elapsed time is  0.16529417037963867
Elapsed time is  0.1419970989227295
Elapsed time is  0.15166854858398438
Elapsed time is  0.12000870704650879
Elapsed time is  0.17534875869750977
<class 'torch.nn.modules.batchnorm.SyncBatchNorm'>
Elapsed time is  2.5087168216705322
Elapsed time is  2.497001886367798
Elapsed time is  2.5204885005950928
Elapsed time is  2.526789903640747
Elapsed time is  2.5080230236053467
Elapsed time is  2.524489641189575
Elapsed time is  2.513214588165283
Elapsed time is  2.5359973907470703
<class 'fairscale.experimental.nn.sync_batchnorm.SyncBatchNorm'>
Elapsed time is  0.4126114845275879
Elapsed time is  0.39051294326782227
Elapsed time is  0.40685415267944336
Elapsed time is  0.4159870147705078
Elapsed time is  0.42383885383605957
Elapsed time is  0.4080159664154053
Elapsed time is  0.41202712059020996
Elapsed time is  0.42400121688842773

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/benchmarks/experimental/sync_batchnorm.py
+++ b/benchmarks/experimental/sync_batchnorm.py
@@ -1,0 +1,40 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+import tempfile
+import time
+
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+import fairscale.experimental.nn
+
+
+def benchmark_bn(rank, world_size, init_file, bn_cls):
+    dist.init_process_group(dist.Backend.NCCL, init_method="file://" + init_file, rank=rank, world_size=world_size)
+    x = torch.randn(50, 2048, 7, 7).to(rank)
+    bn = bn_cls(2048).to(rank)
+    bn = DDP(bn, device_ids=[rank])
+    # Warmup
+    for i in range(50):
+        with torch.no_grad():
+            x = bn(x)
+    torch.cuda.synchronize(rank)
+    t0 = time.time()
+    for i in range(100):
+        with torch.no_grad():
+            x = bn(x)
+    torch.cuda.synchronize(rank)
+    t1 = time.time()
+    print("Elapsed time is ", t1 - t0)
+
+
+if __name__ == "__main__":
+    world_size = torch.cuda.device_count()
+    for cls in [torch.nn.BatchNorm2d, torch.nn.SyncBatchNorm, fairscale.experimental.nn.SyncBatchNorm]:
+        print(cls)
+        mp.spawn(benchmark_bn, args=(world_size, tempfile.mkstemp()[1], cls), nprocs=world_size)
--- a/fairscale/experimental/nn/sync_batchnorm.py
+++ b/fairscale/experimental/nn/sync_batchnorm.py
@@ -44,11 +44,13 @@ def _forward(
     running_var: torch.Tensor,
     total_count: torch.Tensor,
 ) -> torch.Tensor:
-    scale = torch.rsqrt(var + eps)
-    bias = -mean * scale
+    invstd = torch.rsqrt(var + eps)
     if affine:
-        scale = scale * weight_param.reshape(mean.shape)
-        bias = bias + bias_param.reshape(mean.shape)
+        scale = weight_param.reshape(mean.shape) * invstd
+        bias = bias_param.reshape(mean.shape) - mean * scale
+    else:
+        scale = invstd
+        bias = -mean * scale
 
     if track_running_stats:
         with torch.no_grad():
@@ -67,7 +69,7 @@ class SyncBatchNorm(torch.nn.BatchNorm2d):
     """
     Fast re-implementation of ``torch.nn.SyncBatchNorm`` that can achieve a speedup
     of 5x or more over the default implementation depending on size of the input
-    and number of distributed world.
+    and number of distributed workers.
     """
 
     def __init__(
--- a/fairscale/experimental/nn/sync_batchnorm.py
+++ b/fairscale/experimental/nn/sync_batchnorm.py
@@ -1,0 +1,170 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict, Optional, Tuple
+
+import torch
+import torch.distributed as dist
+from torch.distributed import ProcessGroup
+
+if torch.__version__.split(".")[:2] >= ["1", "8"]:
+    from torch.distributed import all_reduce
+else:
+    # Copied from https://github.com/pytorch/pytorch/blob/v1.8.1/torch/distributed/nn/functional.py
+    class _AllReduce(torch.autograd.Function):
+        @staticmethod
+        def forward(ctx, op, group, tensor):  # type: ignore
+            ctx.group = group
+            ctx.op = op
+            tensor = tensor.clone()
+            dist.all_reduce(tensor, op=op, group=group)
+            return tensor
+
+        @staticmethod
+        def backward(ctx, grad_output):  # type: ignore
+            return (None, None) + (_AllReduce.apply(ctx.op, ctx.group, grad_output),)
+
+    def all_reduce(tensor, op=dist.ReduceOp.SUM, group=dist.group.WORLD):  # type: ignore
+        return _AllReduce.apply(op, group, tensor)
+
+
+def _forward(
+    input: torch.Tensor,
+    affine: bool,
+    track_running_stats: bool,
+    mean: torch.Tensor,
+    var: torch.Tensor,
+    momentum: float,
+    eps: float,
+    weight_param: torch.Tensor,
+    bias_param: torch.Tensor,
+    running_mean: torch.Tensor,
+    running_var: torch.Tensor,
+    total_count: torch.Tensor,
+) -> torch.Tensor:
+    scale = torch.rsqrt(var + eps)
+    bias = -mean * scale
+    if affine:
+        scale = scale * weight_param.reshape(mean.shape)
+        bias = bias + bias_param.reshape(mean.shape)
+
+    if track_running_stats:
+        with torch.no_grad():
+            unbiased_var = var * (total_count / (total_count - 1))
+            running_mean += momentum * (mean.reshape(-1) - running_mean)
+            running_var += momentum * (unbiased_var.reshape(-1) - running_var)
+
+    return input * scale + bias
+
+
+if torch.__version__.split(".")[:2] >= ["1", "7"]:
+    _forward = torch.jit.script(_forward)  # type: ignore
+
+
+class SyncBatchNorm(torch.nn.BatchNorm2d):
+    """
+    Fast re-implementation of ``torch.nn.SyncBatchNorm`` that can achieve a speedup
+    of 5x or more over the default implementation depending on size of the input
+    and number of distributed world.
+    """
+
+    def __init__(
+        self, *args: Tuple[Any, ...], process_group: Optional[ProcessGroup] = None, **kwargs: Dict[str, Any]
+    ) -> None:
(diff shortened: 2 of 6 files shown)


min-xu-ai:
This is very cool! Is the speed up mainly from jit.script or mainly from network traffic optimization?

@prigoyal and I are running into sync_bn issues, but not yet around performance. For us, it is tricky get to it work under activation checkpoint + AMP. There seems to be a corner case bug somewhere. (activation checkpoint matter because of the `track_running_stats`, which we set to false in the no_grad forward pass and reenable back to true in the with_grad forward pass in the checkpoint case.

msbaines:
> This is very cool! Is the speed up mainly from jit.script or mainly from network traffic optimization?
> 
> @prigoyal and I are running into sync_bn issues, but not yet around performance. For us, it is tricky get to it work under activation checkpoint + AMP. There seems to be a corner case bug somewhere. (activation checkpoint matter because of the `track_running_stats`, which we set to false in the no_grad forward pass and reenable back to true in the with_grad forward pass in the checkpoint case.

The PyTorch implementation has a DeviceToHost transfer which this implementation does not. It also uses all_gather while here there are two all_reduces (first could potentially be removed) so the communication over-head is much much less here. The PyTorch implementation uses CUDA kernels so its actually much more efficient in that regard. The jit.script enables some fusing but not so much that its material.

benchmarks/experimental/sync_batchnorm.py
```diff
@@ -0,0 +1,40 @@
...
+        with torch.no_grad():
+            x = bn(x)
+    torch.cuda.synchronize(rank)
+    t0 = time.time()
+    for i in range(100):
+        with torch.no_grad():
+            x = bn(x)
+    torch.cuda.synchronize(rank)
+    t1 = time.time()
+    print("Elapsed time is ", t1 - t0)
```
anj-s, line 33: Is this a temporary file or do we intend to add this to our benchmark runs on circleCI?

fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -0,0 +1,170 @@
...
+
+if torch.__version__.split(".")[:2] >= ["1", "7"]:
+    _forward = torch.jit.script(_forward)  # type: ignore
+
+
+class SyncBatchNorm(torch.nn.BatchNorm2d):
+    """
+    Fast re-implementation of ``torch.nn.SyncBatchNorm`` that can achieve a speedup
+    of 5x or more over the default implementation depending on size of the input
+    and number of distributed world.
```
anj-s, line 70: s/number of distributed world/devices perhaps?

fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -0,0 +1,170 @@
...
+        with torch.no_grad():
+            unbiased_var = var * (total_count / (total_count - 1))
+            running_mean += momentum * (mean.reshape(-1) - running_mean)
+            running_var += momentum * (unbiased_var.reshape(-1) - running_var)
+
+    return input * scale + bias
+
+
+if torch.__version__.split(".")[:2] >= ["1", "7"]:
+    _forward = torch.jit.script(_forward)  # type: ignore
```
anj-s, line 63: am curious, what is the gain we see by using JIT?

fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class SyncBatchNorm(torch.nn.BatchNorm2d):
```
anj-s, line 66: any reason we decided to subclass BatchNorm2d? Why not subclass torch.nn.SyncBatchNorm so that we could use the `convert_sync_batchnorm `. Are there things that we have modified in this function?

anj-s:
It also looks like some tests might be failing on older PyTorch versions.

prigoyal:
@msbaines , this is great. I would love to test this out for convergence, speed etc whenever you think makes sense.

fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -0,0 +1,170 @@
...
+        self._process_group = process_group if process_group is not None else dist.group.WORLD
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
+        if not dist.is_initialized() or not self.training:
+            return super().forward(input)
+
+        dim = [d for d in range(input.ndim) if d != 1]
+        count = torch.full((1,), input.numel() // input.size(1), device=input.device, dtype=input.dtype)
+        total_count = count.clone()
+        handle = all_reduce(total_count, group=self._process_group, async_op=True)
```
anj-s, line 86: I can see that this might be useful when we have a partial batch. Curious, are there other reasons why we need this allreduce?

fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -0,0 +1,170 @@
...
+        mean = torch.mean(input, dim=dim, keepdim=True)
+        meansqr = torch.mean(input * input, dim=dim, keepdim=True)
+        vec = torch.cat([mean, meansqr])
+        handle.wait()
+        vec = vec * (count / total_count)
```
anj-s, line 91: why don't we just divide by total_count(which should. be the global batch size * (product of all dimensions)? Maybe I don't get how we calculate mean and meansqr but I am not sure I understand why we need to multiple by count here.

benchmarks/experimental/sync_batchnorm.py
msbaines: This file is useful for profiling and was helpful to me when iterating on the code. I'm not yet sure we should run this on circleci. But I plan to use it to use it in the future whenever I make changes.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: I see a reduction in kernel calls in the trace.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: Unfortunately SyncBatchNorm's `convert_sync_batchnorm` explicitly calls `SyncBatchNorm` instead of `cls` so sub-classing it would still require implementing `convert_sync_batchnorm`. I will modify to sub-class `_BatchNorm` although I'm hesitant to sub-class a private class. I didn't want to sub-class SyncBatchNorm because it felt odd to implement `SyncBatchNorm` which sub-classes `SyncBatchNorm`. I was also worried about future changes to `SyncBatchNorm` breaking my sub-class. Felt safer to sub-class on of the other classes. Sub-classes `_BatchNorm` might be the least-worst option.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: Yeah, the primary motivation is to handle partial batches. Another case would be when batch sizes are not the same. This can sometimes happen in NLP due to variable sequence length although we don't use BatchNorm in NLP.

I am contemplating adding a batch-size argument that would avoid this all_reduce and assert if their is a non-matching batch. But this all_reduce does not impact performance that much since it is just a single float and is async. Although, it could matter in a network with high-latency.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: (count / total_count) is this workers contribution to the mean. This is to handle one worker having a larger batch than another. You can think of it like a probability.

So if total_count is 100 and this worker has a count of 40, it will contribute 0.4 towards the aggregate mean.

msbaines:
> @msbaines , this is great. I would love to test this out for convergence, speed etc whenever you think makes sense.

Cool. I would recommend waiting until this change lands just in case we find some issues.

I've done some preliminary testing with @endernewton's code to verify that loss does not change when SyncBatchNorm is replaced with this implementation. Ideally, the results should be close to identical with epsilon differences attributed to the order of operations.

benchmarks/experimental/sync_batchnorm.py
blefaudeux: just FYI (recent discovery of mine), torch has a `torch.utils.benchmark` util which is very nice for these things, removes some boilerplate code and prettifies the outputs. Example [here](https://github.com/fairinternal/xformers/blob/main/benchmarks/benchmark_core.py). I agree that having these small tools is useful, I've not used that enough myself, should have

VitaliyLi:
This is sooo great, @msbaines, 6x speedup!!!!!!

tests/experimental/nn/test_sync_batchnorm.py
```diff
@@ -0,0 +1,144 @@
...
+import torch.multiprocessing as mp
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+from fairscale.experimental.nn import SyncBatchNorm
+
+pytestmark = pytest.mark.skipif(not torch.cuda.is_available(), reason="cuda required")
+
+
+def pg_worker(rank, world_size, init_file, func, *args):
+    dist.init_process_group(dist.Backend.NCCL, init_method="file://" + init_file, rank=rank, world_size=world_size)
```
anj-s, line 21: Since this is a fundamental module should we test the GLOO backend as well? if things don't work as expected we can then limit to NCCL support but would help surface any GLOO specific issues.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: I just measured this. Removing the JIT usage increases execution in the benchmark by about 40%.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: Good idea! I considered this. You need to all_reduce proportionally to the local batch_size. I could handle that by all_reduce-ing the sum instead of the mean but then the problem is that you then overflow. I could avoid the overflow by first dividing by a "scale factor" that is order of total_count but was worried about the extra complexity that would add.

msbaines:
While debugging the unit tests, I discovered a bug in PyTorch's computation of the  BatchNorm weight gradient:

https://github.com/pytorch/pytorch/issues/57796

tests/experimental/nn/test_sync_batchnorm.py
msbaines: Cool if I handle that in a follow-up change since its marked as experimental? Wanted to unblock testing of the code by others.

anj-s: Absolutely! thanks for following up.

fairscale/experimental/nn/sync_batchnorm.py
anj-s: I think doing the allreduce might be better than requiring an explicit batch size argument.

anj-s:
> While debugging the unit tests, I discovered a bug in PyTorch's computation of the BatchNorm weight gradient:
> 
> [pytorch/pytorch#57796](https://github.com/pytorch/pytorch/issues/57796)

Thank you for the extensive testing! This work has uncovered lot of interesting things.

anj-s (approved):

endernewton:
Hey, @msbaines nice work on this speed-up! I started trying this on my side on 8x8 GPUs but I could not benchmark the speed on my side as it says "OOM" for the default use case I have (which do not have OOM issue). So I am wondering what is the memory usage of this new op compared to PyTorch's default? If it is less efficient, can it be further optimized in terms of memory usage? Thanks! Attached is the screenshot (I haven't checked the details of the code and just try to use it as an end-user).

image

PR #671, opened 2021-05-09
[fix] FSDP forward pass overlap between compute and all-gather

### LINKED ISSUE #623: FSDP does not overlap all gather with forward pass
## 🐛 Bug

FSDP doesn't seem to overlap all-gather calls with forward pass when wrapping individual layers with `wrap()` or `auto_wrap()`. However, the reduce-scatter calls do overlap with computation in the backward pass. 

## To Reproduce

An example of wrapping individual layers with `wrap()` is, 

```
    layers=[]
    fsdp_params = dict(flatten_parameters=True, reshard_after_forward=False)
    with enable_wrap(wrapper_cls=FSDP, process_group=group, **fsdp_params):
        for i in range(num_layers):
            layer=wrap(BERTLayer(hidden_size, hidden_size, num_heads, i))
            assert isinstance(layer,FSDP)
            layers.append(layer)
    m = nn.Sequential(*layers).to(rank)
    opt = base_optimizer(params=m.parameters(), **base_optimizer_arguments)
```

## Problem

The Nsys profile of the above code with 8 layers (num_layers = 8) is: 

image

The highlighted green slice is the first layer of the model in the forward pass. The AllGather calls do not overlap with `Kernels`, but in the backward pass, ReduceScatter calls do overlap with `Kernels` as desired. 

This symptom could be caused by the line: 

https://github.com/facebookresearch/fairscale/blob/b0048b2873de5a1752226977a24a140da22030c7/fairscale/nn/data_parallel/fully_sharded_data_parallel.py#L909

This line adds dependency for AllGather calls to wait on the previous optimizer step (which makes sense), but it also stops the next layer of the model from gathering model parameters from different processors, while the current layer is running computations. This `wait_stream` may be a bit too heavy-handed. It could probably be changed to wait on an event of the previous opt step, rather than on the `current_stream()`, serializing operations between layers. 

## Environment

 - PyTorch Version (e.g., 1.0): 1.9.0a0+df837d0
 - OS (e.g., Linux): Ubuntu 20.04.1 LTS (x86_64)
 - How you installed PyTorch (`conda`, `pip`, source): `conda`
 - Build command you used (if compiling from source): 
 - Python version: 3.8 (64-bit runtime)
 - CUDA/cuDNN version: 11.2.142/8.1.1
 - GPU models and configuration: A100-SXM4-40GB
 - Any other relevant information:

min-xu-ai:
- much thanks for @cyanguwa for reporting this and @QuentinDuval for debugging it
- a new unit test is added to check for this and ensure we detect
  issue with overlapping and cpu/gpu blocking wait calls

# Before submitting

## What does this PR do?
Fixes #623 .

## PR review

## Did you have fun?


tests/nn/data_parallel/test_fsdp_overlap.py
```diff
@@ -0,0 +1,207 @@
...
+            # 10X longer is a safe margin, since the time is around 100X longer
+            # when there is work on GPU vs. no work.
+            assert s * 10 < l, f"{s} and {l}"
+
+    # Check the GPU overlapping when there is all-gather.
+    if world_size > 1:
+        compute_only = e3[2]
+        all_gather_only = e2[3]
+        both = e4[3]
+        assert compute_only + all_gather_only > 1.1 * both, f"{compute_only} {all_gather_only} > 1.1 {both}"
```
min-xu-ai, line 190: @myleott @QuentinDuval, please let me know if you have better ideas for unit testing this kind of issues.

min-xu-ai:
It is super interesting that flatten and non_flatten seems to behave unexpectedly:

flatten: (second line, last number is all_gather time)
image

non_flatten:
image

These are from my desktop but CI seems to have the same issue too.

myleott:
Nice!

We can make this test more robust by using a patched all_gather with a delay, instead of tuning the number of MB.

Also there was a bug in the test, where we weren't doing an all-gather in some settings because the parameters aren't actually part of the graph (so `.item()` doesn't properly block). Both have been fixed in https://github.com/facebookresearch/fairscale/pull/672

min-xu-ai:
Thank you so much @myleott! Saved me a lot of time. Really appreciated.

Do you have any speculations on why was I observing the all-gather time difference between flatten and non-flatten above?

tests/nn/data_parallel/test_fsdp_overlap.py
```diff
@@ -0,0 +1,249 @@
...
+                times.append(mod.get_time())
+
+            # get gpu compute + all_gather time
+            overall_gpu_time = e1.elapsed_time(e2)
+
+            # print(f"rank {rank}", cpu_iter_time, cpu_wait_for_gpu_time, sum(times), "vs", e1.elapsed_time(e2))
+            cpu_iter.add(cpu_iter_time)
+            cpu_wait.add(cpu_wait_for_gpu_time)
+            gpu_compute.add(sum(times))
+            gpu_total.add(overall_gpu_time)
```
myleott, line 177: Copying the discussion from here: https://github.com/facebookresearch/fairscale/pull/672#discussion_r629034005

> I seem to see all of them 1-2 ms: 
> image

Interesting! I see several hundred ms on my local V100 machine. It's very strange that CI is reporting such different results... I wonder if something weird is happening in CI?

min-xu-ai: My local test shows that it is passing with very good values. clearly the overlapping is there. :-)

myleott (approved):
LGTM!

PR #672, opened 2021-05-10
Update + fix overlapping test to use patched all_gather w/ delay

myleott:
* make the test more robust by using a patched all_gather with an explicit delay (instead of tuning MB)
* fix bug in test where `.item()` was not properly blocking, due to parameters not being part of the graph
* fix the actual overlapping bug by introducing a new stream for freeing full parameters

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -960,8 +960,6 @@ class FullyShardedDataParallel(nn.Module):
         self._streams["fp32_to_fp16"] = torch.cuda.Stream()
         # Stream for all-gathering parameters.
         self._streams["all_gather"] = torch.cuda.Stream()
-        # Stream for freeing full parameters.
-        self._streams["free_full_params"] = torch.cuda.Stream()
         # Stream for overlapping grad reduction with the backward pass.
         self._streams["post_backward"] = torch.cuda.Stream()
         # Helper for bucketing reduce-scatter ops. This is also shared with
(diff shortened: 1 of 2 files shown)


tests/nn/data_parallel/test_fsdp_overlap.py
```diff
@@ -37,7 +47,10 @@ def forward(self, x):
 
         # Record the fake forward compute time.
         self.e1.record()
-        torch.cuda._sleep(self.sleep_cycles)
+        if self.sleep_cycles > 0:
+            torch.cuda._sleep(self.sleep_cycles)
+        if self.l is not None:
+            x = x + self.l  # force params to be part of the graph
```
min-xu-ai, line 53: Nice!

tests/nn/data_parallel/test_fsdp_overlap.py
```diff
@@ -108,11 +126,23 @@ def run(compute_cycles, all_gather_mb):
 
             cpu_start = time.process_time()
 
+            all_gather_called = [False]
+
+            def _delayed_all_gather(*args, **kwargs):
+                all_gather_called[0] = True
```
min-xu-ai, line 132: nit: could just use nonlocal without using a list?

tests/nn/data_parallel/test_fsdp_overlap.py
```diff
@@ -143,53 +173,48 @@ def run(compute_cycles, all_gather_mb):
             cpu_iter.add(cpu_iter_time)
             cpu_wait.add(cpu_wait_for_gpu_time)
             gpu_compute.add(sum(times))
-            gpu_total.add(e1.elapsed_time(e2))
+            gpu_total.add(overall_gpu_time)
```
min-xu-ai, line 176: after this change, it seems that CI shows the resulting gpu timing became around 1~2 ms, which is strange since we asked for 100ms and there are 4 layers, should be around 400ms (if perfectly overlapped). Am I misreading the test results?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1451,9 +1453,8 @@ def _free_full_params(self, params: Optional[List[Parameter]] = None) -> None:
         if params is None:
             params = self.params
         self.has_full_params = False
-        # XXX: testing
-        self._streams["all_gather"].wait_stream(torch.cuda.current_stream())
-        with torch.cuda.stream(self._streams["all_gather"]):
+        self._streams["free_full_params"].wait_stream(torch.cuda.current_stream())
```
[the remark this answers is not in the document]
myleott: yeah, it could probably go in the current (compute) stream, lemme do that instead since it's simpler

myleott:
Going to merge into #671 and let CI run there 😄

min-xu-ai:
> Going to merge into #671 and let CI run there 😄

Sounds great. I will debug whatever issues over there as well.

PR #682, opened 2021-05-13
Save FSDP metadata for offline unflattening

sshleifer:
Context: @QuentinDuval  and I agreed that I could support the streaming checkpoint effort by making a method that returns all the extra info needed to combine `local_state_dict`s  into larger unflat checkpoints offline.

VERY rough idea of where to go after saving `shard_reconstruction_info()`.

```python

info = [torch.load(f'checkpoint-layer_1_reconstruction_info_{i}.pt') for i in world_size]
state = [torch.load(f'checkpoint-layer_1_rank_{i}.pt') for i in world_size]
```
#info[i] should correspond to state[i]

1) unpad using numel_padded_per_param
2) torch.cat(unpadded_tensors)
3) unflatten as below

```python
(t.view(s) for (t, s) in zip(flat_param.split(param_numels), unflat_shapes))  # copy the source code from FlattenParamsWrapper.get_param_views
```

cc @QuentinDuval @prigoyal @min-xu-ai

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1470,6 +1470,17 @@ class FullyShardedDataParallel(nn.Module):
             # the Storage to 0 to save memory.
             free_storage_(p._full_param_padded)
 
+    def shard_reconstruction_info(self, consider_nested: bool = True) -> Dict[str, list]:
+        """Get the information needed to reconstruct the model from shards offline."""
+        _fsdp_instances = self._fsdp_instances if consider_nested else [self._fsdp_instances[0]]
+        assert hasattr(_fsdp_instances[0], "_param_numels"), "This method only supports flatten_parameters=True"
+        return dict(
+            num_padded=[m.numel_padded_per_param for m in _fsdp_instances],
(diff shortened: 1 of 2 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1470,6 +1470,17 @@ def _free_full_params(self, params: Optional[List[Parameter]] = None) -> None:
             # the Storage to 0 to save memory.
             free_storage_(p._full_param_padded)
 
+    def shard_reconstruction_info(self, consider_nested: bool = True) -> Dict[str, list]:
+        """Get the information needed to reconstruct the model from shards offline."""
```
min-xu-ai, line 1474: @QuentinDuval, since you will be using this API, may I suggest that you add to this PR:

- example usage of this function in the doctoring
- simple test that actually reconstruct the state using this metadata?

For a shorter name, perhaps rename this function to `shard_metadata_dict` to be symmetric to `state_dict`?

QuentinDuval: Sure, I can had some documentation (I will get this PR and try it this morning). In terms of naming, I would suggest the following: `local_metadata_dict` to mirror `local_state_dict`.

min-xu-ai:
I hope people don't mind closing this. :-)

PR #685, opened 2021-05-15
[fix] auto_wrap: support wrapping based on wrapper_config

### LINKED ISSUE #585: FSDP: easier wrapping
Two use cases:

- easier wrapping based on layer type (like BN) is already handled by auto_wrap_bn
- we can extend this to wrap based on a flag in the module
- also, we can make it easier to wrap a layer with a given precision (perhaps also from a flag in the module)

### LINKED ISSUE #649: [auto_wrap] Supporting multiple auto_wrap
### Context

In the context of VISSL, we are interested in calling `auto_wrap_bn` to deal with the Synchronized Batch Normalization layers and then call `auto_wrap` to automatically wrap some other layers (such as big layers by using a policy like `default_auto_wrap_policy`).

However, the following assertion in `auto_wrap.py` prevents us from doing so:

```
# Make sure no child is not already wrapped.
for _, child in module.named_modules():
    assert not isinstance(child, cast(type, ConfigAutoWrap.wrapper_cls))
```

This assertion seems necessary considering what follows in the code:

```
# We count all params, assuming none of them is already wrapped.
num_params = sum([p.numel() for p in module.parameters()])
```

But in the case of `auto_wrap_bn`, it is not necessary (since the BN will not be shared).

### Workaround

My original idea was to work-around that is to write my own policy that combines `auto_wrap_bn` and the subsequent `auto_wrap` but this is cumbersome:
- I end up not directly using a very useful feature of fairscale
- Since I want different parameters for FSDP for BN and other layers, I will have to hack a bit to make it work

I found another way to circumvent the assertion in my case, by using this hack:

```
class _FSDP_WRAPPER:
    def __new__(cls, module, **kwargs):
        return FSDP(module, **kwargs)

with enable_wrap(auto_wrap_policy=_policy, wrapper_cls=_FSDP_WRAPPER, **fsdp_config):
        return auto_wrap(module)
```

This allows me to avoid the check completely, by using `__new__` to return a `FSDP` wrapper hidden behind another class. 

Nifty, it works, but is there any better solution available to me? (avoiding such hacks)

## Interested parties

CC: @min-xu-ai @prigoyal

min-xu-ai:
- user can use this to avoid assert if auto_wrap is used multiple times on a module
- user can traverse the modules multiple times and assign a wrapper_config
  to the module and then use auto_wrap once to wrap them

fix #649
fix #585

## What does this PR do?
Fixes # (issue).

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [x] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## NEXT - TBD
 ### Fixed
+- wrap: support wrapping based on `wrapper_config` [#685]
 - FSDP: fix extra process groups being created by default. Old behavior can cause excessive GPU memory usage. [#678]
 - FSDP: fix forward pass not overlapping compute and allgather [#671]
 - FSDP: improved frozen weight support [#657]
--- a/fairscale/nn/__init__.py
+++ b/fairscale/nn/__init__.py
@@ -10,6 +10,6 @@ from .data_parallel import FullyShardedDataParallel, ShardedDataParallel
 from .misc import FlattenParamsWrapper
 from .moe import MOELayer, Top2Gate
 from .pipe import Pipe, PipeRPCWrapper
-from .wrap import auto_wrap, default_auto_wrap_policy, enable_wrap, wrap
+from .wrap import auto_wrap, config_auto_wrap_policy, default_auto_wrap_policy, enable_wrap, wrap
 
 __all__: List[str] = []
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -218,7 +218,7 @@ class FullyShardedDataParallel(nn.Module):
         cpu_offload (bool, Optional):
             if ``True``, offload FP32 params to CPU. This is only relevant when
             *``mixed_precision``* is ``True``. Note: This arg will be deprecated in favor of
-            *``move_params_to_cpu``* in an upcoming release. 
+            *``move_params_to_cpu``* in an upcoming release.
     """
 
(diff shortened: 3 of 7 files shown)


sshleifer (commented):
Will this effect fairseq things?

sshleifer (approved):

min-xu-ai:
> Will this affect fairseq things?

No, it should still work. The unit tests that cover existing use cases still pass.

Thanks for the review!

QuentinDuval:
I had a different API in mind which I think could be great. The idea would be to propose an auto_wrap policy combiner.

It would look like this on the call site:

```
with enable_wrap(
    auto_wrap_policy=CombinedPolicy([
        (bn_wrapping_policy, bn_wrapping_config),
        (big_conv_wrapping_policy, big_conv_wrapping_config)
    ]),
    wrapper_cls=FSDP,
    **common_fsdp_config
):
    return auto_wrap(module)
```

How it would work:
- the `bn_wrapping_config` and `big_conv_wrapping_config` would be merged with `common_fsdp_config` depending on which wrap is done (which policy matches).
- the `CombinedPolicy` would keep track of which policy is still active at a given depth (if `big_conv_wrapping_policy` says not to recurse but `bn_wrapping_policy` says it should, then recurse but only with `bn_wrapping_policy` active)
- the `auto_wrap` would make sure that inner wrapping is done before outer wrapping (because of the `self._params` in FSDP, we have a bug is we wrapper an outer layer before an inner layer)

It requires quite a bit of rework, but I think the API would be quite nice.

min-xu-ai:
@QuentinDuval and I chatted and I have added 2 more commits for improve this PR. The more ambitious API is TBD in a future PR if we decide to do it.

PR #691, opened 2021-05-20
[refactor] ShardedGradScaler init and super call

ncilfone:
## What does this PR do?
Makes ShardedGradScaler __init__ mirror GradScaler so super can forward parameters. Without it one cannot configure a ShardedGradScaler object like one can with the PyTorch native GradScaler object.


fairscale/optim/grad_scaler.py
```diff
@@ -30,8 +30,22 @@ class ShardedGradScaler(TorchGradScaler):
...
+        enabled: bool = True,
+        process_group: Any = dist.group.WORLD,
+    ) -> None:
+        super().__init__(
+            init_scale=init_scale,
+            growth_factor=growth_factor,
+            backoff_factor=backoff_factor,
+            growth_interval=growth_interval,
+            enabled=enabled,
+        )  # type: ignore
```
min-xu-ai, line 48: Looks great to me! Thanks for sending in the PR.

The type: ignore seems to be causing issue with py37 & mypy on CI. @ncilfone Can you see the CI error?

cc: @blefaudeux

min-xu-ai (approved):

fairscale/optim/grad_scaler.py
ncilfone: Yeah I see the error on my end when I look at the CircleCI details. mypy test runs cleanly locally for me (OSX + Python 3.8.5) with the `type: ignore` flag but that py37 CI build seemingly doesn't want to work...

ncilfone: @min-xu-ai Looks like a single test of the ~500 is failing for gpu_tests_181-3 on CI as well... also seems a little out of the ordinary

min-xu-ai: That's a timeout, might be a CI infra issue. I have triggered a rerun. The py37 issue might be real and not a flaky one. Once CI passes, I can merge this.

blefaudeux (approved):
LGTM, thanks a lot for the PR !

fairscale/optim/grad_scaler.py
min-xu-ai: Seems that the mypy error is real. Nicolas, I can't really checkout your branch and try to fix it. Maybe this is a py37 issue? Do you have 3.7 locally to try it? If not, I can dig it a bit and in the worse case, we can't disable mypy for 3.7.

ncilfone: Working on debugging locally... will let you know

ncilfone: @min-xu-ai New push should hopefully resolve the mypy issue. Didn't realize there were stubs for `torch.cuda.amp.GradScaler`. Adding a stub for `__init__` resolved the issue locally for me on Python 3.7.9

ncilfone: @min-xu-ai Actually I've made a dumb mistake here by amending to my OG commit and force pushing to master on my fork. Seems like circle CI is used a cached version of the checkout (since it probably has the same SHA). Can you force a re-run without the cache? I can't on my end in CircleCI...

min-xu-ai:
merged, thanks!

Issue #644, opened 2021-04-30
[FSDP] Expose the named_shards_parameters when using flatten_params=True


prigoyal:
The goal is to be able to get information about original parameters. This needs to be handled carefully in FSDP:

FSDP sharding should incorporate the "view" into the sharding and share the final metadata keeping in mind a) flattening or not b) padding or not

When we construct the `optimizer = nn.SGD(model.parameters(), ...)` we should call `optimizer = nn.SGD(model.named_shards_parameters(), ....)` instead.

min-xu-ai:
We can close this since we have both #696 and #697?

PR #692, opened 2021-05-21
[Draft for discussion] Provide shard views on the original parameters

QuentinDuval:
## What does this PR do?

CC: @prigoyal @min-xu-ai @myleott 

Offer a solution to https://github.com/facebookresearch/fairscale/issues/644.

The PR provides a new function `named_shard_view_parameters` in `FullyShardedDataParallel` which:
- unbundles the `flat_params` for `flatten_parameters=True`
- removes the padding in the parameters for `flatten_parameters=False`

The goal is to have the ability to call `optim.SGD(model.named_shard_view_parameters(), ....)` in order to, for instance, apply some different norm-based adaptative learning rate (such as LARC) without having to:
- being forced to use `flatten_parameters=False`
- worry about the influence of padding even with `flatten_parameters=False`

### Views are no compatible with optimizers

Offering the view is not so hard, but the "views" are currently unusable for the optimiser: views are created by selecting a part of the `flat_param` and these operations are recorded by the autograd engine:
- the views are not leaves: this will trigger assertions in all optimizers
- because they are not leaves, their `grad` parameter is not kept: so optimisers cannot work with them

### Accessible views are not the same across GPUs

In addition to the issue above, there is an interesting side effect with `flatten_parameters=True`: based on how the sharding works, views will not be similar in all shards!

For instance, if the following model is ran on 2 GPUs:

```
a = nn.Linear(2, 1, bias=False)  # parameter size is 2
b = nn.Linear(1, 3, bias=False)  # parameter size is 3
model = FullyShardedDataParallel(
    nn.Sequential(a, b),
    flatten_parameters=True
)
```

In the above model, the parameters of "a" and "b" are fused into flat_param of size 5, then divided in 2 (size 3 and size 2):

- Shard at rank 0 will contain the full parameters for `a` (size 2) and the start of parameter `b` (size 1)
- Shard at rank 1 will contain the remaining of parameter `b` (remaining size 2)

And so if we call the method `named_shard_view_parameters`, the outputs will be:
- Rank 0: [("0.weight", torch.Size([2])), ("1.weight", torch.Size([1]))]
- Rank 1: [("1.weight", torch.Size([2]))]

This makes working with the view rather challenging !

## Potential solutions

To deal with the leaf issue, we would need to create a wrapper around the tensor that selects a subpart of it and forwards its call to the wrapped tensor (to get the `grad` and `is_leaf` for instance). This is doable.

But I think we need to discuss this issue all together (@prigoyal, @min-xu-ai, @myleott) because even if it works, I am not sure the feature is so easy to work with (cf my remarks above on each shard having different views)

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -2,9 +2,10 @@
 #
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
-
+import bisect
 import contextlib
 import copy
+import math
 from enum import Enum, auto
 import functools
 import logging
@@ -14,6 +15,7 @@ import traceback
 import typing
 from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, List, Mapping, NamedTuple, Optional, Set, Tuple, Union
 
+import numpy as np
 import torch
 from torch.autograd import Variable
 import torch.distributed as dist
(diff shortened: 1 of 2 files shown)


min-xu-ai:
Thanks for attempting this! I think this requires more thinking and discussions. Like you have listed above, the presence of the 2 issues, esp. the second one IMHO, make this not really useful.

Maybe we need to go back to #644 and think what do we really need? I am not sure exposing the views or leaf-params is necessarily the right solution since the parameters are different at each GPU.

Some alternative directions I have given little thoughts:

1. can a custom optimizer work better with the shards? It can try to perform minimal communication and gather needed per-full-param info to assist the shard-optimizer-update?
2. what is LARC trying to solve for SSL? Is it for avoid divergence in the beginning of the training when # of GPUs is high? If so, can adapative methods like AdaScale help since it will adjust the LR scaling rate with a potentially more stablizing scaling factor?

For 1, LARC needs both weight and grad norm and then adjust per-param LR. I think the weight and grad norms are collectable by FSDP. Perhaps FSDP can keep weight and grad norms in the same sized tensors as the param. That will 3X the param memory, but since it is sharded, perhaps OK. The optimizer perhap can use those norm tensors and update the flatten param again in a point-wise fashion?

QuentinDuval:
Following discussion with @min-xu-ai, we decided to have a more direct support of our use cases through issues:

- https://github.com/facebookresearch/fairscale/issues/696
- https://github.com/facebookresearch/fairscale/issues/697

PR #694, opened 2021-05-22
[perf] SyncBatchNorm: avoid 2nd set of all_reduce when wrapped by checkpoint_wrapper

msbaines:
This change also ensure that we calculate running_{mean,var} correctly
when wrapped.

## What does this PR do?
Fixes # (issue).

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/experimental/nn/sync_batchnorm.py
+++ b/fairscale/experimental/nn/sync_batchnorm.py
@@ -124,6 +124,11 @@ class SyncBatchNorm(torch.nn.BatchNorm2d):
         self.disable_patch_batchnorm = True
 
     def forward(self, input: Tensor) -> Tensor:  # type: ignore
+        # There are 3 modes this is being called:
+        # 1. not wrapped (and there is only a single phase)
+        # 2. wrapped and in checkpointing phase
+        # 3. wrapped and in recomputing phase
+
         if not dist.is_initialized() or not self.training:
             return super().forward(input)
 
--- a/fairscale/experimental/nn/sync_batchnorm.py
+++ b/fairscale/experimental/nn/sync_batchnorm.py
@@ -3,80 +3,60 @@
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple, cast
 
 import torch
+from torch import Tensor
 import torch.distributed as dist
 from torch.distributed import ProcessGroup
 
 
-def _forward(
-    input: torch.Tensor,
-    affine: bool,
-    track_running_stats: bool,
-    mean: torch.Tensor,
-    var: torch.Tensor,
-    invstd: torch.Tensor,
-    momentum: float,
-    weight: torch.Tensor,
-    bias: torch.Tensor,
-    running_mean: torch.Tensor,
-    running_var: torch.Tensor,
-    total_count: torch.Tensor,
-) -> torch.Tensor:
-    if track_running_stats:
-        with torch.no_grad():
-            unbiased_var = var * (total_count / (total_count - 1))
-            running_mean += momentum * (mean.reshape(-1) - running_mean)
-            running_var += momentum * (unbiased_var.reshape(-1) - running_var)
-
+def _forward(input: Tensor, affine: bool, mean: Tensor, invstd: Tensor, weight: Tensor, bias: Tensor) -> Tensor:
     if affine:
         return (input - mean) * (invstd * weight.reshape_as(mean)) + bias.reshape_as(mean)
     else:
         return (input - mean) * invstd
 
 
+def _track_running_stats(
+    running_mean: Tensor, running_var: Tensor, momentum: float, mean: Tensor, var: Tensor, total_count: Tensor
+) -> None:
+    with torch.no_grad():
+        unbiased_var = var * (total_count / (total_count - 1))
+        running_mean += momentum * (mean.reshape(-1) - running_mean)
+        running_var += momentum * (unbiased_var.reshape(-1) - running_var)
+
+
+def _calculate_stats(input: Tensor, eps: float, process_group: ProcessGroup) -> Tuple[Tensor, Tensor, Tensor, Tensor]:
+    dim = [d for d in range(input.ndim) if d != 1]
+    count = torch.full((1,), input.numel() // input.size(1), device=input.device, dtype=input.dtype)
+    total_count = count.clone()
+    all_reduce_handle = dist.all_reduce(total_count, group=process_group, async_op=True)
+    mean = torch.mean(input, dim=dim, keepdim=True)
+    meansqr = torch.mean(input * input, dim=dim, keepdim=True)
+    vec = torch.cat([mean, meansqr])
+    all_reduce_handle.wait()
+    vec = vec * (count / total_count)
+    dist.all_reduce(vec, group=process_group)
+    mean, meansqr = vec.chunk(2)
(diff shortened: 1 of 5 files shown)


fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -138,22 +118,27 @@ def __init__(
...
         super().__init__(*args, **kwargs)  # type: ignore
         self._process_group = process_group if process_group is not None else dist.group.WORLD
+        self.saved_for_2nd_fwd: Optional[Tuple] = None
 
-    def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
+    def forward(self, input: Tensor) -> Tensor:  # type: ignore
         if not dist.is_initialized() or not self.training:
             return super().forward(input)
 
+        if not hasattr(self, "_checkpoint_fwd_counter") or self._checkpoint_fwd_counter == 0:
```
min-xu-ai, line 127: The `_checkpoint_fwd_counter` currently is used by FSDP for the case where a module is forwarded multiple times in a single outer forward pass. In SyncBN's case, do we actually only want to keep the stats once in that case?

i.e.:

```
class Layer():
    self.bn = sync_bn()
    def forward():
       ...
       x = self.bn(x)
       ...

class Mod():
       self.layer = Layer()
       def forward():
             return layer(layer(x))
```

In the above case, the bn inside layer is called 4 times if `Mod` is wrapped by a checkpoint. In torch BN's case, the running stats will be kept in 2 of the 4 times. In this change, using `_checkpoint_fwd_counter`, however, it will be kept in only 1 out of 4 times. I think 2 is probably the right thing to do.

Maybe, this BN should keep a `track_running_stats` flag like torch's BN? That way, checkpoint_wrapper will just use `patch_batchnorm` to toggle that flag to get the right behavior?

tests/experimental/nn/test_sync_batchnorm.py
```diff
@@ -92,6 +93,19 @@ def parity3d_bn():
...
+def parity3d_checkpoint_syncbn():
+    rank = dist.get_rank()
+    torch.cuda.set_device(rank)
+    torch.manual_seed(rank)
+
+    x = torch.randn(4, 3, 4, 4, 4).cuda() * rank
+    torch_bn = torch.nn.SyncBatchNorm(3).cuda()
+    fs_bn = SyncBatchNorm(3).cuda()
+    fs_bn = checkpoint_wrapper(fs_bn, maintain_forward_counter=True)
+    check_parity_ddp(torch_bn, fs_bn, x)
```
min-xu-ai, line 106: does check parity function actually check the buffer values? I think the running stats are kept in the buffers and they are not used during training and only later get used during inference? Sorry if I misunderstood that part.

fairscale/experimental/nn/sync_batchnorm.py
msbaines: This BN does support track_running_stats. I verified parity is correct for a single forward pass. Did not think about 2 forward passes. Will add a test and code for that.

tests/experimental/nn/test_sync_batchnorm.py
msbaines: check_parity_ddp checks that the buffer values match:

https://github.com/facebookresearch/fairscale/blob/d240b748546445374e3823b4f8c3557963ae77f5/tests/experimental/nn/test_sync_batchnorm.py#L74

msbaines:
Added support for calling SyncBatchNorm twice in a checkpoint_wrapper.

fairscale/nn/checkpoint/checkpoint_activations.py
```diff
@@ -18,6 +19,71 @@
...
+
+            def forward(self, input):
+                if not is_recomputing():
+                    self.counter += 1
+                return input
+
+    Returns:
+        bool: :data:`True` if it's under checkpoint recomputation.
+
+    .. see also:: :ref:`Detecting Recomputation`
```
min-xu-ai, line 81: where is the ref defined? I am not sure it is supposed to link to somewhere else or not.

min-xu-ai (approved):
This is very cool! Some nonblocking suggestions.

fairscale/nn/checkpoint/checkpoint_activations.py
```diff
@@ -18,6 +19,71 @@
 from .checkpoint_utils import dec_counter, inc_counter, init_counter, patch_batchnorm
 
 
+class ThreadLocal(threading.local):
+    def __init__(self) -> None:
+        self.is_checkpointing = False
+        self.is_recomputing = False
+
+
+thread_local = ThreadLocal()
```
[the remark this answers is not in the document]
msbaines: A global would be not be thread-safe. I believe PyTorch also uses a thread-local for this sort of thing.

fairscale/experimental/nn/sync_batchnorm.py
```diff
@@ -138,22 +120,32 @@ def __init__(
     ) -> None:
         super().__init__(*args, **kwargs)  # type: ignore
         self._process_group = process_group if process_group is not None else dist.group.WORLD
+        self.saved_for_2nd_fwd: List[Tuple] = []
+        self.disable_patch_batchnorm = True
```
anj-s, line 124: can we add a comment about what this property is about?

tests/experimental/nn/test_sync_batchnorm.py
```diff
@@ -92,6 +97,34 @@ def parity3d_bn():
...
+
+
+@pg_test()
+def parity3d_checkpoint_syncbn_twice():
```
anj-s, line 114: how does this test 2 forward passes?

msbaines: This tests uses an nn.Sequential to call the same BatchNorm instance twice.

fairscale/nn/checkpoint/checkpoint_activations.py
```diff
@@ -18,6 +19,70 @@
...
+    orig = thread_local.is_checkpointing
+    thread_local.is_checkpointing = True
+    try:
+        yield
+    finally:
+        thread_local.is_checkpointing = orig
+
+
+@contextmanager
+def enable_recomputing() -> Generator[None, None, None]:
```
anj-s, line 45: nit: we should ideally test this in isolation of SyncBN. WDYT?

tests/experimental/nn/test_sync_batchnorm.py
anj-s: ah got it! it's the same instance. For some reason I was thinking of two explicit forward calls and checking stats in between.

fairscale/nn/checkpoint/checkpoint_utils.py
```diff
@@ -42,7 +42,7 @@ def post_forward(module: _BatchNorm, input: Tensor, result: Tensor) -> None:
     hooks = []
     for name, child in module.named_modules():
         # _BatchNorm is base for bn1d, bn2d, bn3d and sync_bn, apex_sync_bn, etc.
-        if isinstance(child, _BatchNorm):
+        if isinstance(child, _BatchNorm) and not hasattr(child, "disable_patch_batchnorm"):
```
anj-s, line 45: again, a comment about this property would help here as well.

anj-s:
@msbaines  this is really nice and simple to understand :) Sorry for the late review. Feel free to address (or ignore if not required) in any later PRs you have in the pipeline.

PR #695, opened 2021-05-25
[Offload][feature] Add auto shard functionality to remove requirement of nn.Sequential models.

anj-s:
## What does this PR do?
Removes the requirement for models to be nn.Sequential when using the OffloadModel API by introducing an auto shard functionality. The auto shard functionality uses torch.fx to trace the model. In trace 1, we analyze the cutpoints in the model that will be sharded and in trace 2 we carry out model sharding by introducing placeholder and output nodes. 

- Added a Transformer model to test this out. There is 1 outstanding TODO and will be adding more tests to iron out this approach.
- Modified the OffloadModel API arg type to accept a list of GraphModules as a way to avoid introducing another parameter until this functionality is more baked in.
- Added unit tests for auto sharding as well as integrating with OffloadModel API.

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,186 @@
...
+
+    traced_graph_module = torch.fx.symbolic_trace(model)
+
+    # Find the total number of params in the model and
+    # the number of params per shard we are aiming for.
+    for named_mods in model.named_modules():
+        sum = 0
+        for x in named_mods[1].parameters():
+            mul_dims = reduce(operator.mul, x.size(), 1)
+            sum += mul_dims
```
msbaines, line 59:
```suggestion
            sum += x.nume()
```

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,186 @@
...
+
+        name = named_mods[0].split(".")[0]
```
msbaines, line 61: Won't this result in double-counting?

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,186 @@
...
+
+            min_shard_id = shard_id
+            # For each of the args of a given node, find the arg that is not the
+            # last node we traversed. This is to help us find skip connections
+            # across shards.
+            for arg in node.args:
+                try:
+                    test = arg.name
+                except AttributeError:
+                    continue
```
msbaines, line 85:
```suggestion
                if not hasattr(arg, "name"):
                    continue
```

msbaines (commented):
Still working my way through the code.

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,186 @@
...
+import torch
+import torch.fx
+from torch.fx.node import Node
+
+
+def _get_count(param_count: Dict, node_name: Any) -> int:
+    """Identify different mutations of a given node name."""
+    if node_name in param_count:
+        return param_count[node_name]
+    elif node_name.split("_")[0] in param_count:
```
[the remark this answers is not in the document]
anj-s: When modules are not explicitly named, the traced graph prepends an "_" to the layer name. I am not happy with this code since I think it may lead to errors. Working on finding a better way to track nodes in the graph. I added a TODO to track this.

fairscale/experimental/nn/auto_shard.py
anj-s: Actually it looks like it will be triple counting :) I don't use the individual node params so this does not impact the result. I did modify this to make it more clear.

fairscale/experimental/nn/auto_shard.py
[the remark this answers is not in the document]
msbaines: A valid name could also have an `_` in it. There could multiple names with a common prefix followed by an `_`.

The `_` handling seems unnecessary. Why is it needed? In the fx examples, a dictionary is created directly from `named_modules`:

```
self.modules = dict(self.mod.named_modules())
```
Reference: https://pytorch.org/docs/stable/fx.html#the-interpreter-pattern

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,189 @@
...
+                if not hasattr(arg, "name"):
+                    continue
+
+                if arg.name in node_name_to_shard_id and arg.name != nodes_so_far[-1]:
+                    if node_name_to_shard_id[arg.name] < min_shard_id:
+                        min_shard_id = node_name_to_shard_id[arg.name]
+                        min_node_name = arg.name
+
+            # If we there is an input that is not from the previous shard as expected,
+            # we collapse all shards in between and update the param count per shard.
```
msbaines, line 90:
```suggestion
            # If there is an input not from the previous shard,
            # collapse all shards in between and update shard_to_param_count.
```

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,189 @@
...
+            nodes_so_far.append(node.name)
+        elif node.op in ["get_attr", "call_function", "call_method", "call_module"]:
+
+            min_shard_id = shard_id
+            min_node_name = ""
+            # For each of the args of a given node, find the arg that is not the
+            # last node we traversed. This is to help us find skip connections
+            # across shards.
+            for arg in node.args:
+                if not hasattr(arg, "name"):
```
msbaines, line 81: When would an `arg` not have a name?

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,189 @@
...
+        param_count[named_mods[0]] = sum
+
+    logging.info(f"Total number of params are {param_count['']}")
+    per_shard_param = param_count[""] // shard_count
+    logging.info(f"Per shard param count {per_shard_param}")
+
+    for node in traced_graph_module.graph.nodes:
+        if node.op == "placeholder":
+            node_name_to_shard_id[node.name] = shard_id
+            nodes_so_far.append(node.name)
```
msbaines, line 72: Why add placeholder nodes to `nodes_so_far`?

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,189 @@
...
+    shard_id = 0
+    nodes_so_far = []
+    param_count: Dict[str, int] = {}
+    shard_to_param_count = {}
+
+    traced_graph_module = torch.fx.symbolic_trace(model)
+
+    # Find the total number of params in the model and
+    # the number of params per shard we are aiming for.
+    for named_mods in model.named_modules():
```
msbaines, line 56: To avoid using `named_mods[0]` and `named_mods[1]`:
```suggestion
    for name, module in model.named_modules():
```

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,189 @@
...
+        sum = 0
+        if "." in named_mods[0]:
+            continue
+
+        for x in named_mods[1].parameters():
+            sum += x.numel()
+        param_count[named_mods[0]] = sum
```
msbaines, line 63:
```suggestion
        param_count[name] = sum([x.numel() for x in module.parameters()])
```

fairscale/experimental/nn/auto_shard.py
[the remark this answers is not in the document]
anj-s: The handling is necessary because we need a way to connect the node we are tracing with the node specified in the param count dictionary. When you symbolically trace a graph, the nodes names assigned need to match the ones from named_modules(). If i can get the param count when I trace the graph that would mean a SSOT for the name.

fairscale/experimental/nn/auto_shard.py
anj-s: if the op has args that are inputs to the function they are referenced as 0 and 1 and don't have a name attribute. Added a comment about this.

fairscale/experimental/nn/auto_shard.py
anj-s: We do need this because we want all nodes in the graph to map to a shard especially if an input is mapped to shard 2 instead of shard 1.

anj-s:
@msbaines  Friendly ping on this PR

fairscale/experimental/nn/auto_shard.py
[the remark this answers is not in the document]
msbaines: The[ code](https://pytorch.org/docs/stable/fx.html#the-interpreter-pattern) linked to above suggests that the names in the graph do match the names from `named_modules()`. Are you seeing something different?

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,188 @@
...
+                        break
+                shard_id = min_shard_id
+                # TODO(anj-s): Find a way to raise an error early if this can cause OOM errors.
+                shard_to_param_count = _create_shard_to_param_count(param_count, node_name_to_shard_id)
+
+            # Update state that is tracking node -> shard id and shard id -> param count.
+            node_name_to_shard_id[node.name] = shard_id
+            nodes_so_far.append(node.name)
+            # TODO(anj): This could just be an update, we don't need to recreate the map.
+            shard_to_param_count = _create_shard_to_param_count(param_count, node_name_to_shard_id)
```
msbaines, line 103: Since we only reference this table once, rather than re-creating this every time, might be simpler to just calculate the param_count on the fly via a `_get_shard_param_count()` method instead of recreating the table.

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,188 @@
...
+    env: Dict[str, Node] = {}
+    new_input_node = None
+    # This is the first pass where we attempt to get a map of where
+    # we need to insert placeholder and output nodes.
+    node_name_to_shard_id = _split_nodes(model, shard_count=shard_count)
+
+    traced_graph_module = torch.fx.symbolic_trace(model)
+
+    # dummy value which indicates that this is the first node.
+    prev_shard_id = 1000
```
msbaines, line 150: Nit: why not use an obvious invalid value like `-1`?

fairscale/experimental/nn/auto_shard.py
```diff
@@ -0,0 +1,188 @@
...
+            # we collapse all the shards in between to be part of 1 shard.
+            # and update the param count per shard accordingly.
+            if min_shard_id < shard_id:
+                for node_name in reversed(nodes_so_far):
+                    node_name_to_shard_id[node_name] = min_shard_id
+                    if node_name == min_node_name:
+                        break
+                shard_id = min_shard_id
+                # TODO(anj-s): Find a way to raise an error early if this can cause OOM errors.
+                shard_to_param_count = _create_shard_to_param_count(param_count, node_name_to_shard_id)
```
msbaines, line 97: This calls seems redundant given that we are calling this method below on line 103.

fairscale/experimental/nn/auto_shard.py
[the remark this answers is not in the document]
anj-s: Thats right. I am seeing something different that needs more investigation and discussion with the torch.fx team.

fairscale/experimental/nn/auto_shard.py
anj-s: This call does not need to create the table again as mentioned in the TODO and hence is not redundant with the create call above. Instead I can just update it and maybe that is what you are also suggesting. I would like to do this in a separate PR since I ran into some issues when using an update instead of create here.

fairscale/experimental/nn/auto_shard.py
anj-s: prev_shard_id needs to be a higher value than the first shard id which is why I picked this.

msbaines (approved):

PR #701, opened 2021-05-28
[fix] using dummy tensor to ensure checkpoint backward pass is called in corner cases

min-xu-ai:
This test seems to reveal a case where activation checkpoint doesn't perform the needed backward pass when the inputs and outputs are enclosed in a list. See the prints below. When both inputs and outputs are in a list, we go through the forward pass (print the y.requires_grad and y.shape twice in two instances of the module). But when we call the backward, the backward function doesn't seem to be called.

```
tests/nn/misc/test_checkpoint_activations.py::test_list_input 
XXX fwd
XXX y False torch.Size([4, 2])
XXX y True torch.Size([4, 2])
XXX loss True
PASSED
```

@myleott @msbaines @blefaudeux @QuentinDuval, do you know what might be going on?

## What does this PR do?
Fixes # (issue).

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


tests/nn/misc/test_checkpoint_activations.py
```diff
@@ -252,3 +252,28 @@ def train(model, in1, in2):
...
+            return x
+
+        def forward(self, x):
+            y = []
+            y.append(self.conv(x[0]))
+            print("XXX y", y[0].requires_grad, y[0].shape)
+            return y
+
+    model = nn.Sequential(checkpoint_wrapper(Model().cuda()), Model().cuda())
+    in_data = torch.rand(4, 2).requires_grad_(True).cuda()
```
blefaudeux, line 274: requires_grad is not always enough, if the input is int/long autograd can prune the node from the compute graph and a backward hook would be skipped. Not saying that it's exactly what happens here, just a knee jerk reaction

min-xu-ai: Yeah. Doesn’t seem to be the case here though

min-xu-ai:
interesting, making the input a tensor and wrap it in the module seems to work. See above commit. Its output:

```
tests/nn/misc/test_checkpoint_activations.py::test_list_input XXX fwd
XXX y False torch.Size([4, 2])
XXX y True torch.Size([4, 2])
XXX loss True
XXX bwd
XXX y True torch.Size([4, 2])
PASSED
```

tests/nn/misc/test_checkpoint_activations.py
```diff
@@ -252,3 +252,30 @@ def train(model, in1, in2):
...
+            super().__init__()
+            self.conv = nn.Linear(2, 2)
+
+        def forward1(self, x):
+            x[0] = self.conv(x[0])
+            return x
+
+        def forward(self, x):
+            if not isinstance(x, (list, tuple)):
+                x = [x]
```
anj-s, line 269: This is very interesting! The difference in behavior is between x and [x]. Is that right? Do the `requires_grad` field of the loss tensor change from "False" to "True" when the type() of x changes?

For OffloadModel, I had an issue with triggering the backward hook for int inputs that are common for sequence indices in Transformer models. i use a [dummy input ](https://github.com/facebookresearch/fairscale/blob/master/fairscale/experimental/nn/offload.py#L490) which means I always trigger the backward pass no matter what. Sometimes it might be wasteful if there are no grads to propagate but it handles edge cases and does not require the user to modify the input.

tests/nn/misc/test_checkpoint_activations.py
anj-s: Can you move the input to cuda() and then set the requires_grad field? cuda() should give you a copy of the tensor on the GPU but I've run into instances with Offload where the property of activation tensors was erased when I moved it from CPU->GPU.

tests/nn/misc/test_checkpoint_activations.py
min-xu-ai: Wow, the dummy is very good idea! It seems to work for the test case. Thank you so much @anj-s! Let me test with the real workload.

tests/nn/misc/test_checkpoint_activations.py
min-xu-ai: I have tested this. It seems to make no difference. I think requires_grad is kept when you move the tensor but I do think some other properties are not kept when you move the tensor.

tests/nn/misc/test_checkpoint_activations.py
anj-s: Great! However I am still not sure why the error shows up? From my experience as long as one of the inputs to `forward` is a tensor with requires_grad=True it should work. I wonder why using list causes it to fail.

tests/nn/checkpoint/test_checkpoint_activations.py
```diff
@@ -52,7 +52,7 @@ def get_loss_and_gnorm(model, input):
 class BasicModel(nn.Module):
     """Basic model with a single FFN being checkpointed.
 
-       Used for extensive checkings: equivalency with non-checkpoint, torch-checkpoint, etc.
+    Used for extensive checkings: equivalency with non-checkpoint, torch-checkpoint, etc.
```
min-xu-ai, line 55: seems that black make this formatting change.

tests/nn/checkpoint/test_checkpoint_activations.py
```diff
@@ -267,3 +267,38 @@ def test_deprecated_path():
...
+        def forward(self, x):
+            nonlocal count
+            count += 1
+            y = []
+            for i in x:
+                y.append(self.conv(i))
+            return y
+
+    model = nn.Sequential(checkpoint_wrapper(Model()), Model()).cuda()
+    in_data1 = torch.rand(4, 2).cuda()
```
min-xu-ai, line 294: not setting the inputs's requires_grad here.

tests/nn/misc/test_checkpoint_activations.py
min-xu-ai: yeah, it is more than that. I tested pytorch's checkpoint as well. There, the same issue exists. Worse, when I change the input from list to a tensor and use a list to wrap the tensor inside the forward function, it still doesn't work. I think the dummy tensor method works very well since we don't want to rely on user's input types to get the backward pass called. When we checkpoint, we know we want it called. Also, in the backward function, we do check for all outputs have requires_grad when running under grad_enabled. Therefore, we are pretty safe in detecting unnecessary checkpoint calls.

min-xu-ai:
I have tested this on real workload and it works great. I have cleaned up this PR and it is ready for review. Once merged, the use case in PySlowFast video framework will be able to use this.

tests/nn/checkpoint/test_checkpoint_activations.py
```diff
@@ -267,3 +267,38 @@ def test_deprecated_path():
     # Check if direct import works as before.
     ffn = nn.Sequential(nn.Linear(32, 128), nn.Dropout(p=0.5), nn.Linear(128, 32),)
     ffn = deprecated_checkpoint_wrapper(ffn, {})
+
+
+@skip_if_no_cuda
+def test_list_input():
+    """Test to ensure that we correctly does the checkpoint when the input is a list.
```
anj-s, line 274: nit: rephrase? perhaps replace "correctly does the checkpoint" with "Test checkpointing with input type lists" or something along those lines.

tests/nn/checkpoint/test_checkpoint_activations.py
```diff
@@ -267,3 +267,38 @@ def test_deprecated_path():
...
+
+    Note: Testing shows that PyTorch's torch.utils.checkpoint function does not pass this test.
```
anj-s, line 276: should we follow up about this offline?

tests/nn/checkpoint/test_checkpoint_activations.py
```diff
@@ -267,3 +267,38 @@ def test_deprecated_path():
...
+
+    class Model(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.conv = nn.Linear(2, 2)
+
+        def forward(self, x):
+            nonlocal count
+            count += 1
+            y = []
```
anj-s, line 288: nit: prefer using full names to describe variables instead of single letters.

anj-s (approved):

tests/nn/checkpoint/test_checkpoint_activations.py
min-xu-ai: Sure

tests/nn/checkpoint/test_checkpoint_activations.py
min-xu-ai: I am noting this most for future reference. I think it is better to focus on one checkpoint code instead of two. The other code is useful in debugging in case we found a bug only exists in this implementation not the other.  I think we might want to update the doc on the pytorch version and pointing users to use this version going forward.

tests/nn/checkpoint/test_checkpoint_activations.py
min-xu-ai: In general I agree. Here however it is conventional to use `x` for the forward input. I think in terms of math, using `y` in this simple case is also acceptable. In bigger and more complicated function, I would not use y.

PR #719, opened 2021-06-22
checkpoint_activations: use non blocking cpu transfer

sshleifer:
Use `x.to('cpu', non-blocking=True)` for potential activation checkpointing speedup.


tmarkstrum:
LGTM.

anj-s:
Hey @sshleifer, thank you for the PR. Can we add a description? Thanks!

sshleifer:
@anj-s will the speed benchmarks be able to detect the impact of this change?

anj-s:
> @anj-s will the speed benchmarks be able to detect the impact of this change?

I think I missed this question. We currently don't have any benchmarks so we can't assess the impact of this change.

PR #721, opened 2021-06-23
Make sure requires_grad of FlatParameter to be consistent with requires_grad of original parameters

zhaojuanmao:
While flatten original parameters to the FlatParameter, make sure requires_grad of original parameters are passed to FlatParameter. If requires_grad of original parameters are not consistent, assert errors.

### DIFF
--- a/fairscale/nn/misc/flatten_params_wrapper.py
+++ b/fairscale/nn/misc/flatten_params_wrapper.py
@@ -182,7 +182,10 @@ class FlattenParamsWrapper(nn.Module):
         # Init all flat_params.
         for new_p_set in self._param_sets:
             params = self._init_flatten_params(new_p_set)
-            flat_param = FlatParameter(params)
+            assert (
+                len(set(p.requires_grad for p in params)) == 1
+            ), "expects all parameters in the same parameter group of the module to have same requires_grad"
+            flat_param = FlatParameter(params, params[0].requires_grad)
             self.flat_params.append(flat_param)
 
         self._flatten_params(self.flat_params)
--- a/tests/nn/data_parallel/test_fsdp_freezing_weights.py
+++ b/tests/nn/data_parallel/test_fsdp_freezing_weights.py
@@ -11,6 +11,7 @@
 
 
 from enum import Enum
+from itertools import product
 import tempfile
 
 import pytest
@@ -25,7 +26,7 @@ from fairscale.utils.testing import dist_init, objects_are_equal, rmf, skip_if_s
 
 
 class Model(nn.Module):
-    def __init__(self):
+    def __init__(self, with_fsdp, freeze_after_wrap_fsdp):
         super().__init__()
         self.trunk = nn.Sequential(
             nn.Conv2d(3, 64, kernel_size=3),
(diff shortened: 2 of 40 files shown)


fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -182,7 +182,10 @@ def __init__(self, module: nn.Module, param_list: ParamGroups = None):
         # Init all flat_params.
         for new_p_set in self._param_sets:
             params = self._init_flatten_params(new_p_set)
-            flat_param = FlatParameter(params)
+            assert (
+                len(set(p.requires_grad for p in params)) == 1
+            ), "expects all parameters in the same parameter group of the module to have same requires_grad"
+            flat_param = FlatParameter(params, params[0].requires_grad)
```
min-xu-ai, line 188: Nice fix! Just to confirm, without this line, what's the error msg from the unit test?

tests/nn/data_parallel/test_fsdp_freezing_weights.py
```diff
@@ -34,38 +35,41 @@ def __init__(self):
...
 
     def forward(self, x):
         return self.head(self.trunk(x))
 
-    def _create_block(self, in_channels, out_channels, with_fsdp):
+    def _create_block(self, in_channels, out_channels, with_fsdp, freeze_after_wrap_fsdp):
         block = nn.Sequential(nn.Conv2d(in_channels, out_channels, kernel_size=3), nn.ReLU(inplace=True),)
-        if with_fsdp:
+        if with_fsdp and freeze_after_wrap_fsdp:
             block = FSDP(block)
```
min-xu-ai, line 64: I think this line end up being different between frozen before and frozen after. I suggest we do the following:

```
model:
  def __init__:
    ...
    if with_fsdp and freeze_after_wrap_fsdp:
       self.fsdp_wrap()

  def fsdp_wrap():
    do some wrapping


... later in the test ...

  if not freeze_after_wrap_fsdp:
    model.fsdp_wrap()
  model = fsdp(model)
```

I think this way, we will ensure the wrapping happens in the same function in the same way in both cases. What do you think?

min-xu-ai (commented):
This is really nice! Thanks for the investigation and fix!

There might be some formatting errors with the PR. I have an alias like this:

alias fs_test='black . && isort && flake8 && mypy --ignore-missing-imports --scripts-are-modules --pretty . && echo ALL GOOD'

I run it before committing. Another way is to setup pre-commit hook as well.

fairscale/nn/misc/flatten_params_wrapper.py
zhaojuanmao: error msg is like this, indicating gradients are not calculated correctly

E       AssertionError: For dict key 'trunk.0.weight': With rtol=0.0001 and atol=1e-05, found 1710 element(s) (out of 1728) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.00930418074131012 (-0.1777753382921219 vs. -0.16847115755081177), which occurred at index (49, 1, 2, 2).

tests/nn/data_parallel/test_fsdp_freezing_weights.py
zhaojuanmao: sounds good!

zhaojuanmao:
the format error reported in CI seems to be not related. 

fairscale/optim/adascale.py:536: error:
Call
to
untyped
function
"append"
in
typed
context
    ...self._state[name] = np.append(self._state...

sshleifer (approved):

sshleifer (approved):
LGTM pending @min-xu-ai.

For my understanding, can you partially freeze models now and you couldn't before?

min-xu-ai:
> LGTM pending @min-xu-ai.
> 
> For my understanding, can you partially freeze models now and you couldn't before?

Thanks for the review Sam! The key issue here is that if freezing is done before wrapping FSDP loses it. Partial freezing should also work.

min-xu-ai:
> the format error reported in CI seems to be not related.
> 
> fairscale/optim/adascale.py:536: error:
> Call
> to
> untyped
> function
> "append"
> in
> typed
> context
> ...self._state[name] = np.append(self._state...

This is strange. Let me take a look

min-xu-ai:
I have recreated this both locally and on CI. It seems that numpy version is updated and caused this issue to happen.

```
(venv) circleci@af7bdc42d86e:~$ pip list|grep numpy
numpy              1.21.0
```

Thanks @msbaines for tips on ssh access of the CI. cc @anj-s @zhaojuanmao @pbelevich

zhaojuanmao:
@min-xu-ai  I see, thanks for looking into it. for this numpy update, do I need to do anything..? or I will just go ahead to merge this PR?

anj-s:
> @min-xu-ai I see, thanks for looking into it. for this numpy update, do I need to do anything..? or I will just go ahead to merge this PR?

Fix for this issue is in this [PR](https://github.com/facebookresearch/fairscale/pull/731). Let's merge once I check this in.

anj-s:
> > @min-xu-ai I see, thanks for looking into it. for this numpy update, do I need to do anything..? or I will just go ahead to merge this PR?
> 
> Fix for this issue is in this [PR](https://github.com/facebookresearch/fairscale/pull/731). Let's merge once I check this in.

Just check in the fix. Please sync to master and merge. Thank you!

min-xu-ai:
merged. Thanks @zhaojuanmao

stephenroller:
This PR broke FSDP in ParlAI -- in particular we have Transformer models that have some fixed parameters (position embeddings). Why is the same requires_grad choice necessary for all parameters? Can't they just be skipped during the flattening?

min-xu-ai:
Hi Stephen,

Thanks for the heads up! We are not aware of that use case. Now that we know it exists, we will add a test case and fix it to prevent regression.

To be clear, what's the right semantics in this case? Are you suggesting that we skip the Params that doesn't have requires_grad when flatten?

Thanks!

zhaojuanmao:
@stephenroller, without this PR, those fixed parameters will be converted to be required_grad=True, and grads of these parameters could be updated. That will be unexpected behavior for your application, right?

@min-xu-ai skip Params with require_grad=False before flatten sound good to me, I can try to add this to see whether it fixes @stephenroller's case

min-xu-ai:
I am curious how did this work originally? My understanding is that it should never worked unless @stephenroller was running with flatten=False. If flatten=True, then his params will not be frozen. FSDP doesn't support skipping some params while flattening yet.

In fact, to support skipping some params while flatten, it is better leave that change to me. I have will have a PR that hooks up FSDP with newer flatten param module class that can more easily do this kind of thing.

stephenroller:
https://github.com/facebookresearch/ParlAI/blob/3e21b4e130049fba4e55a4978faadc5ba504c0d8/parlai/utils/fsdp.py#L71

🤷‍♂️

zhaojuanmao:
> In fact, to support skipping some params while flatten, it is better leave that change to me. I have will have a PR that hooks up FSDP with newer flatten param module class that can more easily do this kind of thing.

sounds good!

min-xu-ai:
> 
> 
> https://github.com/facebookresearch/ParlAI/blob/3e21b4e130049fba4e55a4978faadc5ba504c0d8/parlai/utils/fsdp.py#L71
> 
> 🤷‍♂️

In this case, are you sure the original requires_grad=False was even respected by FSDP before Yanli's change that asserting on it? I doubt it was the case unless I am misunderstanding something.

PR #723, opened 2021-06-24
fixing bug in setting dependencies in partition handler

mrzzd:
removing implicit assumption on matching number of inputs and number of consumers for a partition

### DIFF
--- a/fairscale/experimental/nn/distributed_pipeline/partition_handler.py
+++ b/fairscale/experimental/nn/distributed_pipeline/partition_handler.py
@@ -108,15 +108,13 @@ class DistributedPipelineRecord:
         # with this constraint, replace the condition 'self.rank > 0' below with
         # a more accurate one.
         if chunk != 0 and self.consumers and self.rank > 0:
-            dependant_tensors = []
             batch = self.batches[chunk]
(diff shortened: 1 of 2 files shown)


fairscale/experimental/nn/distributed_pipeline/partition_handler.py
```diff
@@ -104,15 +104,13 @@ def fence(self, chunk: int) -> None:
         # with this constraint, replace the condition 'self.rank > 0' below with
         # a more accurate one.
         if chunk != 0 and self.consumers and self.rank > 0:
-            dependant_tensors = []
             batch = self.batches[chunk]
             assert batch is not None
-            for tensor, remote_ph_list in zip(batch.tensors, self.forwarded_phony[chunk - 1]):
-                dependant = tensor
+            dependant_tensors = list(batch.tensors)
```
anj-s, line 109: Do we have a test that covers this change?

mrzzd: modified the test to cover the situation that is fixed by this PR.

anj-s (approved):

PR #731, opened 2021-06-25
[fix] Add the latest numpy version to requirements.txt

anj-s:
## What does this PR do?
Certain PRs (looks like forks) are running into mypy errors due to a bumped up numpy version that is being pulled in. 
* Bumped up numpy to >=1.21 as a temporary fix. We ideally don't want to have any other dependency other than PyTorch.
* Since np.append is untyped to begin with I did not see any fix other than adding an ignore. Maybe I am missing something? 
* I am still debugging how we are pulling in numpy 1.21 for PRs that are opened outside of the fairscale repo. PRs within the repo are using numpy 1.20. I tried invalidating the cache but that did not make a difference. PRs on different forks don't share a cache with the repo  according to [circleci](https://circleci.com/docs/2.0/caching/#caching-and-open-source). Maybe we want to consider that in the future.

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [x] N/A
  - [x] N/A

## PR review

### DIFF
--- a/fairscale/optim/adascale.py
+++ b/fairscale/optim/adascale.py
@@ -533,7 +533,7 @@ class AdaScale(Optimizer):
                 continue
             # must be a np array, extend it with the right value and check the shape.
             val = 1 if name == "grad_sqr_avg" else 0
-            self._state[name] = np.append(self._state[name], val)
+            self._state[name] = np.append(self._state[name], val)  # type: ignore
             assert self._state[name].shape == (len(self._optimizer.param_groups),)
 
     def zero_grad(self) -> None:
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 # FairScale should only depends on torch, not things higher level than torch.
 torch >= 1.6.0
+numpy >= 1.21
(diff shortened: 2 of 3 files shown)


min-xu-ai (approved):
LGTM

omry:
More info:
numpy dropped Python 3.6 support in version 1.20.
As of now, fairseq still supports Python 3.6.

This diff makes Python 3.7 a hard requirement for fairscale, and thus for fairseq which depends on it.

Since the motivation here seems to be typing, I don't think it's worth changing the minimal supported Python version for fairseq and fairscale for it.

min-xu-ai:
Thanks for the heads up, @omry! How about we add the requirement to requirements-test.txt? That should make our CI happy without require users to have newer numpy. @anj-s, what do you think?

omry:
I get an uneasy feeling thinking that the CI will test on versions other than the official minimal versions.

anj-s:
Thank you for the headsup @omry  As you can see in the PR description I am still debugging why some users ran into this. This was only a quick fix to unblock users.
@min-xu-ai  your suggestion sounds good. Opened a PR to make this change since we don't want to break fairseq.

anj-s:
Merged [PR](https://github.com/facebookresearch/fairscale/pull/732). @omry  Let me know if this unblocks you. Thanks!

Issue #697, opened 2021-05-25
Flatten parameters by group


QuentinDuval:
## 🚀 Feature

FSDP to offer the possibility to flatten parameters by group, for instance, to flatten all biases separately from the other weights.

## Motivation

Following issue https://github.com/facebookresearch/fairscale/issues/644 and the attempt in solving it in PR https://github.com/facebookresearch/fairscale/pull/692, providing a view of the parameters in the view is too low level and does not offer the best user experience as there are plenty of ways to trip and fall (see PR for the limitations: https://github.com/facebookresearch/fairscale/pull/692).

Following a discussion with @min-xu-ai, the best is to go back at the uses cases motivating https://github.com/facebookresearch/fairscale/issues/644:

1. be able to easily compute the weight and gradient norms when `flatten_parameters=True` needed for LARC like optimisers
2. be able to have separate LR, regularisation for each parameters (for instance, regularise only bias and not weight or the other way around)

This issue proposes to solve item 2.

## Workarounds

There is no workaround for the moment when `flatten_parameters=True`: having different wrappers for different modules does not offer the granularity required to flatten biases and weights of the `nn.Linear` layer separately for instance.

## Interested parties

CC: @min-xu-ai @myleott @prigoyal

min-xu-ai:
Do you need this first or #696 first?

QuentinDuval:
> Do you need this first or #696 first?

Since we have a workaround for LARC (and training is ongoing with it), I think this issue is more important.

jramapuram:
Any update on this? LAMB/LARC/LARS/AGC all rely on being able to set layer-wise scaling.

min-xu-ai:
Thanks for pinging. I haven't got time working on this lately. I will try to find some time again for it. The underlying code for multiple flatten groups is already there. But the API to expose it in terms of reflattening isn't there yet.

jramapuram:
@min-xu-ai : just want to add here that this feature is also necessary for varying weight decay on a different subset of parameter groups (important for transformer scaling for example).

min-xu-ai:
@jramapuram, thanks for checking. I haven't had chance working on this further. But @tmarkstrum had a private branch that can support weights/bias decay differently but it is not general enough. CC @anupambhatnagar in case he has time looking at this.

anupambhatnagar:
I don't think this work is planned for the near future.

PR #746, opened 2021-07-24
[feat]: prepare FSDP to handle multiple flatten params and fixed metadata saving for MoE

min-xu-ai:
## What does this PR do?

This is step 3 to address #697. In this PR, we update FSDP to deal with flatten and nonflatten params in a more uniform fashion. The API is unchanged and still only allow either all flatten or none flatten. But internally, FSDP keeps flatten and nonflatten params in `m.params` in an uniform way.

There is no more `m.module` pointing to a FPW or not. It is always a FPW regardless whether flatten is used. When none of the parameter is flattened, FPW is just a noop wrapper, it simply won't have any FlatParameter objects in its `self.flat_params` list.

After this change, FSDP won't care there is 0 or more flat params and it delegates most flatten related logic to FPW. It retains the sharding related logic as well as logic related to autograd hooks etc.

It was tricky to update `local_metadata_dict` and `consolidate_shard_weights` APIs for this change, but the resulting code is about half as long as before. 

I also added a feature to allow users to name their flat parameters in the future. This will be needed when they want to name different groups of params and later use them in the optimizer.

In the next step, I will add API changes to allow users to group params for flatten for real.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -12,7 +12,21 @@ from math import inf
 import time
 import traceback
 import typing
-from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, List, Mapping, NamedTuple, Optional, Set, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Mapping,
+    NamedTuple,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+    cast,
+)
 
 import torch
 from torch.autograd import Variable
@@ -293,18 +307,22 @@ class FullyShardedDataParallel(nn.Module):
                 params.append(param)
 
         self._has_params = len(params) > 0
-        if not self._has_params:
-            self.flatten_parameters = False
 
+        # For now, it is either all flatten or none. This will be extended to multiple flatten
+        # groups later.
+        to_be_flatten_params: List[List[Parameter]] = [[]]
+        remining_params = params
         if self.flatten_parameters:
-            self._fsdp_wrapped_module: nn.Module = FlattenParamsWrapper(module, param_list=params)
-            del module  # free original module in case it helps garbage collection
-            self.param_paths = ["flat_param"]
-            self.params = [self._fsdp_wrapped_module.flat_param]
-        else:
-            self._fsdp_wrapped_module = module
-            self.param_paths = param_names
-            self.params = params
+            to_be_flatten_params = [params]
+            remining_params = []
+
+        self._fsdp_wrapped_module: nn.Module = FlattenParamsWrapper(module, param_list=to_be_flatten_params)
+        del module  # free original module in case it helps garbage collection
+
+        # Now, keep a list of to the flatten and remining params for doing sharding, gradient
+        # hooks, etc. in FSDP wrapper.
+        self.params = cast(List[Parameter], self._fsdp_wrapped_module.flat_params) + remining_params
+        self._num_flatten_params = len(self._fsdp_wrapped_module.flat_params)
 
         # Shard module parameters in place
         self._shard_parameters_()
@@ -367,8 +385,10 @@ class FullyShardedDataParallel(nn.Module):
         self.gradient_postdivide_factor = post
 
     @property
-    def module(self) -> nn.Module:
-        return self._fsdp_wrapped_module  # note: may be a FlattenParamsWrapper instance
+    def module(self) -> FlattenParamsWrapper:
+        """ make model.module accessible, just like DDP. """
+        assert isinstance(self._fsdp_wrapped_module, FlattenParamsWrapper)
+        return self._fsdp_wrapped_module
 
     def apply(self, fn: Callable[[nn.Module], None]) -> "FullyShardedDataParallel":
         """
@@ -668,11 +688,7 @@ class FullyShardedDataParallel(nn.Module):
                 state_dict = super().state_dict(*args, **kwargs)
         else:
             maybe_cast_buffers(torch.float32)
-            if self.flatten_parameters:
-                assert isinstance(self.module, FlattenParamsWrapper)
-                state_dict = self.module.flat_state_dict(*args, **kwargs)
-            else:
-                state_dict = super().state_dict(*args, **kwargs)
+            state_dict = self.module.flat_state_dict(*args, **kwargs)
 
         if self.move_params_to_cpu:
             for k in state_dict.keys():
@@ -827,13 +843,15 @@ class FullyShardedDataParallel(nn.Module):
             full_tensors = self._rebuild_full_params(force_full_precision=True)
             assert full_tensors is not None
             with contextlib.ExitStack() as stack:
-                if self.flatten_parameters and self.module.is_flattened:
+                if self.module.is_flattened:
                     # Update flattened views to point to fully-sized tensors. We
-                    # use self.params[0] instead of full_tensors since the
+                    # use self.params instead of full_tensors since the
                     # latter may contain padding.
-                    assert len(self.params) == 1
-                    assert isinstance(self.module, FlattenParamsWrapper)
-                    stack.enter_context(self.module.unflatten_params(flat_params=[self.params[0]]))
+                    stack.enter_context(
+                        self.module.unflatten_params(
+                            flat_params=[p.data for p in self.params[: self._num_flatten_params]]
+                        )
+                    )
                 try:
                     yield
                 finally:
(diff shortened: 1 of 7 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -293,18 +307,31 @@ def __init__(
...
-            self.params = params
+            to_be_flatten_params = [params]
+            non_flatten_params = []
+            param_name_groups = [param_names]
+        del param_names
+
+        self._fsdp_wrapped_module: nn.Module = FlattenParamsWrapper(module, param_list=to_be_flatten_params)
+        del module  # free original module in case it helps garbage collection
+
+        # Now, in this FSDP wrapper class, we keep a list of to-be-flatten and remining params
```
sshleifer, line 325: (typo) remining

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1529,58 +1560,36 @@ def _free_full_params(self, params: Optional[List[Parameter]] = None) -> None:
...
-            # used in the FlattenParamsWrapper
-            else:
-                param_names = []
-                for module_path, param_name in m.param_path_infos:
-                    full_param_path = module_path + "." + param_name if module_path else param_name
-                    param_names.append(_clean_path(full_param_path))
-                params_metadata.append(
-                    {
-                        "fsdp_path": _clean_path(path),
-                        "is_flat": True,
```
sshleifer, line 1570: Is it an issue that we don't retain the 
`no_broadcast_optim_state` and `is_flat` keys?

I am about to integrate `local_metadata_dict`/`consolidate_shard_weights` into `fairseq:gshard` which will get merged into `fairseq:master` soon, so it would be good if there was test coverage for `MixtureOfExperts` in `test_fsdp_metadata.py`.

min-xu-ai: The is_flat entry is no longer needed. The no_broadcast flag was not used by the consolidation function.

Can you share a test case for MoE? I can add it to the unit test and make sure it is covered. For this PR, I made sure the metadata and consolidation functions are consistently updated and unit test was passing. But it is possible moe case is somewhat special?

sshleifer: Its broken before your change :( 
https://github.com/facebookresearch/fairscale/pull/747
we definitely need `no_broadcast_optim_state`. We can probably do without `is_flat`.

sshleifer: It's also broken for vanilla transformers.

I'm looking at it today, but won't be offended if this gets merged while I'm debugging :)

min-xu-ai: Thanks for the PR with supplied tests! I am not in a hurry to merge. I will try to test my new changes with your tests. If making the current master working with your test is a lot of work, there is no need for you to spend too much time on it since I am changing this code anyway. It is better for me to make the new code work with the supplied test cases.

min-xu-ai: > 
> 
> It's also broken for vanilla transformers because it doesn't consider `shared_param_info`.
> 
> I'm looking at it today, but won't be offended if this gets merged while I'm debugging :)

That's a great point on shared_param_info. It is now part of the FlatParameter class and FPW proxy the information to the caller. It should be similar to how I handle the param_info data structure. You can directly work on my branch here if you prefer that. I do think my current code is cleaner than the master version. :-)

sshleifer: Lol I read this after I fixed :(

sshleifer: pushed fixes here!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1529,58 +1560,45 @@ def _free_full_params(self, params: Optional[List[Parameter]] = None) -> None:
...
+                        backing_param_name = m._param_name_groups[i][0]
+                        names = [backing_param_name]
+                        shapes = [p._orig_size]
+                        numels = [p._orig_size.numel()]
+                    backing_param_name = _clean_path(backing_param_name)
+                    metadata["params"][backing_param_name] = {
+                        "names": [_clean_path(n) for n in names],
+                        "shapes": shapes,
+                        "numels": numels,
+                        "padding_of_backing": m.numel_padded_per_param[i],
```
sshleifer, line 1596: I would just call this padding

the of_backing names make sense internally but not to a random caller reading a returned dictionary.

sshleifer (approved):
LGTM, and now I've played with it!  The sooner you merge the better for me :)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: That's awesome, Sam! Really appreciate the team work here!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -293,18 +307,31 @@ def __init__(
...
-        else:
-            self._fsdp_wrapped_module = module
-            self.param_paths = param_names
-            self.params = params
+            to_be_flatten_params = [params]
+            non_flatten_params = []
+            param_name_groups = [param_names]
+        del param_names
+
+        self._fsdp_wrapped_module: nn.Module = FlattenParamsWrapper(module, param_list=to_be_flatten_params)
```
anj-s, line 322: +1 on this change and making it consistent.

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -115,19 +127,27 @@ class FlattenParamsWrapper(nn.Module):
...
+            The module to wrap.
         param_list (Optional[List[List[nn.Parameter]]]):
-            only flatten parameters appearing in the given groups
+            Only flatten parameters appearing in the given groups.
+            If the param_list is an empty list, then no parameters will get flattened.
+            Note, if a single param is in one of the list, it still get flattened and the
+            original param is removed and replaced with the flatten one.
             Default: None, flatten all parameters (if any)
+        flat_param_names (Optional[List[str]]):
+            orignally, give each flat_param a unique name. Note a "flat_param_"
```
anj-s, line 146: type(orignally)

tests/nn/data_parallel/test_fsdp_metadata.py
```diff
@@ -2,14 +2,23 @@
...
 import torch.distributed as dist
 import torch.multiprocessing as mp
 import torch.nn as nn
+from torch.optim import Adam
 
 from fairscale.nn import FullyShardedDataParallel
 from fairscale.utils.testing import in_temporary_directory, skip_if_single_gpu, temp_files_ctx
+from tests.nn.data_parallel.test_fsdp import DistributedTest, MixtureOfExperts, rename_test, spawn_and_init
+
+USE_TEMPFILE = True  # False for debugging
```
anj-s, line 21: Should we remove this?

anj-s (approved):

min-xu-ai:
Once the CI passes, I will merge this in.

@anj-s, we can make a release today for fairseq after this in. I can make it if you are not already planning to do it. Just let me know. Thanks!

fairscale/nn/misc/flatten_params_wrapper.py
```diff
@@ -220,41 +241,50 @@ def flat_param(self) -> nn.Parameter:
...
                     else:
-                        shared_param_memo[p] = (m, n)
+                        shared_param_memo[p] = (module_name, m, n)
                         param_infos.append((module_name, m, n))
                         params.append(p)
         del shared_param_memo
 
-        assert len(set(p.dtype for p in params)) == 1, "expects all parameters in module to have same dtype"
+        assert len(set(p.dtype for p in params)) == 1, "expects all parameters to have same dtype"
+        assert len(set(p.requires_grad for p in params)) == 1, "expects all parameters to have same requires_grad"
```
zhaojuanmao, line 277: are you planning to skip flattening parameters if requires_grad are not the same in params?

min-xu-ai: no, I am going to assume in the future user will just group their params in that use case. I.e. group the params with requires_grad==True together and requires_grad==False together. Then, this assert won't trigger since those will be different FlatParameters.

PR #749, opened 2021-07-25
Fix shared params and MoE logic

sshleifer:
### Previously

- reconstructed checkpoints would not have the shared parameter. This is different behavior. than `state_dict`.
- fsdp instances with `no_broadcast_optim_state` would still have weights concatenated from all shards. This is different behavior. than `state_dict`. 

### Now
- Behavior in aforementioned cases matches state_dict + test coverage
- we store a few more strings in `FPW._shared_param_info` so that we can easily copy shared parameters to the correct path. I'm fine for this logic to be moved to `FPW.shared_param_paths` or some such.
- 

[1/Aside] `no_broadcast_optim_state` is a bad name now, since we also don't want to broacast weights for these instances. IMO, the flag should be called `is_expert_group` since there is one use case: MoE.


sshleifer:
CI failure is spurious, merging.

PR #752, opened 2021-07-27
FSDP: supporting gradient accumulation without no_sync context manager to save GPU memory

myleott:
Ideally we can do gradient accumulation without the `no_sync` context. In this case it should do the reduce-scatter in the backward, but accumulate into the sharded gradient so that memory usage is small.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1565,7 +1565,7 @@ class FullyShardedDataParallel(nn.Module):
                     p._saved_grad_shard = p.grad.data
                     p.grad = None
                 else:
-                    raise AssertionError(f"unexpected grad shape: {p.grad.size()}"
+                    raise AssertionError(f"unexpected grad shape: {p.grad.size()}")
 
     @torch.no_grad()
     def _free_full_params(self, params: Optional[List[Parameter]] = None) -> None:
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1160,7 +1160,7 @@ class FullyShardedDataParallel(nn.Module):
             else:
                 self._use_full_params()
 
-            # Make sure p.grad has the correct size/device (or set it to None).
+            # Prepare p.grad.
             self._prep_grads_for_backward()
 
         def _register_hook(t: torch.Tensor) -> torch.Tensor:
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -87,6 +87,7 @@ class FullyShardedDataParallel(nn.Module):
     """
     A wrapper for sharding Module parameters across data parallel workers. This
     is inspired by `Xu et al.`_ as well as the ZeRO Stage 3 from DeepSpeed_.
+    FullyShardedDataParallel is commonly shorten to FSDP.
 
     .. _`Xu et al.`: https://arxiv.org/abs/2004.13336
     .. _DeepSpeed: https://www.deepspeed.ai/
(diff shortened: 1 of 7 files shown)


min-xu-ai:
Hi @myleott, I think this is ready. Do you want to give it a spin?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1278,7 +1284,8 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
             # Free full params. As a special case, we don't free the full params
             # when in a ``no_sync`` context (as inversely indicated by
             # ``self._require_backward_grad_sync``), since the params will not
-            # get updated before the next forward.
+            # get updated before the next forward. This saves networking
+            # bandwidth but uses more GPU memory.
             self._free_full_params([param])
```
zhaojuanmao, line 1289: I'm wondering whether full params could be freed and local grad shard is kept in 'no_sync' mode, in this case, 'no_sync' can save GPU memory, right?

min-xu-ai: Good question. In no sync mode, it is both full params and full grads that are consuming the GPU memory.

- We keep full params since the params are not updated while we accumulate the grads
- We keep the full grads since we are not reducing the scatter them (i.e. shard them)

In this new change, we free and reshard full params and shard grads and accumulate into grad shards. So both of the memory overhead above are avoided at the cost of networking overhead.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -783,19 +784,23 @@ def load_local_state_dict(
     @contextlib.contextmanager
     def no_sync(self) -> Generator:
```
sshleifer, line 785: this naming might be canonical and this comment might be out of scope, but `no_grad_sync` would be a clearer name.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1536,10 +1549,23 @@ def _use_full_params(self) -> None:
...
+                    pass
+                elif p.grad.size() == p._fp32_shard.shape:
+                    # This is gradient accumulation without no_sync context.
+                    # We save the grad shard and set p.grad to None for this backward pass.
+                    # We will accumulate after this pass's grad is generated and reduced and
+                    # sharded.
+                    p._saved_grad_shard = p.grad.data
+                    p.grad = None
+                else:
+                    assert False, "unexpected grad shape"
```
sshleifer, line 1568:
```suggestion
                    raise AssertionError(f"unexpected grad shape: {p.grad.size()}"
```

tests/nn/data_parallel/test_fsdp_grad_acc.py
```diff
@@ -17,11 +18,15 @@
 from .test_fsdp import DistributedTest, NestedWrappedModule, rename_test, spawn_and_init
 
 
-class TestNoSync(DistributedTest):
+class TestGradAcc(DistributedTest):
```
sshleifer, line 21: <3 naming

tests/nn/data_parallel/test_fsdp_grad_acc.py
```diff
@@ -17,11 +18,15 @@
...
     def test_transformer(self):
         fn = functools.partial(self._test_transformer, config={})
         spawn_and_init(fn)
 
+    def test_transformer_grad_acc_without_no_sync(self):
+        fn = functools.partial(self._test_transformer, config={}, use_no_sync_context=False)
+        spawn_and_init(fn)
```
sshleifer, line 28: would be cool to assert same loss between the two configs. Since the only tradeoff seems is speed vs. mem.

sshleifer (approved):
LGTM, really clear! 
@myleott ping me if you want me to benchmark this on the fairseq side.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: I think this name comes from DDP, which has this no_sync context. I think we want to keep FSDP and DDP the same. I can certainly change it if people think otherwise.

tests/nn/data_parallel/test_fsdp_grad_acc.py
min-xu-ai: I changed the name since this is test both no_sync and without no_sync now. In both cases, it is testing grad acc. So I named it TestGradAcc. Let me know or just put in a github suggestion for a better name.

tests/nn/data_parallel/test_fsdp_grad_acc.py
min-xu-ai: Sure, let me look into it.

myleott:
This is great, thanks for the quick fix! Tested in fairseq and it's working well. Is it possible to do a new release with this fix?

min-xu-ai:
> 
> 
> This is great, thanks for the quick fix! Tested in fairseq and it's working well. Is it possible to do a new release with this fix?

Of course!

PR #753, opened 2021-07-28
[FSDP] Move final backward callback queueing to pre-backward hook of root instance

zhaojuanmao:
### Summary:

- Move final backward callback queueing to pre-backward hook of root FSDP instance,
so that it is always attached to the outer most backward call of all the modules and fired
after all inner/re-entrant backward calls are completed.

- Added flags to check final backward callback is fired when final
backward callback is required.

- If root FSDP is checkpointed and called multiple times in forward,
checkpointer counter is used to make sure final backward callback is queued 
inside the last inner backward call as well.

### Test Plan:
 unit tests, without this diff, for modified test case `test_nested_all_wrapped_model`, it will fail with following errors as final backward callback is attached to the inner backward call of checkpointed module

```python
dist init r=0, world=1
Asserting FSDP instance is: FullyShardedDataParallel(
  world_size=1, flatten_parameters=True, mixed_precision=True, 
  (_fsdp_wrapped_module): FlattenParamsWrapper(
    (_fpw_module): Linear(in_features=8, out_features=4, bias=True)
  )
)
ERROR: expected to be in states [<TrainingState.BACKWARD_POST: 4>] but current state is TrainingState.IDLE
--------------------------------------------------------- Captured stderr call ----------------------------------------------------------
/home/yanlizhao/venv2/lib/python3.8/site-packages/torch/utils/checkpoint.py:25: UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn("None of the inputs have requires_grad=True. Gradients will be None")
  File "/home/yanlizhao/venv2/lib/python3.8/site-packages/torch/autograd/function.py", line 87, in apply
    return self._forward_cls.backward(self, *args)  # type: ignore[attr-defined]
  File "/home/yanlizhao/fairscale/fairscale/nn/checkpoint/checkpoint_activations.py", line 311, in backward
    torch.autograd.backward(outputs_with_grad, args_with_grad)
  File "/home/yanlizhao/venv2/lib/python3.8/site-packages/torch/autograd/__init__.py", line 147, in backward
    Variable._execution_engine.run_backward(
  File "/home/yanlizhao/venv2/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
  File "/home/yanlizhao/fairscale/fairscale/nn/data_parallel/fully_sharded_data_parallel.py", line 1408, in _wait_for_post_backward
    m.assert_state(TrainingState.BACKWARD_POST)
  File "/home/yanlizhao/fairscale/fairscale/nn/data_parallel/fully_sharded_data_parallel.py", line 1739, in assert_state
    traceback.print_stack()
```

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -371,6 +371,10 @@ class FullyShardedDataParallel(nn.Module):
         # This is reset at the end of the backward pass.
         self._pre_backward_hook_has_run = False
 
+        # Flag to check post backward callback was fired in previous backward pass.
+        # This is checked and reset at each forward pass.
+        self._post_backward_callback_fired = True
+
     def _get_gradient_predivide_factor(self, world_size: int) -> float:
         factor: int = 1
         while world_size % factor == 0 and world_size / factor > factor:
@@ -885,7 +889,6 @@ class FullyShardedDataParallel(nn.Module):
     def _reset_lazy_init(self) -> None:
         """Reset instance so :func:`_lazy_init` will run on the next forward."""
         self._is_root: Optional[bool] = None
-        self._queue_wait_for_post_backward_closure: Optional[Callable] = None
         self._streams: Dict[str, torch.cuda.Stream] = {}
         self._reducer: Optional[ReduceScatterBucketer] = None
         for p in self.params:
@@ -1002,8 +1005,6 @@ class FullyShardedDataParallel(nn.Module):
             return
         # No FullyShardedDataParallel instance wraps this, else _is_root would be set to False.
         self._is_root = True
-        assert self._queue_wait_for_post_backward_closure is None
-        self._queue_wait_for_post_backward_closure = self._queue_wait_for_post_backward
         # As the root, we now set all children instances to False and
         # give them a closure to try to queue a wait_for_post_backward.
         self.children_share_process_group = True
@@ -1015,14 +1016,6 @@ class FullyShardedDataParallel(nn.Module):
                 assert m._is_root is None or not m._is_root
                 if m._is_root is None:
                     m._is_root = False
-                    # When root instance doesn't have params, allow children instances
-                    # to queue the post_backward hook.
-                    #
-                    # TODO (Min): we should think if we can have a empty param at the root
-                    #             so that root always have a callback on the backward graph.
-                    if not self._has_params:
-                        assert m._queue_wait_for_post_backward_closure is None
-                        m._queue_wait_for_post_backward_closure = self._queue_wait_for_post_backward
                 if m.process_group != self.process_group:
                     self.children_share_process_group = False
 
@@ -1129,6 +1122,25 @@ class FullyShardedDataParallel(nn.Module):
 
         return outputs
 
+    def _checkpoint_module_last_backward_call(self):
+        # If this is a checkpointed FSDP module, e.g. checkpoint(FSDP(module)),
+        # we check if the following counter reaches 0. If it is, it is the last
+        # inner backward call for this FSDP module.
+        if (hasattr(self._fsdp_wrapped_module, "_checkpoint_fwd_counter")
+            and self._fsdp_wrapped_module._checkpoint_fwd_counter != 0):
+            return False
+        return True
+
+    def _require_final_backward(self):
+        assert self._is_root
+        for m in self.modules():  # includes self
+            if (isinstance(m, FullyShardedDataParallel)
+                and any(p.requires_grad for p in m.parameters())
+                and self._require_backward_grad_sync
+            ):
+                return True
+        return False
+
     def _register_pre_backward_hooks(self, outputs: Any) -> Any:
         """Register pre-backward hook to run before the wrapped module's
         backward. Hooks should be attached to all outputs from the forward.
@@ -1139,7 +1151,26 @@ class FullyShardedDataParallel(nn.Module):
         if not torch.is_grad_enabled():
             return outputs  # don't register hooks if grad isn't enabled
 
+        if self._is_root:
+            # This actually means that only root instance has this field
+            # defined. Accidentally accessing this field will assert on all
+            # other instances, giving us a nice bug checker.
+            self._post_backward_callback_queued = False
+            # Make sure post backward callback was fired in the preivous iteration
+            assert self._post_backward_callback_fired
+            # Reset flag if require final backwward callback to be fired,
+            # this helps checking whether post backward callback was fired later on
+            if self._require_final_backward():
+                self._post_backward_callback_fired = False
+
         def _pre_backward_hook(*unused: Any) -> None:
+            # try to queue final backward callback only once for root, so
+            # that final backward callback is attached to the outer most
+            # backward graph task and called after all the backward
+            # calls are completed.
+            if self._is_root:
+                self._queue_wait_for_post_backward()
+
             if self._pre_backward_hook_has_run:
                 return  # only run once (from multiple outputs or multiple forward passes)
             self._pre_backward_hook_has_run = True
@@ -1204,11 +1235,6 @@ class FullyShardedDataParallel(nn.Module):
         """
         if not torch.is_grad_enabled():
             return  # don't register grad hooks if grad isn't enabled
-        if self._is_root:
-            # This actually means that only root instance has this field
-            # defined. Accidentally accessing this field will assert on all
-            # other instances, giving us a nice bug checker.
-            self._post_backward_callback_queued = False
         for p in self.params:
             if p.requires_grad:
                 if hasattr(p, "_shard_bwd_hook"):
@@ -1281,14 +1307,6 @@ class FullyShardedDataParallel(nn.Module):
         # Switch to FP32 shard after backward.
         self._use_fp32_param_shard([param])
 
-        # (try to) Enqueue a callback at the end of the backward pass to ensure that all
-        # post-backward work has finished. We only need one callback and all instances
-        # of FSDP (root and children) make this attempt here to queue to ensure it is queued
-        # no matter which instance(s) has(have) params.
-        assert self._queue_wait_for_post_backward_closure is not None or not self._is_root
-        if self._queue_wait_for_post_backward_closure is not None:
-            self._queue_wait_for_post_backward_closure()
-
         if not self._require_backward_grad_sync:
             return
 
@@ -1354,13 +1372,12 @@ class FullyShardedDataParallel(nn.Module):
     def _queue_wait_for_post_backward(self) -> None:
         """Try to queue a `wait_for_post_backward` callback.
 
-        Only called on root and only queue one callback. But can be called by
-        children FSDPs via a closure in case the root instance doesn't own any
-        params.
+        Only called on root and only queue one callback at the beginning of
+        outer most backward.
         """
         assert self._is_root
-        self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
-        if not self._post_backward_callback_queued:
+        if not self._post_backward_callback_queued and self._checkpoint_module_last_backward_call():
+            self.assert_state([TrainingState.IDLE])
             self._post_backward_callback_queued = True
             Variable._execution_engine.queue_callback(self._wait_for_post_backward)
 
@@ -1409,11 +1426,20 @@ class FullyShardedDataParallel(nn.Module):
                     else:
                         m.assert_state(TrainingState.BACKWARD_PRE)
                 else:
-                    # Unlikely case. When `m` and its children has no params or has params but
-                    # none with `requires_grad==True`, then m's pre-backward and post-backward
-                    # hooks aren't called by autograd. Therefore, it is in IDLE state.
-                    m.assert_state(TrainingState.IDLE)
+                    # When `m` and its children has no params or has params but
+                    # none with `requires_grad==True`, it is possible that
+                    # output tensors are `requires_grad==True`. In this case,
+                    # pre-backward hook is still registered and '_wait_for_post_backward'
+                    # is queued in the pre-backward hook.
+                    # Therefore, it is in BACKWARD_PRE state.
+                    m.assert_state(TrainingState.BACKWARD_PRE)
                 m.training_state = TrainingState.IDLE
+
+                if m._is_root:
+                    # reset this flag for cases like "one forward pass + multiple backward passes"
+                    self._post_backward_callback_queued = False
+                    # indicate this post backward callback was fired
+                    self._post_backward_callback_fired = True
 
     @torch.no_grad()
     def _rebuild_full_params(self, force_full_precision: bool = False) -> Optional[List[Tuple[torch.Tensor, bool]]]:
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1005,7 +1005,7 @@ class FullyShardedDataParallel(nn.Module):
         # If final backward callback is queued, the callback should be finished
         # and the state was reset to be IDLE.
         # This should be asserted at the beginning of forward pass in the root instance only.
-        # For children instances, if they are checkpointed, state will not be reset as
+        # For children instances, if they are checkpointed, state will not be reset to
         # IDLE after each inner forward/backward.
         self.assert_state(TrainingState.IDLE)
         # Check if the root instance is being checkpointed. It doesn't make sense to
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1005,8 +1005,11 @@ class FullyShardedDataParallel(nn.Module):
             return
         # No FullyShardedDataParallel instance wraps this, else _is_root would be set to False.
         self._is_root = True
-        # Do not allow to checkpoint root instance and run forward multiple times.
-        assert getattr(self, "_checkpoint_fwd_counter", 0) == 0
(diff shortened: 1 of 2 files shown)


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1129,6 +1122,28 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
 
         return outputs
 
+    def _checkpoint_module_last_backward_call(self) -> bool:
+        # If this is a checkpointed FSDP module, e.g. checkpoint(FSDP(module)),
+        # we check if the following counter reaches 0. If it is, it is the last
+        # inner backward call for this FSDP module.
+        if (
```
sshleifer, line 1129: `return getattr(self._fsdp_wrapped_module, "_checkpoint_fwd_counter", 0) == 0`

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1129,6 +1122,28 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
+            hasattr(self._fsdp_wrapped_module, "_checkpoint_fwd_counter")
+            and self._fsdp_wrapped_module._checkpoint_fwd_counter != 0
+        ):
+            return False
+        return True
+
+    def _require_final_backward(self) -> bool:
```
sshleifer, line 1136: `@property` ?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1139,7 +1154,26 @@ def _register_pre_backward_hooks(self, outputs: Any) -> Any:
         if not torch.is_grad_enabled():
             return outputs  # don't register hooks if grad isn't enabled
 
+        if self._is_root:
+            # This actually means that only root instance has this field
```
sshleifer, line 1158:
```suggestion
            # This actually means that only root instance has _post_backward_callback_queued
```

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -591,12 +591,15 @@ def _maybe_wrap(layer):
...
+        # Test it with checkpoint_wrapper as well to validate final backward callback
+        # is queued correctly when root FSDP does not have any params and every layer is
+        # wrapped as FSDP(checkpoint(module)).
         if wrap_everything:
             self.module = nn.Sequential(
-                _maybe_wrap(nn.Linear(8, 4)),
-                _maybe_wrap(nn.Linear(4, 16)),
-                _maybe_wrap(nn.Linear(16, 4)),
-                _maybe_wrap(nn.Linear(4, 8)),
+                _maybe_wrap(checkpoint_wrapper(nn.Linear(8, 4))),
```
sshleifer, line 599: Should we keep test coverage for the case without checkpoint_wrapper?

sshleifer (commented):
I don't totally understand the bug, left nitpicks.

Is there a way to print out or write the hooks and the order they were called before and after this change?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: I'm not sure whether it should be @property, as it is only called by root

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1129,6 +1122,23 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
 
+    def _checkpoint_module_last_backward_call(self) -> bool:
+        # If this is a checkpointed FSDP module, e.g. checkpoint(FSDP(module)),
+        # we check if the following counter reaches 0. If it is, it is the last
+        # inner backward call for this FSDP module.
+        return getattr(self._fsdp_wrapped_module, "_checkpoint_fwd_counter", 0) == 0
+
+    def _require_final_backward(self) -> bool:
+        assert self._is_root
+        for m in self.modules():  # includes self
```
min-xu-ai, line 1133: I am not sure the purpose of this loop. If you need to check whether any trainable parameters exists inside the root, just m.parameters() would be good enough? No need to loop over all instances?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1139,7 +1149,26 @@ def _register_pre_backward_hooks(self, outputs: Any) -> Any:
         if not torch.is_grad_enabled():
             return outputs  # don't register hooks if grad isn't enabled
 
+        if self._is_root:
+            # This actually means that only root instance has
+            # _post_backward_callback_queued defined. Accidentally accessing this field
+            # will assert on all other instances, giving us a nice bug checker.
+            self._post_backward_callback_queued = False
+            # Make sure post backward callback was fired in the preivous iteration
+            assert self._post_backward_callback_fired
```
min-xu-ai, line 1158: nice check!

min-xu-ai: actually, I am wondering if you can assert this using TrainingState? Just check the state to be BACKWARD_POST should be the same?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1354,13 +1370,12 @@ def _post_reduction_hook(self, param: Parameter, reduced_grad: torch.Tensor) ->
...
-        Only called on root and only queue one callback. But can be called by
-        children FSDPs via a closure in case the root instance doesn't own any
-        params.
+        Only called on root and only queue one callback at the beginning of
+        outer most backward.
         """
         assert self._is_root
-        self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
-        if not self._post_backward_callback_queued:
+        if not self._post_backward_callback_queued and self._checkpoint_module_last_backward_call():
```
min-xu-ai, line 1377: maybe we can simplify things a bit by asserting root module is not checkpointed? It doesn't seem to make a lot of sense to checkpoint the root instance, it would not save any GPU memory. So we can just assert that during __init__, the root's module doesn't have a checkpoint counter attached to it and simplify the logic here?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1409,12 +1424,21 @@ def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
...
                 else:
-                    # Unlikely case. When `m` and its children has no params or has params but
-                    # none with `requires_grad==True`, then m's pre-backward and post-backward
-                    # hooks aren't called by autograd. Therefore, it is in IDLE state.
-                    m.assert_state(TrainingState.IDLE)
+                    # When `m` and its children has no params or has params but
+                    # none with `requires_grad==True`, there are two cases:
+                    # 1. output tensors are `requires_grad==True`. In this case,
+                    # pre-backward hook is still registered, so it is in BACKWARD_PRE state.
+                    # 2. output tensors are `requires_grad==False`. In this case,
```
min-xu-ai, line 1431: Are you sure output tensors have `requires_grad` field? I thought that field is only for params. I am a bit worried that we can see the IDLE state here. That would be a case where the pre-backward callback is not even called?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1409,12 +1424,21 @@ def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
...
+                    # pre-backward hook is not registered, so it is in IDLE state.
+                    m.assert_state([TrainingState.BACKWARD_PRE, TrainingState.IDLE])
                 m.training_state = TrainingState.IDLE
 
+                if m._is_root:
+                    # reset this flag for cases like "one forward pass + multiple backward passes"
+                    self._post_backward_callback_queued = False
+                    # indicate this post backward callback was fired
+                    self._post_backward_callback_fired = True
```
min-xu-ai, line 1440: again, I think the flag might be replaceable by asserting on the TrainingState? Perhaps need to add one more state for before and after the final call back?

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -591,13 +597,24 @@ def _maybe_wrap(layer):
...
         if wrap_everything:
-            self.module = nn.Sequential(
-                _maybe_wrap(nn.Linear(8, 4)),
-                _maybe_wrap(nn.Linear(4, 16)),
-                _maybe_wrap(nn.Linear(16, 4)),
-                _maybe_wrap(nn.Linear(4, 8)),
-            )
+            if checkpoint:
+                self.module = nn.Sequential(
+                    _maybe_wrap(checkpoint_wrapper(nn.Linear(8, 4))),
```
min-xu-ai, line 606: nice for adding more test cases!

min-xu-ai (commented):
This is a really great improvement and potentially make things a lot cleaner than before! Thank you thank you thank you!!!

I have some suggestions and let's iterate on it a bit more to make it more awesome!

min-xu-ai:
cc'ing @anj-s @blefaudeux @tmarkstrum. This is a great PR for enhancing core FSDP & autograd interaction logic to make it work with more model structures and checkpointing configs.

tests/nn/data_parallel/test_fsdp.py
min-xu-ai: I think this is the case where textray people ran into lately.

zhaojuanmao:
@sshleifer, **with this diff, the order of firing hooks and final callback is like this:** 
```
final backward callback is queued
param post hook is fired
param post hook is fired
param post hook is fired
param post hook is fired
param post hook is fired
param post hook is fired
param post hook is fired
param post hook is fired
final backward callback is fired
```

**without this diff, the order is like this:**

```
param post hook is fired
final backward callback is queued
param post hook is fired
final backward callback is fired
Asserting FSDP instance is: FullyShardedDataParallel(
  world_size=1, flatten_parameters=False, mixed_precision=False, 
  (_fsdp_wrapped_module): FlattenParamsWrapper(
    (_fpw_module): Linear(in_features=8, out_features=4, bias=True)
  )
)
ERROR: expected to be in states [<TrainingState.BACKWARD_POST: 4>] but current state is TrainingState.IDLE
```

As you can see, final backward callback is fired earlier without this diff, because final callback is associated with backward graph task.  without this diff, final callback is associated with re-entrant backward call inside checkpoint function; with this diff, final callback is always associated with the out most backward call. 

Is the explanation clearer?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: that is true, we do not need to loop here!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: yes, every tensor could have requires_grad field, pre-backward hooks are registered only to output_tensors with requires_grad=True. 

IDLE state is checked here even before this diff. actually it makes sense to me. for some FSDP layers, if there is no params or no any param with requires_grad = true, the output tensors have usually requires_grad=False, so the pre-backward hook is not registered;  for some rare cases, e..g, in test_fsdp_overlap.py, it intentionally set output tensors with requires_grad = true and return, so pre-backward hook is registered

tests/nn/data_parallel/test_fsdp.py
zhaojuanmao: yes!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: right now, trainingState was reset to be IDLE after final callback is done, let me see whether we can add a new state after the final call back

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: sounds good, let us explicitly not allow to checkpoint a root module!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: had a second thought, I think it makes more sense that trainingState was reset to be IDLE when final callback is done, as many other functions in the codes rely on checking 'IDLE' state to make sure those functions are not called in a forward/backward pass. so it is better to reset state to be IDLE when a pass is done. 

That being said, it is probably better to use flag here.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1002,8 +1005,8 @@ def _set_is_root(self) -> None:
             return
         # No FullyShardedDataParallel instance wraps this, else _is_root would be set to False.
         self._is_root = True
-        assert self._queue_wait_for_post_backward_closure is None
-        self._queue_wait_for_post_backward_closure = self._queue_wait_for_post_backward
+        # Do not allow to checkpoint root instance and run forward multiple times.
```
min-xu-ai, line 1008:
```suggestion
        # Check if the root instance is being checkpointed. It doesn't make sense to
        # checkpoint the root instance since it won't save GPU memory.
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1002,8 +1005,8 @@ def _set_is_root(self) -> None:
...
+        assert getattr(self, "_checkpoint_fwd_counter", 0) == 0
```
min-xu-ai, line 1009:
```suggestion
        assert getattr(self, "_checkpoint_fwd_counter", 0) == 0, "Is the root FSDP module wrapping an activation checkpointed module? If so, please remove that."
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: I see. Make a lot of sense. Perhaps we can ensure the output of root FSDP's always have requires_grad==True? Will that simplify the case and ensure the pre-back is always called?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: Cool. Let me know if this can be done. I think one less bool for checking will help make the code more readable and maintainable.

min-xu-ai (commented):
Yay! The tests are all green! The new code is so much shorter and we can remove so much old hacks. That's very nice! Only a few minor comments and suggestions. The main things left for me are (1) the extra checking boolean and (2) if we can ensure pre-backward is always fired by setting output.requires_grad==True.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: we do not have to require root FSDP's outputs with requires_grad=True, if all outputs of root FSDP are with requires_grad=False, that is fine, in this case, no backward can be called, so no need to register final backward. 

Here if codes are running at this point, that means root FSDP's pre-backward hook has already been fired and final backward callback was fired.  assert_state(IDLE) is for submodules, not for root module

min-xu-ai: OK, make sense. In that case, should we set the fired flag to True in the beginning of the iteration since we *expect* the final callback not being called? In fact, in that case, none of the backward callbacks will be called. It is like a pure inference case or torch.no_grad() case, right?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: had a second thought, I think it makes more sense that trainingState was reset to be IDLE when final callback is done, as many other functions in the codes rely on checking 'IDLE' state to make sure those functions are not called in a forward/backward pass. so it is better to reset state to be IDLE when a pass is done.

That being said, it is probably better to use flag here.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: > OK, make sense. In that case, should we set the fired flag to True in the beginning of the iteration since we _expect_ the final callback not being called?
hmm, which fired flag?

> In fact, in that case, none of the backward callbacks will be called. It is like a pure inference case or torch.no_grad() case, right?
yes

min-xu-ai: sorry, by the "fired flag", I mean the `_post_backward_callback_fired`  boolean. I think I am still confused by this flag. Let's have a chat about it. I will ping you.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1001,6 +1001,10 @@ def _set_is_root(self) -> None:
             return
         # No FullyShardedDataParallel instance wraps this, else _is_root would be set to False.
         self._is_root = True
+        # If final backward callback is never been queued, state should be IDLE.
+        # If final backward callback is queued, the callback should be finished
+        # and the state was reset to be IDLE.
```
min-xu-ai, line 1006: Can you add a comment on why asserting on root only? It is good to assert on root only, but good to know why it is the case.

min-xu-ai (approved):
This is awesome! Approved!

sshleifer:
@zhaojuanmao thanks for the explanation. I get it now!

sshleifer (approved):

PR #755, opened 2021-07-30
[FSDP][feature] Support returning the original parameter names after a model has been wrapped with FSDP

anj-s:
## What does this PR do?
`named_parameters` currently returns the modified names of the parameters for flatten=False and a single `flatten_param_0` (if there is a single parameter group) for flatten=True. This PR returns the original names of the parameters under two conditions:

Use case 1: flatten_params=False
If you call module.named_parameters() it will return the original names of the parameters along with the local shards. If you want the original param, you should call this function under a `summon_full_params` context.

Use case 2: flatten_params=True
If you call module.named_parameters it will original names of the parameters along with the full params only if you are in a `summon_full_params` context.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


anj-s:
There is 1 [test failure](https://app.circleci.com/pipelines/github/facebookresearch/fairscale/3320/workflows/f2ac1157-7b2e-41ef-a4e0-0c84d4a480d2/jobs/32866) because I've changed the contract on the parameter name and have a follow up question regarding returning wrapped parameter names as is when users call `named_parameters`. 

@min-xu-ai  Will setup some time to discuss these points offline as well. Thanks!

min-xu-ai:
Nice PR! Thanks for adding the use case 2 and all_shards option. The original use case is that when user uses this API to match params from a model for freezing, it may fail to match either because the name contains FSDP paths or it was flattened. So I think use case 1 is important in that situation and return the views in the flattened case on the full data might be good. 

I think it is not possible to return views of the original params using the shards since it contains incomplete data. 

Also, I suggest we use “full” vs. “shard” to name the not-sharded and sharded data. Let avoid introducing new names like all-shard or unsharded to avoid confusions?

cc: @QuentinDuval  @zhaojuanmao @myleott

anj-s:
@min-xu-ai Updated the PR as per our offline conversation with the following changes:
- Support use cases that don't require a `summon_full_params` context call. 
- Add comments about not supporting the flatten_params=True when not in a `summon_full_params` context.

anj-s:
@min-xu-ai  Friendly ping on this PR.

min-xu-ai (approved):
nice! Thank for the added test!

PR #760, opened 2021-08-02
Adding missing imports in docstring

mrshenli:
`wrap` from `auto_wrap` is used in the docstring example, but was missed in the imports.

```python
```


anj-s (approved):
Thank you!

mrshenli:
Hey @anj-s, how do I land this PR?

PR #761, opened 2021-08-10
[fix] Add an additional assert for checking if the params of a module requires_grad=True

### LINKED ISSUE #758: FSDP always sets requires_grad to False? Breaking checkpoint_wrapper.
## ❓ Questions and Help

Hi, I get the following error when trying to use FSDP with checkpoint_wrapper: 
RuntimeError:  None of the outputs have requires_grad=True, this checkpoint() is not necessary

I found this weird so I started digging around. I found that in the checkpoint wrapper function, all the input tensors had requires_grad=False stashed in the forward pass which led to this error. I checked that the inputs themselves do have requires_grad=True (at least one of them) and tried to find where this got clobbered. It is happening in the call to cast_floats_to_right_precision(True, True, *args, **kwargs) in the forward method of FullyShardedDataParallel.

I am really confused by this method since it seems that it methodically sets requires_grad to False for all inputs. If relevant, the layer being wrapped with a checkpoint is a Resnet-like block that is an intermediate layer in a network, the layer before it is an un-checkpointed Conv2D. So the input to the block is N x C x H x W and it has requires_grad=True, but when the forward call is dispatched (outputs = self.module(*args, **kwargs)), this tensor now has requires_grad=False. This breaks checkpoint_wrapper :(

Help would be much appreciated!

anj-s:
## What does this PR do?
Potentially fixes #758 

Issue:
The requires_grad for a given model's params were set to False post initialization which means the post backward hooks were never attached and triggered. This condition was not a conditional that we used when asserting on the POST_BACKWARD state which led to the following error:
Error in FSDP:ERROR: expected to be in states [<TrainingState.BACKWARD_POST: 4>] but current state is TrainingState.BACKWARD_PRE

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1417,7 +1421,11 @@ def _remove_shard_bwd_hook(fsdp_module: FullyShardedDataParallel) -> None:
                 _remove_shard_bwd_hook(m)
                 m._pre_backward_hook_has_run = False
                 if any(p.requires_grad for p in m.parameters()):
-                    if m._has_params:
+                    # Check if the module has params and if any of them has
+                    # the `requires_grad` field set. If `requires_grad=False` for
+                    # all the params, the post_backward hook will not fire and the
+                    # state will remain in `TrainingState.BACKWARD_PRE`.
+                    if m._has_params and any([p.requires_grad for p in m.params]):
```
anj-s, line 1428: cc @zhaojuanmao I had to add this condition along with the one above for ensuring that we check the BACKWARD_POST state if the root instance has param and they require grads. There is a use case where they may not. Let me know if this makes sense.

anj-s:
Will add tests once I confirm that the fix works as expected.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1381,7 +1381,11 @@ def _queue_wait_for_post_backward(self) -> None:
     def _wait_for_post_backward(self) -> None:
         """Wait for post-backward to finish. Only called on root instance."""
         assert self._is_root
-        if self._has_params:
+        # Check if the root module has params and if any of them has
+        # the `requires_grad` field set. If `requires_grad=False` for
+        # all the params, the post_backward hook will not fire and the
+        # state will remain in `TrainingState.BACKWARD_PRE`.
+        if self._has_params and any([p.requires_grad for p in self.params]):
```
min-xu-ai, line 1388: perhaps self._has_params is not needed since in that case self.params will just be an empty list?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: same as above, perhaps checking _has_params is not needed?

min-xu-ai (approved):
LGTM! Thanks for fixing them! Minor comments below. If all tests pass, let's merge!

PR #784, opened 2021-09-10
[fix] FSDP intra-backwards gradient accumulation.

### LINKED ISSUE #780: Race condition in gradient reduction
Following up on the previous issue, as I couldn't reopen: https://github.com/facebookresearch/fairscale/issues/777

There seems to be a more fundamental issue here unrelated to mixed precision, and the race condition I mentioned in that issue appears real.

When `method FC` is used, enabling `maintain_forward_counter` in `checkpoint_wrapper` is not optional, as I would expect.

The forward counters seem like a trade-off. With them, you can reduce network overhead at the cost of more VRAM, by retaining the full parameters until the last backward. But in my use case, I have multiple large modules called like Sequence[A, B, C, .., A, B, C...]. I want to reshard after both forward and backward, so using the counters is undesirable.

The unit test fails to capture this error, because the weight tensors are so small that the race condition I was concerned about actually happens. If you make the weights a little larger or force synchronous gradient reduction, the test will fail, without the forward counters. See the diffs below for how to trigger it.

These errors could be avoided with stronger consistency in FSDP's internal state. In my code, I partially avoid the problem by using `method CF` with some small hacks, but the race condition is still technically present. Perhaps a better solution would use `register_full_backward_hook` or similar to detect when we have completed the backward pass for the FSDP module, and reset the appropriate state. Also have a barrier in place to ensure any previous async reduction work is done before the next pre-backward call. This was discussed in the previous issue as well.

Diff for unit test
```
diff --git a/tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py b/tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
index b4127bc..6776e79 100644
--- a/tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
+++ b/tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
@@ -56,10 +56,10 @@ class Model2(nn.Module):
 
     def __init__(self):
         super().__init__()
-        self.block1 = nn.Sequential(nn.Conv2d(3, 4, kernel_size=3), nn.BatchNorm2d(4), nn.ReLU(inplace=True))
-        self.block2 = nn.Sequential(nn.Conv2d(4, 4, kernel_size=3), nn.BatchNorm2d(4), nn.ReLU(inplace=False))
-        self.block3 = nn.Sequential(nn.Conv2d(4, 8, kernel_size=3), nn.BatchNorm2d(8), nn.ReLU(inplace=True))
-        self.head = nn.Sequential(nn.AdaptiveAvgPool2d(output_size=(1, 1)), nn.Flatten(), nn.Linear(8, 10))
+        self.block1 = nn.Sequential(nn.Conv2d(3, 400, kernel_size=3), nn.BatchNorm2d(400), nn.ReLU(inplace=True))
+        self.block2 = nn.Sequential(nn.Conv2d(400, 400, kernel_size=3), nn.BatchNorm2d(400), nn.ReLU(inplace=False))
+        self.block3 = nn.Sequential(nn.Conv2d(400, 800, kernel_size=3), nn.BatchNorm2d(800), nn.ReLU(inplace=True))
+        self.head = nn.Sequential(nn.AdaptiveAvgPool2d(output_size=(1, 1)), nn.Flatten(), nn.Linear(800, 10))
 
     def forward(self, x):
         if isinstance(x, torch.Tensor):
@@ -91,9 +91,9 @@ def _create_model(
             if with_model2:
                 model.block3 = auto_wrap_bn(model.block3, single_rank_pg=False, fsdp_config=fsdp_config)
         if with_checkpoint:
-            model.block2 = checkpoint_wrapper(model.block2, maintain_forward_counter=True)
+            model.block2 = checkpoint_wrapper(model.block2, maintain_forward_counter=False)
             if with_model2:
-                model.block3 = checkpoint_wrapper(model.block3, maintain_forward_counter=True)
+                model.block3 = checkpoint_wrapper(model.block3, maintain_forward_counter=False)
         with enable_wrap(
             wrapper_cls=FSDP,
             flatten_parameters=flatten,
```

Alternatively, force synchronous reduction, and the unit test will fail with small weights.
```
diff --git a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
index f9a75c8..b77bc08 100644
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1360,6 +1360,8 @@ class FullyShardedDataParallel(nn.Module):
                 assert self._reducer is not None
                 grad_chunks = chunk_and_pad(param.grad.data, self.world_size)
                 self._reducer.reduce_scatter_async(grad_chunks, group=self.process_group, callback_fn=callback_fn)
+                self._reducer.flush()
+                self._streams["post_backward"].synchronize()
             else:
                 # Currently the only way for _is_sharded to be False is if
                 # world_size == 1. This could be relaxed in the future, in which
```

Example error:

`NCCL_P2P_DISABLE=1 python -m pytest -s tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py::test_multiple_forward_checkpoint[bn-model2-no_auto_wrap_bn-flatten-full]`
```
E       torch.multiprocessing.spawn.ProcessRaisedException: 
E       
E       -- Process 1 terminated with the following error:
E       Traceback (most recent call last):
E         File "/home/darryl/miniconda3/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 59, in _wrap
E           fn(i, *args)
E         File "/home/darryl/Projects/fairscale/tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py", line 200, in _distributed_worker
E           loss.backward()
E         File "/home/darryl/miniconda3/lib/python3.8/site-packages/torch/_tensor.py", line 255, in backward
E           torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
E         File "/home/darryl/miniconda3/lib/python3.8/site-packages/torch/autograd/__init__.py", line 147, in backward
E           Variable._execution_engine.run_backward(
E         File "/home/darryl/miniconda3/lib/python3.8/site-packages/torch/autograd/function.py", line 87, in apply
E           return self._forward_cls.backward(self, *args)  # type: ignore[attr-defined]
E         File "/home/darryl/Projects/fairscale/fairscale/nn/checkpoint/checkpoint_activations.py", line 323, in backward
E           torch.autograd.backward(outputs_with_grad, args_with_grad)
E         File "/home/darryl/miniconda3/lib/python3.8/site-packages/torch/autograd/__init__.py", line 147, in backward
E           Variable._execution_engine.run_backward(
E       RuntimeError: The size of tensor a (720600) must match the size of tensor b (1441200) at non-singleton dimension 0
```

dbarnhart:
Ensure gradient reduction accumulates into the unsharded gradient tensor
within a backwards pass. This matters when an FSDP module is called
multiple times within a forward pass, and reduction is _not_ deferred
using activation checkpoint forward counters, bucketing or some other
mechanism.

## What does this PR do?
Fixes #780.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -369,7 +369,7 @@ class FullyShardedDataParallel(nn.Module):
             f"FSDP.__init__(done): total_init_time: {(init_end - init_start): .4f} num_params: {(sum(p.numel() for p in self.params))}"
         )
 
-        # Flag to guard multiple pre-backward hook being executed per iteration.
+        # Flag to guard against preparing gradients multiple times per iteration.
         # This is reset at the end of the backward pass.
         self._pre_backward_hook_has_run = False
 
@@ -1206,16 +1206,18 @@ class FullyShardedDataParallel(nn.Module):
 
             # Only run the ``self._prep_grads_for_backward`` once per iteration (i.e. in case
             # it is multiple outputs or multiple forward passes).
-            if self._pre_backward_hook_has_run:
-                return
-            self._pre_backward_hook_has_run = True
+            if not self._pre_backward_hook_has_run:
+                self._pre_backward_hook_has_run = True
+                # Start of a backward pass for the first time in an iteration.
+                self.assert_state([TrainingState.IDLE, TrainingState.BACKWARD_PRE])
+                # Prepare p.grad so that it is in the right shape, device, accumulated values, etc.
+                self._prep_grads_for_backward()
 
-            # Start of a backward pass for the first time in an iteration.
-            self.assert_state([TrainingState.IDLE, TrainingState.BACKWARD_PRE])
-            self.training_state = TrainingState.BACKWARD_PRE
-
-            # Prepare p.grad so that it is in the right shape, device, accumulated values, etc.
-            self._prep_grads_for_backward()
+            # Transition to BACKWARD_PRE state if currently IDLE. We can transition from BACKWARD_POST
+            # to IDLE when FSDP is within activation checkpointing and called multiple times, due to the
+            # extra forward pass for re-computation.
+            if self.training_state == TrainingState.IDLE:
+                self.training_state = TrainingState.BACKWARD_PRE
 
         def _register_hook(t: torch.Tensor) -> torch.Tensor:
             if t.requires_grad:
@@ -1298,14 +1300,8 @@ class FullyShardedDataParallel(nn.Module):
         the local optimizer only sees the relevant parameter shard.
         """
         # First hook callback will see PRE state. If we have multiple params,
-        # then subsequent hook callbacks will see POST state. When checkpoint
-        # fwd counter is used, IDLE is also possible since the pre-backward hook
-        # is not triggered (see ``auto_wrap_bn`` below, we have to use
-        # FSDP(checkpoint(conv, FSDP(bn), ...)), with reshard_after_forward=False).
-        if hasattr(self, "_checkpoint_fwd_counter"):
-            self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST, TrainingState.IDLE])
-        else:
-            self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
+        # then subsequent hook callbacks will see POST state.
+        self.assert_state([TrainingState.BACKWARD_PRE, TrainingState.BACKWARD_POST])
         self.training_state = TrainingState.BACKWARD_POST
         if param.grad is None:
             return
@@ -1354,18 +1350,24 @@ class FullyShardedDataParallel(nn.Module):
                 # Average grad by world_size for consistency with PyTorch DDP.
                 param.grad.data.div_(self.gradient_predivide_factor)
 
-            callback_fn = functools.partial(self._post_reduction_hook, param)
             if param._is_sharded:
-                assert param._is_sharded
                 assert self._reducer is not None
-                grad_chunks = chunk_and_pad(param.grad.data, self.world_size)
+                # Save the unsharded grad for reduction. We will asynchronously accumulate the reduced gradient into
+                # param._saved_grad_shard. If this FSDP module was called multiple times it's possible that multiple
+                # gradient reductions will happen in an undefined order. But addition commutes, so this order doesn't
+                # matter, neglecting rounding.
+                grad = param.grad.data
+                # Clear grad on the tensor, so any repeated gradient computations do not interfere with this reduction.
+                param.grad = None
+                callback_fn = functools.partial(self._post_reduction_hook, param)
+                grad_chunks = chunk_and_pad(grad, self.world_size)
                 self._reducer.reduce_scatter_async(grad_chunks, group=self.process_group, callback_fn=callback_fn)
             else:
                 # Currently the only way for _is_sharded to be False is if
                 # world_size == 1. This could be relaxed in the future, in which
                 # case grads should be all-reduced here.
                 assert self.world_size == 1
-                callback_fn(param.grad.data)
+                self._post_reduction_hook(param, param.grad.data)
 
             # After _post_backward_hook returns, orig_grad_data will eventually
             # go out of scope, at which point it could otherwise be freed for
@@ -1377,33 +1379,36 @@ class FullyShardedDataParallel(nn.Module):
     def _post_reduction_hook(self, param: Parameter, reduced_grad: torch.Tensor) -> None:
         """Hook to call on each param after the reduce-scatter."""
         assert torch.cuda.current_stream() == self._streams["post_backward"]
-        assert param.grad is not None
         self.assert_state(TrainingState.BACKWARD_POST)
-        param.grad.data = reduced_grad
         if self.gradient_postdivide_factor > 1:
             # Average grad by world_size for consistency with PyTorch DDP.
-            param.grad.data.div_(self.gradient_postdivide_factor)
+            reduced_grad.data.div_(self.gradient_postdivide_factor)
         # Cast grad to param's dtype (typically FP32). Note: we do this
         # before the move_grads_to_cpu step so that this entire hook remains
         # non-blocking. The downside is a bit more D2H transfer in that case.
         if self.mixed_precision:
-            orig_param_grad_data = param.grad.data
-            param.grad.data = param.grad.data.to(dtype=param.data.dtype)
+            orig_param_grad_data = reduced_grad.data
+            reduced_grad.data = reduced_grad.data.to(dtype=param.data.dtype)
             # Don't let this memory get reused until after the transfer.
             orig_param_grad_data.record_stream(torch.cuda.current_stream())
-        if hasattr(param, "_saved_grad_shard") and param._saved_grad_shard is not None:
-            assert (
-                param._saved_grad_shard.shape == param.grad.shape
-            ), f"{param._saved_grad_shard.shape} vs {param.grad.shape}"
-            param.grad.data += param._saved_grad_shard
-            delattr(param, "_saved_grad_shard")
-        # Optionally move gradients to CPU, typically used if one is running
-        # the optimizer on the CPU.
+
+        if param._is_sharded:
+            # Accumulate into the gradient shard.
+            if getattr(param, "_saved_grad_shard", None) is None:
+                param._saved_grad_shard = reduced_grad.data
+            else:
+                assert (
+                    param._saved_grad_shard.shape == reduced_grad.shape
+                ), f"{param._saved_grad_shard.shape} vs {reduced_grad.shape}"
+                param._saved_grad_shard.data += reduced_grad.data
+            reduced_grad = param._saved_grad_shard.data
+
+        # Optionally move gradients to CPU, typically used if one is running the optimizer on the CPU. Once the full
+        # backwards pass completes, we will set `.grad` to the CPU copy.
         if self.move_grads_to_cpu:
-            param._cpu_grad.copy_(param.grad.data, non_blocking=True)
+            param._cpu_grad.copy_(reduced_grad.data, non_blocking=True)
             # Don't let this memory get reused until after the transfer.
-            param.grad.data.record_stream(torch.cuda.current_stream())
-            param.grad.data = param._cpu_grad
+            reduced_grad.data.record_stream(torch.cuda.current_stream())
 
     def _queue_wait_for_post_backward(self) -> None:
         """Try to queue a `wait_for_post_backward` callback.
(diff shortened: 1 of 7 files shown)


min-xu-ai:
Thank you so much @dbarnhart for your PR!

I guess you were not able to assign me to review so that I didn't get any notification on this PR. Feel free to @ tag me next time. I will take a look tomorrow.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1354,18 +1350,24 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
-                assert param._is_sharded
                 assert self._reducer is not None
-                grad_chunks = chunk_and_pad(param.grad.data, self.world_size)
+                # Save the unsharded grad for reduction. We will asynchronously accumulate the reduced gradient into
+                # param._saved_grad_shard. If this FSDP module was called multiple times it's possible that multiple
+                # gradient reductions will happen in an undefined order. But addition commutes, so this order doesn't
+                # matter, neglecting rounding.
+                grad = param.grad.data
+                # Clear grad on the tensor, so any repeated gradient computations do not interfere with this reduction.
+                param.grad = None
```
min-xu-ai, line 1361: This may increase the memory usage a bit due to 2 copies of the grad for the multiple forward pass + checkpointing case, right? Did you see the impact from that? Worth adding a comment about it here?

tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
```diff
@@ -310,39 +338,73 @@ def test_multiple_forward_checkpoint(precision, flatten, wrap_bn, model_type, bn
...
-                world_size,
-                with_model2,
-                with_sync_bn,
-                with_fsdp,
-                with_checkpoint,
-                mixed_precision,
-                flatten,
-                wrap_bn,
-                fp32_reduce_scatter,
+            for with_fwd_counters in [True, False]:
```
min-xu-ai, line 358: did this change make "fwd_counter" an optional thing now? If so, I think we should remove the fwd_counter from fsdp and checkpoint_wrapper since it only make things more complex. CC: @zhaojuanmao

min-xu-ai (commented):
This is truly awesome! I only have one main question below around the fwd_counter. I'd love to have at least 1 more FSDP expert take a look as well. I have added the likely folks to the reviewers.

tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
dbarnhart: Yes, it's optional now. There may be a use case for it with checkpointed recurrent blocks, but I'll let others speak to that.

You may be able to achieve the same efficiency in other ways, like using no-sync contexts, or toggling reshard after forward. But then the post backward hook should capture more of the FSDP internal state, so it matches the forward pass. We do the same thing with the AMP state in activation checkpointing, for example.

Still setting that state correctly might be a bit confusing, since the first call would require sync, but not the ones after, due to the inverted order of execution during backward. The counters are a bit more intuitive, perhaps.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
dbarnhart: The memory usage should be the same as we had before in the case of traditional gradient accumulation over the batch dimension. In that case, we would have saved the sharded gradient as well. Also if there is only one forward pass, or the reduction always happens fast enough, the memory usage should be unchanged. Only one unsharded gradient tensor would exist at a time. 

I'll take a closer look tomorrow, but I'm expecting a reduction in memory usage vs the forward counters. Keeping one more sharded tensor should be much better than keeping any unsharded tensors, especially with say 8 gpus.

tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
min-xu-ai: interesting. what's the difference between recurrent blocks and block that have multiple fwd passes? I haven't worked on RNN's yet. Thanks again for this awesome work!

dbarnhart: Just the module call orders. RNNs would be like [A, A, A, ...], whereas my case is interleaved [A, B, C, A, B, C]. If the calls are consecutive, there's no point in resharding the full parameters. It wouldn't change the high water mark amount of vram used.

min-xu-ai: Thanks. Do you mind just remove the fwd counter then? It would simplify the code and you don't need to test so many cases in the test as well.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
dbarnhart: The memory usage is definitely reduced with these changes. I'm able to run larger models locally than I could before, without OOM.

I elaborated the comment to discuss memory usage, including a description of how we can mitigate the 2 copy edge case by using CUDA events to limit the number of outstanding reduction operations.

tests/nn/data_parallel/test_fsdp_multiple_forward_checkpoint.py
dbarnhart: Done.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1206,16 +1201,18 @@ def _pre_backward_hook(*unused: Any) -> None:
...
+                self._pre_backward_hook_has_run = True
+                # Start of a backward pass for the first time in an iteration.
+                self.assert_state([TrainingState.IDLE, TrainingState.BACKWARD_PRE])
+                # Prepare p.grad so that it is in the right shape, device, accumulated values, etc.
+                self._prep_grads_for_backward()
+
+            # Transition to BACKWARD_PRE state if currently IDLE. We can transition from BACKWARD_POST
+            # to IDLE when FSDP is within activation checkpointing and called multiple times, due to the
+            # extra forward pass for re-computation.
+            if self.training_state == TrainingState.IDLE:
```
min-xu-ai, line 1214: Useful to add an assert_state before this if statement?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1354,18 +1338,34 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
                 assert self._reducer is not None
-                grad_chunks = chunk_and_pad(param.grad.data, self.world_size)
+                # Save the unsharded grad for reduction. We will asynchronously accumulate the reduced gradient into
+                # param._saved_grad_shard. If this FSDP module was called multiple times it's possible that multiple
+                # gradient reductions will happen in an undefined order. But addition commutes, so this order doesn't
+                # matter, neglecting rounding.
+                grad = param.grad.data
+                # Clear grad on the tensor, so any repeated gradient computations do not interfere with this reduction.
+                #
+                # The effect on memory consumption is not usually significant. No extra memory is allocated if this
```
min-xu-ai, line 1350: nice!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1377,33 +1377,36 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
+                param._saved_grad_shard = reduced_grad.data
+            else:
+                assert (
+                    param._saved_grad_shard.shape == reduced_grad.shape
+                ), f"{param._saved_grad_shard.shape} vs {reduced_grad.shape}"
+                param._saved_grad_shard.data += reduced_grad.data
+            reduced_grad = param._saved_grad_shard.data
+
+        # Optionally move gradients to CPU, typically used if one is running the optimizer on the CPU. Once the full
+        # backwards pass completes, we will set `.grad` to the CPU copy.
```
min-xu-ai, line 1405: thanks for the comment!

min-xu-ai (approved):
Wow, so clean! Only one minor thing.

I am glad to hear that you can run bigger models! Do you have any urgency in terms of merging this in this week? if not, perhaps early next week we merge it after others have a chance to take a look?

@anj-s @zhaojuanmao @myleott @sshleifer any comments from you on this?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
dbarnhart: It's going to be a pretty liberal assertion. The only state that's disallowed is TrainingState.FORWARD. Still, it's probably wise to check this here, since it's obviously an error to be in FORWARD state at this point.

dbarnhart:
I can run my work off of the branch for now, using a custom build of fairscale, so waiting is not a problem. Happy to have more eyes on this change. It involves reasoning about order-of-events under concurrency / async / complexity, which is always a bit nuanced.

Thanks for all the positive feedback!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
dbarnhart: Done.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1354,18 +1339,34 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
                 param.grad.data.div_(self.gradient_predivide_factor)
 
-            callback_fn = functools.partial(self._post_reduction_hook, param)
             if param._is_sharded:
-                assert param._is_sharded
                 assert self._reducer is not None
-                grad_chunks = chunk_and_pad(param.grad.data, self.world_size)
+                # Save the unsharded grad for reduction. We will asynchronously accumulate the reduced gradient into
+                # param._saved_grad_shard. If this FSDP module was called multiple times it's possible that multiple
+                # gradient reductions will happen in an undefined order. But addition commutes, so this order doesn't
```
zhaojuanmao, line 1346: since multiple gradient reductions are queued into the same stream, iiuc, the order should be deterministic?

zhaojuanmao (commented):
very nice, lgtm!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
dbarnhart: I initially made weak assumptions about the reducer. But the current implementation preserves order, as you say, even with bucketing. This could be rephrased. The main observation is that this logic does not require deterministic ordering.

PR #779, opened 2021-09-07
[test] Added disable_checkpointing unit test

JARACH-209:
## What does this PR do?
Fixes #778 .
https://github.com/facebookresearch/fairscale/issues/778#issue-988653326

## Before submitting

- [x] Did you have fun?
  - Make sure you had fun coding 🙃
- [x] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?
- [x] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
  - [ ] N/A
- [ ] Did you make sure to update the docs?
  - [ ] N/A
- [x] Did you write any new necessary tests?
  - [ ] N/A
- [ ] Did you update the [changelog](https://github.com/facebookresearch/fairscale/blob/master/CHANGELOG.md)? (if needed)
  - [ ] N/A

## PR review
Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

PR #787, opened 2021-09-10
[feat] set requires_grad of output tensors of checkpointed modules properly

xiaoxiao26:
Before this commit, output tensors of checkpointed modules always
require grad, even if they shouldn't. This commit makes it so that
the outputs of checkpointed modules only require grad if either
the input requires grad or if the parameters require grad.
To achieve this, this commit also adds a new _unflattened_param_views
attribute to modules being flattened. This allows the checkpointing
to still access the parameters and check if gradients need to be
computed.

## What does this PR do?

This enables users to use the checkpoint wrapper on modules even if no outputs require gradients. Previously, the blocker was that the backward pass will still trigger even if the module does not need gradients, causing an assertion to fail. This is because currently the checkpoint wrapper makes the output always require gradients, so autograd will always call backward, even if the outputs don't actually need gradients.

To fix this, this PR adds a feature to compute if outputs need gradients in the forward pass by checking if the module's parameters need gradients OR if the inputs need gradients. If the output does not need gradient, then it is detached to not need gradients. 

To achieve this, this PR also adds a new feature in FlattenParamsWrapper to keep track of the param views into the flattened params. This is so the checkpointing can still check if the module's parameters need gradients even when deleted by FlattenParamsWrapper.

*Testing*

Added a new unit test to test that backward pass is not triggered for modules not needing gradients (by counting number of times forward pass ran). 

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/nn/checkpoint/checkpoint_activations.py
+++ b/fairscale/nn/checkpoint/checkpoint_activations.py
@@ -191,13 +191,33 @@ def _checkpointed_forward(
     # when original_forward's input are non-tensor (i.e. a tuple). Using this dummy tensor
     # avoids requiring users to set their input tensors's requires_grad flag. In the case
     # of tuple type inputs, setting the flag won't even trigger the backward pass.
+    #
+    # One implication of this is that since we always feed in a dummy tensor
+    # needing grad, then the output will always require grad, even if it originally
+    # wouldn't, such as if the module and original input both do not require grad.
+    # We get around this by saving the desired requires_grad value in output and
+    # detaching the output if needed.
     output = CheckpointFunction.apply(
         torch.tensor([], requires_grad=True), original_forward, parent_ctx_dict, kwarg_keys, *flat_args
     )
+    output_requires_grad = parent_ctx_dict["output_requires_grad"]
     if not isinstance(output, torch.Tensor):
+        # If output should not require grad, then detach it, since otherwise it will
+        # always have requires_grad = True due to our dummy tensor input above that
+        # requires_grad
+        output = [x.detach() if not output_requires_grad else x for x in output]
+
         packed_non_tensor_outputs = parent_ctx_dict["packed_non_tensor_outputs"]
         if packed_non_tensor_outputs:
             output = unpack_non_tensors(output, packed_non_tensor_outputs)
+
+    else:
+        # If output should not require grad, then detach it, since otherwise it will
+        # always have requires_grad = True due to our dummy tensor input above that
+        # requires_grad
+        if not output_requires_grad:
+            output = output.detach()
+
     return output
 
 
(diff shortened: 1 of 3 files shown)


min-xu-ai:
Thanks @xiaoxiao26 ! This looks great to me.

Just one question, in this case, does the unnecessary checkpoint actually causes forward pass to run twice? If so, perhaps this error is still good so that users do end up being slowed down by this error?

If not, is it possible to add a simple unit test like it is in this PR? https://github.com/facebookresearch/fairscale/pull/779

xiaoxiao26:
Thanks for the quick response @min-xu-ai. I believe the forward pass does run twice (line 305 in checkpoint_activations.py). However, the current error throwing makes it impossible to use checkpointing in certain cases:

Suppose we have a checkpointed module that does not require grad. Then its output will only require grad if its input requires grad. Theoretically, we could feed it inputs requiring grad half the time, and inputs not requiring grad the other half. This makes it impossible to use the checkpoint wrapper as it is, unless the user tries to keep a reference to both the checkpointed module and the original module, detect if the input requires grad, and feed it to the proper module, which is messy.

I wonder if we can just detect if the output requires grad in the forward function, save that result to ctx, and then in backward just not run the forward pass if output does not require grad. The backward function would then return None for each input.

min-xu-ai:
> I wonder if we can just detect if the output requires grad in the forward function, save that result to ctx, and then in backward just not run the forward pass if output does not require grad. The backward function would then return None for each input. Let me know if this sounds good and I can try implementing this.

This sounds like a good idea to me. @anj-s @tmarkstrum do you have time to work on this? Thank you!

xiaoxiao26:
I put up a proposed implementation to achieve proper behavior when requires_grad is False:

1. Since we run forward with torch.no_grad, we cannot actually retrieve the output's requires_grad parameter. As a result, I compute it manually by checking if the input requires grad or if the module's parameters requires grad

2. This method assumes that the input's requires_grad field is accurate. Currently checkpointing always sets the output requires_grad to True, which causes future inputs to have requires_grad = True. The reason it is always set to true is because in the following code snippet:

output = CheckpointFunction.apply(
    torch.tensor([], requires_grad=True), original_forward, parent_ctx_dict, kwarg_keys, *flat_args
)

Since torch.tensor([], requires_grad=True) is a dumy input with requires_grad = True, Autograd will think any outputs also have requires_grad=True, even when this is not the case. To fix this, I added a mechanism to detach the output if requires_grad should actually be False

3. I added some tests to verify that these 2 fixes do indeed prevent checkpointing from running forward twice when the outputs of a checkpointed module do not require grad.

Please let me know if this approach is sound!

min-xu-ai:
Thanks a lot for this!

The dummy tensor was used to fix an issue when input is a list or dict. See this: https://github.com/facebookresearch/fairscale/pull/701

I like your approach. @anj-s should also take a look since she had experiences on checkpointing and dummy process.

min-xu-ai (approved):
Looks really good! I would wait at least one more reviewer before merging this.

xiaoxiao26:
Thanks Min! Looks like some GPU tests failed though, taking a look....

zhengwy888:
pretty sweet test cases!

xiaoxiao26:
I found the issue: FlattenParamsWrapper deletes the parameters, so I couldn't check if the parameters required gradients or not because they were gone! I got around this by having FlattenParamsWrapper save the parameter views into the module itself. This now passes a local run I had, hopefully circleci is fixed too.

I've also restructured this PR to be "set requires_grad of output tensors of checkpointed modules properly" rather than "allow checkpoint wrapper even if no outputs require gradients", because as long as we set the requires_grad properly, Autograd will not trigger backward pass on the modules that don't have outputs with gradients, and everything will work out well.

anj-s (approved):

anj-s:
@xiaoxiao26  Thank you for the PR! I think the implemented solution should work.

PR #798, opened 2021-09-16
checkpoint + nonflat + fp16

min-xu-ai:
## What does this PR do?
Fixes # (issue).

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


min-xu-ai:
Hi Yanli,

While debugging an issue with @arbabu123, I found a curious case that FSDP(checkpoint()) + mixed_precision + non_flatten doesn't work with FSDP since the post backward hook fails to fire.

I extended this simple unit test to cover 32 different cases and 4 of them fails as the following.

image

Do you know why doesn't the post backward hook in this case? I am very surprised that flatten actually helps in this case. So because we move to flat tensor, we may want to make sure this case works with flat tensors.

Also C->F (i.e. checkpoint(FSDP())) works in this case, which is also good. But that config is likely slower than F->C.

CCing @myleott @anj-s in case they are interested.

tests/nn/data_parallel/test_fsdp_with_checkpoint_wrapper.py
```diff
@@ -102,10 +130,17 @@ def _check_params(model, expected_param_shapes):
 
 
 class SimpleModuleWithCheckpointing(nn.Module):
-    def __init__(self):
+    def __init__(self, flatten, mixed_precision, fsdp_wrap_ckpt):
         super().__init__()
+        if fsdp_wrap_ckpt:
+            middle_module = FSDP(checkpoint_wrapper(nn.Linear(3, 3)), flatten_parameters=flatten, mixed_precision=mixed_precision)
```
min-xu-ai, line 136: this is the config that fails when flatten = False and mixed_precision=True.

zhaojuanmao:
did some debugging last Friday, confirmed the backward hooks are registered with parameters in F(C(module)), also confirmed the grads of these parameters are computed. but the hooks are not fired when fp16 is enabled.  

I am suspecting the grad fns are mutated somehow when F(C()) with f16, but I can not confirm with it, as looks like p.grad_fn is always None when backward starts, I can not print the grad_fn object in backward.  

Also, I feel this weird behavior is related to p.data usage, which is not safe and may impact auto grad calculation.  ==> but it seems not easy to remove p.data in current repo.  I tried to p.set_(p.fp16_shard), it will throw error 'data type does not match'

PR #836, opened 2021-10-28
[feat] [FSDP]: add experimental support to shared weights

min-xu-ai:
## What does this PR do?

This adds experimental and limited support of shared weights between FSDP wrapper.

### Why is this needed?

When a weight is shared between layers, before this change, it needs to be part of a single FSDP wrapper. This means that while the weight is not being used, it might still be stored in GPU memory in full due to this limitation.

When we can separately wrap layers with shared weights, GPU memory can be saved. E.g.

```
L0 -> L1 -> L2 -> L3
L3.weight = L0.weight
```

Without this change, L0 and L3's weight are in the outer (root) FSDP and can't be sharded when L1 and L2 are computing.

With this change, L0 and L3 can be separately wrapped. Thus, allow the weight and gradient to be sharded as soon as it is done computing.

A special type of L3 kernel will be needed due to autograd execution order. An example of that will be in a separate PR.

### Why is the limiting and experimental

The FSDP wrappers for the weight sharing layers need to be un-flatten and special care need to be taken with the layers. 

Supporting general sharing case is likely very difficult and not necessarily needed. This is to enable particular kind of big scale training for research purposes.

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -414,6 +414,30 @@ class FullyShardedDataParallel(nn.Module):
         assert isinstance(self._fsdp_wrapped_module, FlattenParamsWrapper)
         return self._fsdp_wrapped_module
 
+    def append_shared_param(self, p: Parameter) -> None:
+        """ Add a param that's already owned by another FSDP wrapper.
+
+            .. warning:: This is experimental!
+
+            This only works with all sharing FSDP modules are un-flattened.
+
+            p must to be already sharded by the owning module.
+
+            Check the corresponding unit test to see how is it used and tested.
+            In particular, the sharing FSDP wrappers are "siblings" not "parent"
+            and "child" of each other in the nested module structure.
+
+        Args:
+            p (Parameter):
+                The shared parameter.
+        """
+        assert self._is_root is None
+        assert not self.flatten_parameters
+        assert isinstance(p, Parameter)
+        assert p._is_sharded
+        p._is_shared = True
+        self.params.append(p)
+
     def apply(self, fn: Callable[[nn.Module], None]) -> "FullyShardedDataParallel":
         """
         Applies ``fn`` recursively to every submodule (as returned by
@@ -916,8 +940,16 @@ class FullyShardedDataParallel(nn.Module):
                     yield
                 finally:
                     stack.close()
-                    assert len(full_tensors) == len(self.params)
-                    for p, (full_tensor, safe_to_free) in zip(self.params, full_tensors):
+                    non_shared_params = self.params
+                    # filter out shared params for all but the owner FSDP module.
+                    if len(full_tensors) < len(non_shared_params):
+                        non_shared_params = list(
+                            filter(lambda p: not (hasattr(p, "_is_shared") and p._is_shared), self.params)
+                        )
+                    assert len(full_tensors) == len(
+                        non_shared_params
+                    ), f"{len(full_tensors)} vs. {len(non_shared_params)}"
+                    for p, (full_tensor, safe_to_free) in zip(non_shared_params, full_tensors):
                         if not volatile:
                             # Copy any changes made to the full params back into
                             # the corresponding local shards.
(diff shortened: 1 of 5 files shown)


zhaojuanmao:
> @zhaojuanmao
> 
> Now this test's results make more sense.
> 
> * If nothing is shared, of course, FSDP and nonFSDP match.
> * If something is shared:
>   
>   1. shape mismatch error (non flat cases)
>   2. no shape mismatch but numeric mismatch (as if nothing is shared under FSDP) (flat cases)

I see, that makes sense to me as well!

min-xu-ai:
@zhaojuanmao. I added some limited and experimental support for shared param. If you want to, please take a look!

sshleifer (commented):
LGTM. Can we revert/keep the code simpler if this doesn't end up being useful?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -414,6 +414,33 @@ def module(self) -> FlattenParamsWrapper:
...
+                The shared parameter.
+        """
+        assert self._is_root is None
+        assert not self.flatten_parameters
+        assert isinstance(p, Parameter)
+        assert p._is_sharded
+        p._is_shared = True
+        assert (
+            len(list(filter(lambda p: not (hasattr(p, "_is_shared") and p._is_shared), self.params))) > 0
+        ), "Must have at least 1 non-shared param."
```
anj-s, line 441: what does this requirement satisfy?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1367,6 +1402,17 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
         if param.grad is None:
             return
 
+        if hasattr(param, "_linked_param"):
```
anj-s, line 1405: where do we set this property?

tests/nn/data_parallel/test_fsdp_shared_weights.py
```diff
@@ -0,0 +1,134 @@
...
+        for _ in range(steps_per_iter):
+            out = model(in_data)
+            out.sum().backward()
+        # Simulate gradient means between ranks.
+        if steps_per_iter > 1:
+            with torch.no_grad():
+                for p in model.parameters():
+                    p.grad /= steps_per_iter
+        optim.step()
+        model.zero_grad(set_to_none=True)
```
anj-s, line 134: nit: this is usually recommended to be at the start of the loop. gives the best performance and avoids errors.

tests/nn/data_parallel/test_fsdp_shared_weights.py
```diff
@@ -0,0 +1,134 @@
...
+            assert sharing is None or sharing == "share_none"
+
+        if with_fsdp:
+            # Shared layers much be un-flatten.
+            self.l1 = FSDP(self.l1, flatten_parameters=False)
+            self.l2 = FSDP(self.l2, flatten_parameters=inner_flat)
+            self.l3 = FSDP(self.l3, flatten_parameters=False)
+
+            if sharing in ["share_only_weights"]:
+                self.l3.append_shared_param(self.l1.module.weight)
```
anj-s, line 49: Is this a change users will have to make in their models to use this feature?

min-xu-ai:
> LGTM. Can we revert/keep the code simpler if this doesn't end up being useful?

Thanks for the review. Yes definitely.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: This is due to autograd’s hook firing order

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: This is set in the kernel, which is the next PR

tests/nn/data_parallel/test_fsdp_shared_weights.py
min-xu-ai: I think at the end is in fact slightly better since it reduces the memory immediately after they are used. Where is the source on beginning is better?

tests/nn/data_parallel/test_fsdp_shared_weights.py
min-xu-ai: Yes

min-xu-ai:
Merged this so that my next PR can properly be based on this. Always OK to address more comments of course.

PR #839, opened 2021-11-02
[feature]Add support for SSD offload with FSDP for eval workloads

anj-s:
## What does this PR do?
Adds experimental support for offloading parameters to disk for FSDP API. Note: this only works for eval right now and support for training is being added in an upcoming PR.

- Added a state Enum to represent when the SSD buffer is on disk.
- Every FSDP module owns a SSD buffer which points to a list of flattened params owned by the module.
- We read params from disk to CPU memory and then use the existing mixed precision mechanism for copying onto the device.  Note: this will be deprecated soon once we add support for SsdParameter which should do this automatically. The code should be a lot cleaner at that point.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -323,6 +339,16 @@ def __init__(
 
         self._has_params = len(params) > 0
 
+        # TODO(anj): Should we conditionally do this only if we have params?
+        # TODO(anj): this won't work for MOE params.
+        self.buffer_size = sum(p.numel() for p in params)
+        self.ssd_buffer_filename = ""
+        if self.ssd_offload:
+            self.ssd_buffer_filename = f"{randint(1, int(10E6))}_rank{self.rank}"
```
min-xu-ai, line 347: is the file name always xxxx_rankN, where xxxx is a rand number? That can still collide and it looks bad. Why not use the tempfile module to get a temp file name with proper prefix that is more readable?

min-xu-ai: Also, which dir will this file be in? The offload dir seems need to be an argument that's passed in as well?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -323,6 +339,16 @@ def __init__(
...
+            self.ssd_buffer = ssd_offload.SsdBuffer(self.buffer_size, self.ssd_buffer_filename)
+            self.move_grads_to_cpu = True
+            self.move_params_to_cpu = True
```
min-xu-ai, line 350: perhaps asserting these are set correctly by the user instead of changing them on the fly? this way users won't be surprised when they find out these are changed.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -616,18 +650,36 @@ def _shard_parameters_(self) -> None:
...
+                p.data, num_padded = self._get_shard(p.data)
+                self.numel_padded_per_param.append(num_padded)
+                free_storage_(orig_data)
 
-            p._is_sharded = True
         assert len(self.numel_padded_per_param) == len(self.params)
 
+        # Move SSD buffer to disk.
+        if self.ssd_offload:
+            self.ssd_buffer.to_disk()
```
min-xu-ai, line 681: Can you please restructure this part to:
```
< original code>

if ssd_offload:
   <ssd offload code>
```
This way, the impact of offload code is easier to see and to separate for debugging etc.

min-xu-ai: if reserving the order is easier, that's fine too, like:

```
if ssd:
   <ssd code>
   return

<original code>
```

A bit of duplicated code is fine too but I see the code above is mostly separately anyway. so there shouldn't be any duplicates.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -704,6 +756,18 @@ def fixup(p: Parameter, is_sharded: bool, size: torch.Size) -> Parameter:
...
+        """Returns an iterator over the module parameters, yielding all the parameters
+        part of the model.
+        """
+        # TODO(anj): Use `copy_into_tensor` in order to provide a copy of the
+        # parameters and not the actual parameters. Ideally we don't users to operate on
+        # actual params.
+        if self.ssd_offload and self.ssd_buffer.storage_state == StorageState.ON_DISK:
+            self.ssd_buffer.from_disk(self.buffer_size)
+
+        return super().parameters(recurse=recurse)
```
min-xu-ai, line 769: can this just call named_parameters and just remove the names from the results using an transformation function?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -810,11 +880,23 @@ def local_state_dict(self, *args: Any, **kwargs: Any) -> Any:
...
+        if self.ssd_offload:
+            # Move params from disk to memory before returning the local state dict.
+            self._move_params_to_memory()
+
         try:
             yield
         finally:
             self._return_full_state_dict = backup
 
+    def _move_params_to_memory(self) -> None:
```
min-xu-ai, line 893: doctoring?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -810,11 +880,23 @@ def local_state_dict(self, *args: Any, **kwargs: Any) -> Any:
...
+        if self.ssd_buffer.storage_state == StorageState.ON_DISK:
```
min-xu-ai, line 894: this seems like an abstraction leak? Shouldn't ssd_buffer maintain the state internally and decide what to do with using it? Caller using the state and decide what to do is error prune since caller may not know all the contracts.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1228,8 +1328,16 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
             torch.clear_autocast_cache()
 
+        self._free_ssd_offload()
+
         return outputs
 
+    @torch.no_grad()
+    def _free_ssd_offload(self) -> None:
+        if self.ssd_offload:
+            if self.ssd_buffer.storage_state == StorageState.IN_MEMORY:
```
min-xu-ai, line 1338: again, this seems like an abstraction leak.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1044,7 +1133,10 @@ def _init_param_attributes(self, p: Parameter) -> None:
             # If we plan to keep the FP32 parameters on CPU, then pinning
             # memory allows us to later use non-blocking transfers when moving
             # the FP32 param shard to compute_device.
-            p._fp32_shard = p._fp32_shard.pin_memory()
+            if not self.ssd_offload:
+                # We don't pin memory when using ssd_offload since that results in OOM when
```
min-xu-ai, line 1137: I don't see a change about not using non-blocking transfers when ssd offload is on. The comment on 1134 mentioned that pin_memory is needed for non-blocking transfer. So I suspect that something is not be tested properly here.

min-xu-ai (commented):
Took a quick look. Nice initial change. I have suggestions around keeping the logic more separated and also I have a concern on an abstraction leak.

Since @tmarkstrum  and @anupam-fb may work on FSDP later, I think they can take a look too. CC'ing @blefaudeux since he always have sharp eyes for great code reviews.

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -264,7 +285,7 @@ def get_tensors(self) -> List[SsdTensorHandle]:
 
     def to_disk(self) -> None:
         """Writes all tensors backed by handles to disk."""
-        assert list(self.buffer.size()) != [1]
+        assert self.storage_state == StorageState.IN_MEMORY
         # TODO(anj): Add comment about why we use `narrow`.
```
another-pjohnson, line 289: We use narrow so the write is only for the valid tensors. If the SSDBuffer is allocated with a large 10MB buffer, but we only allocate 1MB in tensors, we don't want to write all 10MB to disk, only the 1MB that is valid.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: @anj-s, it should be enough to modify SsdBuffer.from_disk to add a check at the top:

Something like:
```
if num_elems < self.offset:
            raise RuntimeError(
                f"Attempted to load from file ssdbuffer of size: {self.offset} into a buffer that is of size: {num_elems}"
            )
if self.storage_state = StorageState.IN_MEMORY:
    return
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1669,6 +1778,18 @@ def update_p_data(custom_output_tensor: Optional[torch.Tensor] = None) -> None:
             # Trim any padding and reshape to match original size.
             p.data = p.data[: p._orig_size.numel()].view(p._orig_size)
 
+        if self.ssd_offload:
+            if self.ssd_buffer.storage_state == StorageState.ON_DISK:
```
another-pjohnson, line 1782: Same comment as @min-xu-ai  mentioned about abstraction leak

another-pjohnson (commented):
changes look good to me, take a look at the comments I've made when you get a chance.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -272,6 +280,7 @@ def __init__(
         force_input_to_fp32: bool = False,
         verbose: bool = False,
         cpu_offload: bool = False,
+        **kwargs: Dict[str, Any],
```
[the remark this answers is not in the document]
anj-s: I think we already have a lot of input params that are confusing to say the least. I was trying to see if I could add the params needed for offload in more structured way. I need to specify the feature as well as the file location. There may be other params like buffer size etc. and adding to this list may not be the best way.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -323,6 +339,16 @@ def __init__(
 
         self._has_params = len(params) > 0
 
+        # TODO(anj): Should we conditionally do this only if we have params?
+        # TODO(anj): this won't work for MOE params.
+        self.buffer_size = sum(p.numel() for p in params)
```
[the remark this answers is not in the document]
anj-s: No, its number of elements.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -323,6 +339,16 @@ def __init__(
...
+        self.ssd_buffer_filename = ""
+        if self.ssd_offload:
```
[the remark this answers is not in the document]
anj-s: Makes sense, added an assert for this check.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: I haven't decided the way forward here hence I can add a TODO to address this in a follow up PR.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: That makes sense but what would they set this to with ssd_offload. The default is False and the docs should mention that ssd_offload will imply movement of grads and params to CPU.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: You don't need to set the non_blocking argument if pin_memory is not used. AFAIK the non_blocking will default to the synchronization point and not yield a perf gain. I added comments for ssd_offload.
I've tested this pretty extensively on large models and they haven't failed. In fact i was only able to test it once I fixed this memory leak in pin_memory.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: restructured it within the is_sharded vs not blocks. Let me know if this is what you had in mind as well.

anj-s:
> Took a quick look. Nice initial change. I have suggestions around keeping the logic more separated and also I have a concern on an abstraction leak.
> 
> Since @tmarkstrum and @anupam-fb may work on FSDP later, I think they can take a look too. CC'ing @blefaudeux since he always have sharp eyes for great code reviews.

Thanks! Yep, agree on the abstraction leak and it is an easy fix :)
The main thing that a follow up PR will address is the API and how we will specify the ssd files.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: The base class does the same thing so would prefer deferring to it. let me know if there is a reason why we should not.

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -68,6 +69,20 @@ def read(input_tensor: torch.Tensor, filename: str, file_offset_bytes: int = 0)
...
+
+    UNALLOCATED = auto()
+    ON_DISK = auto()
+    ON_CPU = auto()
+    ON_GPU = auto()
+
+
 class SsdTensorHandle(torch.Tensor):
     """
     This class extends from torch.Tensor and represents a Tensor which is backed by SSD storage.
```
blefaudeux, line 88: super late to the party, but could that relate to the "meta" tensor discussions (virtual device) which are being developed in current pytorch ? worth being in touch if it's not already the case ? I see that as a specialized meta tensor basically

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -104,6 +119,7 @@ def __init__(self, shape: Tuple[int, ...], dtype: torch.dtype, requires_grad: bo
         # valid if loaded to memory
         self.tensor: Optional[torch.Tensor] = None
         self.requires_grad = requires_grad
+        self.storage_state = StorageState.UNALLOCATED
 
     @classmethod
     def from_file(
```
blefaudeux, line 125: late again, but would it be valuable to pass in an optional cuda Stream here, so that you can make sure that some of these transfers are done in parallel with some other compute ? I do realize that the rest of the class is not cuda specific though

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -68,6 +69,20 @@ def read(input_tensor: torch.Tensor, filename: str, file_offset_bytes: int = 0)
...
+    Simple enum to indicate whether the tensor handle is pointing
+    to data on disk or memory. This is useful for asserting on 
+    whether the tensor is available for operations or if it needs
+    to be moved from disk to CPU or device.
+    """
+
+    UNALLOCATED = auto()
+    ON_DISK = auto()
+    ON_CPU = auto()
+    ON_GPU = auto()
```
blefaudeux, line 83: (related to my CUDA stream remark): I don't see this being used actually, is that planned in a follow up or is the expectation that it will be done somewhere else on the stack (by the user who consumes this) ?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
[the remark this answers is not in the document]
blefaudeux: could be with an optional config structure for instance (the C way) ? 
- Optional would convey the fact that you can skip that
- A structure would convey the fact that all these settings are related
- With a structure you could define defaults somewhere else
- You could extend this structure down the line without changing the API here

I agree with Min, API inputs are 80% of the understanding for people who did not write the code (completely made up statistic), it's a little hard to discover this way

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -704,6 +759,18 @@ def fixup(p: Parameter, is_sharded: bool, size: torch.Size) -> Parameter:
...
 
+    def parameters(self, recurse: bool = True) -> Iterator[Parameter]:
+        """Returns an iterator over the module parameters, yielding all the parameters
+        part of the model.
+        """
+        # TODO(anj): Use `copy_into_tensor` in order to provide a copy of the
+        # parameters and not the actual parameters. Ideally we don't users to operate on
+        # actual params.
+        if self.ssd_offload:
+            self.ssd_buffer.from_disk(self.buffer_size)
```
blefaudeux, line 770: lacking some context here (these collapsed views are killing me), but to make sure I understand, ssd_buffer is a single flat tensor which holds all of this module's parameters, right ? Noob question, but is there any value in having a mixed strategy ? (guess is no, but I'm wondering about parameters which would be shared across modules for instance)

fairscale/experimental/nn/ssd_offload.py
anj-s: We use the same streams as that used by FSDP so in a sense we are not modifying the CPU offload mechanism that we have in place. This class is only responsible for moving params from disk to CPU memory not the GPU memory. One thing to consider is that we can't use pin_memory when using ssd offload. This is because we would essentially need to account for the entire model memory which leads to OOM errors.

fairscale/experimental/nn/ssd_offload.py
anj-s: good point! We don't use it right now and we probably won't since this is the only class that should be handing this state. Initially I was contemplating sharing this structure with the FSDP codebase. Will remove it for now.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: I don't think there is any value right now. Maybe it is something to consider in the future. Prefer to keep it simple in the beginning :)

fairscale/experimental/nn/ssd_offload.py
anj-s: We've implemented this in collaboration/discussion with the PyTorch team. This was the recommended way for us to implement a new Tensor type where we could essentially hook into the dispatch process. We prototyped a meta tensor implementation as part of adding support that had its own issues.
The meta tensor you mention is also related for sure but for our purposes is required more from an initialization POV. We will be using the new [`init_meta`](https://github.com/pytorch/pytorch/pull/66317) API for our delayed initialization needs.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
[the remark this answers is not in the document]
anj-s: Maybe, I am still writing down options and a structure is what I have in mind as well. API inputs are great to see but I don't think having a laundry list of flags that users have to read about helps. Definitely unintentional consequence but we have that right now.

Don't disagree with either @min-xu-ai  or you. Maybe I should have explained that this is a short term change before the API PR. I know API discussions can be long and hence did not want to block development on the feature. There are still a few changes in flight before we can call this "ready".

anj-s:
hey folks, please let me know if you have any other comments. I need to merge this PR to avoid blocking more changes. thanks!

PR #851, opened 2021-11-12
[fix] [MEVO]: make mevo work with eval and optim_state checkpointing

min-xu-ai:
## What does this PR do?

While doing integration with fairseq, these are needed fixes for eval and checkpointing.

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1790,6 +1797,18 @@ def update_p_data(custom_output_tensor: Optional[torch.Tensor] = None) -> None:
 
             self.has_full_params = False
 
+        if self._has_shared_params:
+            # self.has_full_params flag can be out of sync if a shared param is
+            # sharded by another FSDP instance. An example is that in eval case
+            # with reshard_after_forward=False but the sharing instance has
+            # reshard_after_forward=True. Then, on the second forward, the
```
anj-s, line 1804: Can you clarify what happens in the second forward?Does the reshard happen by the first instance and therefore the second instance is unaware of this?

Also, do we need a test for this? i noticed that we don't have unit tests for any of the changes in the PR.

anj-s (approved):

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: Thanks for the review!

On the second forward, one FSDP instance performs reshard on the shared weight but the other one is not aware and other one is thinking that the weight should still be there since it has reshard turned off. Therefore, there is a problem.

Good catch on the tests, I missed them. Let me add them.

PR #853, opened 2021-11-15
Update changelog, removed meta.yml and requirements cleanup

anupambhatnagar:
## What does this PR do?
- Updated the Changelog.md.
- Updated circle ci config file.
- Updated pre-commit config file.
- Updated Contributing.md.
- Removed meta.yml. This file has not been in use for a long time.
- Merged `requirements-dev.txt` and `requirements-test.txt`. Deleted `requirements-test.txt`
- Updated `requirements-benchmarks.txt`.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [x] N/A
  - [ ] N/A

## PR review


anj-s:
Will you need to update circleCI config as well since we use [requirements-text.txt](https://github.com/facebookresearch/fairscale/blob/main/.circleci/config.yml#L90).

CONTRIBUTING.md
```diff
@@ -67,19 +71,13 @@ pre-commit install
 After the above, your `git commit` command will automatically trigger pre-commit
 checks, which are static code analysis tools we use.
 
-### Run statis analysis by hand (without using pre-commit)
```
min-xu-ai, line 70: I suggest we keep this in case people want to run them by hand. We can note this is not preferred method.

min-xu-ai (approved):
LGTM. One small thing.

anupambhatnagar:
> Will you need to update circleCI config as well since we use [requirements-text.txt](https://github.com/facebookresearch/fairscale/blob/main/.circleci/config.yml#L90).

No, all the package versions in requirements-test have moved to requirements-dev. `.pre-commit-config.yaml` needs to be kept in sync with requirements-dev.

CONTRIBUTING.md
anupambhatnagar: this can become cumbersome to maintain as we add continue more things to `pre-commit`.

anupambhatnagar: i'll add it back and mark it as deprecated.

anupambhatnagar: done

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.3] - TBD
+
+### Added
+- Sharded Grad Scaler works with cpu offload in mixed and full precision.
```
anj-s, line 10: this is missing the PR number that other entries have.

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
...
+- Chore: `Pre-Commit` will be the de-facto standard to run all linter checks (black, isort, mypy, flake8 etc.)
+on all commits locally and in CI. [#849]
+
+### Changed
+- Cleanup: Moving forward we would be testing all of our code with Python 3.9.7, CUDA 11.2 and the following three versions of PyTorch [#847]:
+  - the most recent stable version
```
anj-s, line 16: Have we added the current versions we are supporting? I don't see it in the README. We should also update the [installation requirements](https://github.com/facebookresearch/fairscale/blob/main/docs/source/installation_instructions.rst)

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
...
+  - the most recent LTS version
+  - a recent nightly build
+
+- Cleanup: Developer requirements of python packages will no longer be split into `requirements-dev.txt` and
+`requirements-test.txt`. All package requirements have been merged into `requirements-dev.txt`. [#853]
+
+### Removed
+- Cleanup: `meta.yml` in conda/recipe is no longer supported. [#853]
+- Cleanup: `requirements-test.txt` will no longer be needed as all required libraries are listed in `requirements-dev.txt`. [#853]
```
anj-s, line 25: We don't need multiple CHANGELOG entries for the same PR. You can have a single entry under Added. We don't have entries like Changed or Removed. We just have Fixed and Added and I prefer just maintaining two sections.

CONTRIBUTING.md
```diff
@@ -80,6 +84,11 @@ mypy --ignore-missing-imports --scripts-are-modules --pretty .
 
 ## Testing
 
+FairScale code is tested on Python 3.9.7, CUDA 11.2 and the following three PyTorch versions:
```
anj-s, line 87: Please add the current versions we are supporting here or in the README. We should have it written down somewhere.

CONTRIBUTING.md
```diff
@@ -57,19 +57,23 @@ outlined on that page and do not file a public issue.
   and less development overhead in maintaining an importing list.
 * Please setup pre-commit before opening up your PR.
 
-### Pre-commit
+### Pre-Commit
```
anj-s, line 60: Add a bracket like: "(Recommended)".

anj-s (approved):

CHANGELOG.md
anupambhatnagar: These cleanup changes are minor and may not even be worth mentioning, in some respect. On the other hand, I think we should have sections like Changed/Removed. How else would we inform our users if we deprecate/remove an API in the future? Also I'm following the conventions laid out in https://keepachangelog.com/en/1.0.0/

CHANGELOG.md
anupambhatnagar: I have updated the torch version in the installation requirements file.

CONTRIBUTING.md
anupambhatnagar: added versions to the readme and also added a note in the release.md to check the versions are correct before each release.

CHANGELOG.md
anupambhatnagar: fixed

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.3] - TBD
+
+### Added
+- Sharded Grad Scaler works with cpu offload in mixed and full precision. [#831]
+- Chore: `Pre-Commit` will be the de-facto standard to run all linter checks (black, isort, mypy, flake8 etc.)
```
anj-s, line 11: I would skip this since we already had a precommit config before this PR. The only change is that it will be the recommended way forward.

CHANGELOG.md
anj-s: I think keeping the sections short helps users understand what is going on with FairScale. I personally prefer only 1 comment per PR. If you end up with multiple comments that actually tells me you should be sending multiple PRs and are doing too much in 1. We can always change conventions so I prefer not adding too many sections :)  I think Changed/Removed are not very different and folks are going to have different interpretations. I agree with Deprecated. Maybe we can add that if we deprecate an API since we need to call that out.

CHANGELOG.md
anupambhatnagar: the versions have been added in the readme

PR #856, opened 2021-11-17
[not to be merged] added separate pg for reduce scatter

ngoyal2707:


anj-s:
We have a PR from @tmarkstrum so I think this can be closed.

PR #858, opened 2021-11-18
Fsdp prefetch params

ngoyal2707:
Option to prefetch previous layer's parameters in zero3 configuration. This allows to have `ReduceScatter` overlap with backward computation.

* Baseline profile: 
Screen Shot 2021-11-17 at 5 37 07 PM
The `ReduceScatter` and `AllGather` are overlapped with backward computation in zero3 config.

* Prefetch during post backward profile:
Screen Shot 2021-11-17 at 5 37 15 PM
`ReduceScatter` are perfectly overlapped with backward computation. (this gives around 20% speed up for the model I benchmarked) 

I also tried Prefetch during pre backward profile but that doesn't seem to help further, this still needs a bit more debugging

PR #865, opened 2021-11-19
prefetch params of previous fsdp module

ngoyal2707:
Simpler PR for https://github.com/facebookresearch/fairscale/pull/858

Synced with @myleott about not needing to make it configurable.


myleott:
Should we maintain `_fsdp_backward_ordering` instead of relying on `reverse(forward_ordering)`?

One advantage of the forward-ordering version is that it works on the very first step, whereas the backward-ordering version will only start working from the second step onward.

anj-s:
@ngoyal2707  Looks like there is another prefetch PR but this looks to be the latest? Give we have gone with the 2 PG solution, does it make sense to revisit this? cc @tmarkstrum

anj-s:
Since we've gone with an alternate solution  of using multiple process groups, I am closing this PR. Please feel free to reopen as needed.

PR #872, opened 2021-11-25
Update requirements-dev.txt

tanujdhiman:
## What does this PR do?
Fixes # (issue). When I am installing requirements-dev.txt file it gives an error in `numpy` library

error

Rest of the dependencies are okay. No confliction.

Thanks
Tanuj

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -27,4 +27,4 @@ docutils == 0.17
 pynvml == 8.0.4
 
 # For mypy typing
-numpy >= 1.21
+numpy == 1.19.5


anj-s:
@tanujdhiman It seems like you are not able to pull in the latest release of numpy as seen in [PyPI](https://pypi.org/project/numpy/#history). Can you try installing the latest version using == and seeing if that works. 

We have recently updated and standardized the package versions that we want to test on. I am not sure rolling back to 1.19 is the right way to go about this.

tanujdhiman:
Hi @anj-s !!

Yes I've tried other versions too. like 1.21.1, 1.21.2, 1.21.3, 1.21.4 But here is the error!!
Then I think my pip is not upgraded for fetching the latest version then I upgraded my pip too. Again I am facing this error.

min-xu-ai:
You are on windows, right? That's a different platform as we are testing.

anupambhatnagar:
@tanujdhiman Can you share the exact platform name, platform version and pip version you are using? 

We conduct our testing on linux and at present there are no plans to support other platforms.

tanujdhiman:
Thanks for replying !!

OS Platform - Windows
Pip - Latest

Thanks

anupambhatnagar:
Closing the PR as we don't plan to roll back the numpy version to support windows.

PR #879, opened 2021-11-30
[feature] [experimental] Layerwise Gradient Scaler

anupambhatnagar:
## What does this PR do?
Implements Layerwise Gradient Scaling.

Added:
* `layerwise_gradient_scaler.py` - contains the core implementation
* `test_layerwise_gradient_scaler.py` - contains the tests to verify the implementation

Minor changes:
- Added `sklearn` package to requirements-dev.txt
- Added test cases to ci_test_list1.txt
- Fixed an import statement linter warning in grad_scaler.py 

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/optim/feed_forward2.py
```diff
@@ -0,0 +1,204 @@
...
+    layers_to_scale = list()
+    for layer in model2.named_children():
+        print(layer)
+        if isinstance(layer[1], nn.Linear):
+            layers_to_scale.append(layer)
+
+    layerwise_grad_scaler = LayerwiseGradientScaler(layers_to_scale)
+    for _, layer in enumerate(layers_to_scale):
+        layer[1].register_full_backward_hook(layerwise_grad_scaler.scale_up)
+        layer[1].register_full_backward_hook(layerwise_grad_scaler.scale_down)
```
MidoAssran, line 105: why not register the hooks inside the init for the LayerwiseGradientScaler

fairscale/optim/feed_forward2.py
```diff
@@ -0,0 +1,204 @@
...
+        self.per_layer_scale = [2 ** (1 / 1000) for _ in modules]
+        # print('length of modules %s' % len(modules))
+
+    def scale_up(self, m: nn.Module, inputs: torch.Tensor, outputs: torch.Tensor) -> None:
+        if inputs[0] is not None:
+            torch.mul(inputs[0], (2 ** 13))
+
+    def scale_down(self, m: nn.Module, inputs: torch.Tensor, outputs: torch.Tensor) -> None:
+        if outputs[0] is not None:
+            torch.mul(outputs[0], 1 / (2 ** 13))
```
MidoAssran, line 40: Right now it does not associate the per_layer_scale to the module itself, perhaps consider storing the modules (or some reference to them) in a list during init, and then using this variable to pull the corresponding layer_scale

fairscale/optim/feed_forward2.py
MidoAssran, line 105: also why do we need two hooks for these? why not just do scale_up/scale_down in a single hook

fairscale/optim/feed_forward3.py
```diff
@@ -0,0 +1,133 @@
...
+        self.scaling_factors = scaling_factors
+        self.scale_up_functions: List = []
+        self.scale_down_functions: List = []
+
+    # create a scale down and scale up function for each layer
+    def make_scaling_functions(self) -> None:
+        def scale_down(m: nn.Module, inputs: torch.Tensor, outputs: torch.Tensor) -> None:
+            if outputs[0] is not None:
+                # logging.info("item value in scale down is %s", item)
+                torch.mul(outputs[0], 1 / item)
```
MidoAssran, line 88: this is the gradient being sent back to this layer, it will have been scaled by item(layer+1), so should be unscaled by item(layer+1)

fairscale/optim/feed_forward3.py
```diff
@@ -0,0 +1,133 @@
...
+
+        def scale_up(m: nn.Module, inputs: torch.Tensor, outputs: torch.Tensor) -> None:
+            if inputs[0] is not None:
+                torch.mul(inputs[0], item)
```
MidoAssran, line 92:
```
for g in inputs:
    if g is not None:
        # scale g
```

fairscale/optim/feed_forward3.py
MidoAssran, line 88:
```
for g in outputs:
    if g is not None:
        # unscale g

File: fairscale/optim/feed_forward4.py [hunk 5]
```diff
@@ -0,0 +1,163 @@
...
+                torch.mul(outputs[idx], self.scale_down_factor)
+            else:
+                logging.info("outputs[%d] is None" % idx)
+
+        # scale up the inputs
+        for idx in range(len(inputs)):
+            if inputs[idx] is not None:
+                torch.mul(inputs[idx], self.scale_up_factor)
+            else:
+                logging.info("inputs[%d] is None" % idx)
```
MidoAssran, line 95: torch.mul(inputs[idx], self.scale_up_factor * self.scale_down_factor)

MidoAssran: not changing the gradient actually being passed back

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.4.5] - TBD
 
 ### Added
+- Implemented per layer gradient scaling [#879]
```
anj-s, line 10: Can you add more details here?

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        layer_name: name of the layer e.g. fc1, conv1, relu1
+        layer_type: type of the layer e.g. Linear, Conv2d, ReLU
+        scaling_factor: user configurable scaling factor for the layer, defaults to 1.0
+        found_inf_or_nan: a boolean indicating if any parameter of layers' gradient contains inf/nan
+        growth_tracker: tracks number of step since last time scale was increased
+        scale_layer: a boolean indicating if the layer should be scaled or not
+        """
+        self.layer_name = name
+        self.layer_type = layer
```
anj-s, line 23: should this be `layer` instead of `layer_type`? type makes me thing we are storing a string rather than the layer itself.

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    A class to record the layer attributes.
+    """
+
+    def __init__(self, name: str, layer: nn.Module, scale: float = 1.0, scale_layer: bool = False) -> None:
+        """
+        layer_name: name of the layer e.g. fc1, conv1, relu1
+        layer_type: type of the layer e.g. Linear, Conv2d, ReLU
+        scaling_factor: user configurable scaling factor for the layer, defaults to 1.0
+        found_inf_or_nan: a boolean indicating if any parameter of layers' gradient contains inf/nan
```
anj-s, line 18: is this a typo: `layers'` vs `layer's`?

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+
+            def forward(self, x):
+                ...
+
+        # Create an instance of the model
+        model = ConvNet()
+        optimizer = torch.optim.SGD(model.parameters())
+
+        # specify the layers to scale and their scaling factor
+        layer_scale_dict = {"conv1": 2**10, "conv2": 2**8, "fc1": 2**10, "fc2": 2**9}
```
anj-s, line 80: (Not something to implement in this PR but just a thought)
What happens if the number of layers are in the order of 10s or 100s? Should there be difference scaling factors applied in terms of depth i.e the first 25% are scale factor 1, next are scale factor 2? I suspect @MidoAssran 's model may not be that large hence you may not need it immediately. Just something we should think about from a API perspective for use cases we may need to support. WDYT?

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: layer_name is the name as a string. layer_type is Linear, Conv2d etc. I gave examples on line 16.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: In deep models the scaling factors will have to be set carefully by the user. Initializing all layers with the same scaling factor may not produce the best results. After training and observing the gradient overflow we can determine which layers require gradient scaling. The common practice is to initialize scaling factor to a large value and let the Gradient Scaler unscale the scaling factor so that it comes into a reasonable range.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: fixed

CHANGELOG.md
anupambhatnagar: done

fairscale/optim/layerwise_gradient_scaler.py
anj-s: I guess we assign to layer which is a nn.Module hence my question.

anupambhatnagar: I can change it if that improves readability. Do let me know.

At this point I'm deep in the trenches with this code, so a fresh perspective on naming convention (and anything else) is very welcome :)

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def scale(self) -> None:
+        """
+        For each layer calculates the scaling factor for preceding layers' grad inputs
+        and current layers' grad outputs. These values are used to register a full backward
+        hook. The handle returned from registering the backward hook is appended to a list
+        of handles. New hooks are created and registered at every step and a new list of
+        handles is created. The handles are flushed out in the unscale function.
+        """
+        if self._apply_layerwise_scaling:
```
anj-s, line 166: nit: to prevent nesting I prefer early returns.

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+            for idx in range(len(self.layer_info)):
+                elt = self.layer_info[idx]
```
anj-s, line 168: what does "elt" stand for?

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+                logging.debug("name = %s \t scale = %s" % (layer_name, elt.scaling_factor))
+
+    def unscale(self) -> None:
+        """
+        For each layer, check if any of the layers' parameters contain an inf/nan.
+        If there are no inf/nan in the gradient, then gradient of that layer is
+        unscaled by the reciprocal of the scaling factor for that layer.
+        Finally, all handles recorded while registering the hooks are deleted.
+        """
+        if self._apply_layerwise_scaling:
```
anj-s, line 188: Lets do an early return statement here to prevent deep nesting.

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+            for elt in self.layer_info:
```
anj-s, line 189: Can we use a filter function to only get the layers_info objects that don't have inf or nans? That would improve the readability here I think.

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+        For each layer, check if any of the parameters with a gradient attribute
+        contain an inf/nan. If any of the parameters' gradient contain an inf/nan,
+        then that layers' found_inf_or_nan attribute is set to True and all
+        remaining parameters for that layer are skipped.
+        """
+        for elt in self.layer_info:
+            elt.found_inf_or_nan = False
+            for _, param in elt.layer_type.named_parameters():
+                if hasattr(param, "grad") and param.grad is not None:
+                    if torch.isinf(param.grad).any().item() is True or torch.isnan(param.grad).any().item() is True:  # type: ignore
```
anj-s, line 213: do we need "is True" for these statements?

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+                        elt.found_inf_or_nan = True
+                        break  # skip all remaining named parameters
+
+    def step(self, optimizer) -> None:  # type: ignore
+        """
+        If there are NO inf/nan in the gradients' of ALL layers, then optimizer
```
anj-s, line 219: Let us have all of the letters in lowercase? I know you may want to warn the user about it :)

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+        layer to 0. Else, increment the growth tracker of the layer. If growth
+        tracker equals the growth interval, then multiply the scaling factor of
+        the layer by the growth factor and reset the layers' growth tracker to 0.
+        Finally, clip the scaling factor to the range
+        [self.min_scaling_factor, self.max_scaling_factor]. The min/max scaling
+        factor values are user configurable.
+        """
+        if self._apply_layerwise_scaling:
+            for layer in self.layer_info:
+                if not layer.found_inf_or_nan:
```
anj-s, line 247: If you have if/else constructs its generally easier to read if (true): else(false) instead of having `if not (true)`.

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+            if not inf_nan_found:
+                optimizer.step()
+            self._update_scale()
+        # not using layerwise gradient scaling
+        else:
+            optimizer.step()
+
+    def _update_scale(self) -> None:
+        """
+        For each layer, if an inf/nan is found, then multiply the scaling factor
```
anj-s, line 236: Can you explain what is the logic behind this? I think you are explaining the code below but what is useful is the reasoning behind how we are attempting to update the scale.

tests/optim/test_layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def test_linear_model() -> None:
+    model1 = FeedForward(2, 10)
+    model2 = FeedForward(2, 10)
+
+    vanilla_model = train_linear_model(model1, False)
+    scaled_model = train_linear_model(model2, True)
+
+    def get_params_with_grad(trained_model):
```
anj-s, line 116: this can be factored out as a file level util i think.

tests/optim/test_layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,220 @@
...
+def test_vision_model() -> None:
+    # Remove randomness from various sources while testing.
+    torch.use_deterministic_algorithms(True)  # type: ignore
+    # set environment variable in CircleCI for test to pass: CUBLAS_WORKSPACE_CONFIG = :4096:8
+
+    m1 = SimpleConvNet()
+    m2 = SimpleConvNet()
+
+    vision_model = train_vision_model(m1, False)
+    scaled_vision_model = train_vision_model(m2, True)
```
anj-s, line 207: I am guessing that without the layerwise scaling scaled_vision_model would have inf grads? is it worth adding a test for that?

fairscale/optim/layerwise_gradient_scaler.py
```diff
@@ -0,0 +1,278 @@
...
+            for idx in range(len(self.layer_info)):
+                elt = self.layer_info[idx]
+                layer_name, layer_type = elt.layer_name, elt.layer_type
+
+                inputs_multiplier = 1.0
+                if idx > 0:
+                    inputs_multiplier = self.layer_info[idx - 1].scaling_factor
+
+                outputs_multiplier = 1.0 / elt.scaling_factor
+                helper = GradientHelper(layer_name, inputs_multiplier, outputs_multiplier)
```
anj-s, line 176: can you explain the inputs_multiplier and outputs_multiplier a bit? what are example values and how do they help scaling?

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: element

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: They are not needed but if I remove them it's not clear what the expected returned type is. If it's 1 or "I wonder what the return type is"? Both would return True.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: sure :) I'll happily assume that readers are reading my comments with full concentration.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: If there is an inf/nan, then we should reduce the scale as the current scale is causing a gradient overflow. 

On the other hand if the gradient is finite we can increase the scale, but doing so at every step would again lead to an overflow within a short number of steps so we increase the scale gradually by pacing it. 
The growth interval determines how frequently we would increase the scale and the growth tracker measures where we are in relation to the growth interval.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: done

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: done.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: done

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: agreed and changed.

tests/optim/test_layerwise_gradient_scaler.py
anupambhatnagar: yes, nice catch. done.

tests/optim/test_layerwise_gradient_scaler.py
anupambhatnagar: At present neither of the models encounter any infs/nans. It is tricky to build a test case which uses mixed precision and has inf in the gradient which stabilize with layerwise gradient scaling. These tests check that the correct scaling factor is being applied on each layer and that unscaling is working correctly.

anupambhatnagar: The test_linear_model does use mixed precision as we wrap the forward pass with the `autocast` context.

fairscale/optim/layerwise_gradient_scaler.py
anupambhatnagar: Consider this example graph:

|layer | scaling_factor	|
|--- | --- |
| input	| None 		|
| fc1   	| 128		|
| relu1	| 1 			|
| fc2 	| 512		|
| relu2 	| 1 			|
| sigmoid	| 1 			|
| output |  None |

Following the vocabulary of [register_full_backward_hook](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_full_backward_hook),

 **inputs_multiplier = grad_inputs** 
**outputs_multiplier = grad_outputs**.

During the backward pass, the grad_inputs will be multiplied by the scaling factor of the next layer and grad_outputs will be multiplied by the reciprocal of the scaling factor of the current layer.

| layer | inputs_multiplier | outputs_multiplier |
| ---- | --- | --- |
| sigmoid  | 1    | 1     |
|relu2        | 512 | 1 |
| fc2 |     1 | 1/512 |
| relu1 |  128 | 1 |
| fc1 | 1 | 1/128 | 

The inputs_multiplier value is applied to the next layer (going top to bottom in the table above). Note that 

```inputs_multiplier[i] * outputs_multiplier[i+1] = 1```

 This is the scaling and unscaling process.

Note: value of the inputs multiplier in the fc1 layer does not make any difference as there is no layer on which it is applied. 

Edge Case: What if the last layer of a graph has a scaling factor greater than 1? The solution is - append the identity layer to the graph and now we are back in the case described above.

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 13 above)
Comment (anupambhatnagar) [reply fairscale/optim/layerwise_gradient_scaler.py:189]: done

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 14 above)
Comment (anj-s) [reply fairscale/optim/layerwise_gradient_scaler.py:222]: The reason I mentioned this is that it can shorten the line since you can just check for the implicit True condition.

Comment (anupambhatnagar) [reply fairscale/optim/layerwise_gradient_scaler.py:222]: I noticed the long line length as well but then I let the linter (black) decide on how to format it.

Comment (anupambhatnagar) [reply fairscale/optim/layerwise_gradient_scaler.py:222]: So we set the line length to 120 here https://github.com/facebookresearch/fairscale/blob/562542478a145fae410cb23ffd46a8d7c2497abb/pyproject.toml#L8-L9
TIL: comment is not considered part of line length by black

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 7 above)
Comment (anj-s) [reply fairscale/optim/layerwise_gradient_scaler.py:23]: From my perspective its a layer your are tracking and not its type. No need to change it since it is a nit.

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 11 above)
Comment (anj-s) [reply fairscale/optim/layerwise_gradient_scaler.py]: Can this be modified to something more representative?

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 14 above)
Comment (anj-s) [reply fairscale/optim/layerwise_gradient_scaler.py:222]: This is not a comment and checking for is True is redundant right? should we remove it?

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 20 above)
Comment (anj-s) [reply fairscale/optim/layerwise_gradient_scaler.py]: Can you add comments and document this somewhere?

File: tests/optim/test_layerwise_gradient_scaler.py (hunk 19 above)
Comment (anj-s) [reply tests/optim/test_layerwise_gradient_scaler.py:208]: Discussed offline: there is a test where we see infs that can be used here. 

I agree that is tricky to build a tests case but unless we do we cannot test if this works. I would also suggest that we use larger models if needed to simulate this behavior as a one off test case. 

Let us add this as a TODO. Shall we move this to experimental/ in the meantime? WDYT?

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 7 above)
Comment (anupambhatnagar) [reply fairscale/optim/layerwise_gradient_scaler.py:23]: changed

File: fairscale/optim/layerwise_gradient_scaler.py (hunk 14 above)
Comment (anupambhatnagar) [reply fairscale/optim/layerwise_gradient_scaler.py:222]: removed True

File: tests/optim/test_layerwise_gradient_scaler.py (hunk 19 above)
Comment (anupambhatnagar) [reply tests/optim/test_layerwise_gradient_scaler.py:208]: I'll document this and moving it to experimental.

**Review by anj-s — APPROVED** [review]

**Comment by anj-s** [comment]
LGTM! One last thing, can you update the PR title to follow the guidelines such that it says something like "[feature] ...."

**Comment by anupambhatnagar** [comment]
> LGTM! One last thing, can you update the PR title to follow the guidelines such that it says something like "[feature] ...."

File: fairscale/optim/layerwise_gradient_scaler.py [hunk 21]
```diff
@@ -0,0 +1,289 @@
...
+        Finally, all handles recorded while registering the hooks are deleted.
+        """
+        if not self._apply_layerwise_scaling:
+            return
+
+        layers_with_finite_values = self._get_layers_with_finite_values()
+        for item in layers_with_finite_values:
+            for param_name, param in item.layer.named_parameters():
+                if hasattr(param, "grad"):
+                    logging.debug("%s scaling down %s by %s" % (item.layer_name, param_name, 1.0 / item.scaling_factor))
```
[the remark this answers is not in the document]
anupambhatnagar: yup, I did write that on L221 but somehow missed it here.

PR #880, opened 2021-12-01
[fix] [FSDP] Do not lose original reshard_after_forward

### LINKED ISSUE #878: [FSDP] reshard_after_forward = False for root instance
```
if self._is_root:
    # Buffers stay on GPU, and don't get sharded. Since _cast_buffers
    # applies recursively, we only call this from the root instance.
    self._cast_buffers()
            
    # Don't free the full params for the outer-most (root) instance,
    # since those params will be needed immediately after for the
    # backward pass.
    self.reshard_after_forward = False

    # Due to the use of streams, we need to make sure the previous
    # ``optim.step()`` is done before we all-gather parameters.
    self._wait_for_previous_optim_step()
```
According to the following code (from fully_sharded_data_parallel.py), we can find that reshard_after_forward is set to False for root instance (for better performance). However, when we use autowrap() to wrap model, we only wrap leaf nodes, i.e., every leaf is an FSDP root instance. In this case, every FSDP instance's reshard_after_forward will be set to False even if we set reshard_after_forward to True (in initialization). However, due to a memory issue, I still want to reshard after forward when wrapping with autowrap(). 
I have a few questions,
1. Could I just comment "self.reshard_after_forward = False" as an work around solution?
2. Could you add a parameter in __init__() of FullyShardedDataParallel to decide whether self.reshard_after_forward need to be set to False for root instance?
3. For the FSDP wrapping using autowrap(), is the performance affected by the reshard_after_forward? Since the FSDP instances in this case are not outer-most (root) instances, these params will not be needed immediately for the backward pass

min-xu-ai:
- In a corner case we can lose this value
- Saving it and use it in the reset function fixed it
- A trivial case probably not worth a dedicated test for now

## What does this PR do?
Fixes #878 .

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [x] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Fixed a corner case of FSDP init order and losing one of the flags [#880]
 
 ## [0.4.3] - 2021-11-18
 
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -307,7 +307,7 @@ class FullyShardedDataParallel(nn.Module):
         self.process_group = process_group or get_process_group_cached()
         self.rank = self.process_group.rank()
         self.world_size = self.process_group.size()
-        self.reshard_after_forward = reshard_after_forward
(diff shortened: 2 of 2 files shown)


min-xu-ai:
A small fix while I was using mevo and nllb.

I think we need a better way to ensure not just this flag but all flags are correctly reset. But that's out of scope of this PR.

tmarkstrum (approved):

PR #881, opened 2021-12-02
Fix for Key Error that can happen in certain FSDP wrapping scenarios of Huggingface model sub-modules (issue #876)

### LINKED ISSUE #876: Auto wrapping Huggingface models with FullyShardedDataParallel can break them
## 🐛 Bug

If you auto wrap a HuggingFace model with `FullyShardedDataParallel`, depending on what sub-modules are wrapped, and how the output of the modules are accessed in the model, calling `forward` can result in exceptions. 

## What happens

Some sub-modules of the HuggingFace model output a child implementation of [`ModelOutput`](https://github.com/huggingface/transformers/blob/8b240a06617455eae59e1116af6a1a016664e963/src/transformers/file_utils.py#L1884), which is itself a descendent of `OrderedDict`.

An example is, for instance, [`CausalLMOutputWithCrossAttentions` ](https://github.com/huggingface/transformers/blob/8b240a06617455eae59e1116af6a1a016664e963/src/transformers/modeling_outputs.py#L375).

You can access attributes of `ModelOutput` instance derivatives in three ways, e.g. like this:

- `hidden_states = transformer_outputs[0]`
- `hidden_states = transformer_outputs.hidden_states`
- `hidden_states = transformer_outputs["hidden_states"]`

In the HuggingFace library these three types of accessing attributes are used interchangeably.

Now, when a sub-module, which outputs a `ModelOutput` derived instance,  is wrapped with `FullyShardedDataParallel`, this output instance is converted in to an `OrderedDict`, in the [`apply_to_tensors`](https://github.com/facebookresearch/fairscale/blob/fecb665b812b6bfc38442e1fb1557e21508917f4/fairscale/utils/containers.py#L21) function, in `utils/containers.py`, since the `ModelOutput` derives from `OrderedDict`! :

```
        elif isinstance(x, OrderedDict):
            od = OrderedDict()
            for key, value in x.items():
                od[key] = _apply(value)
            return od
```

After it has been converted into a plain `OrderedDict` by `FullyShardedDataParallel`, when the output of the submodule is, for instance, accessed using an index (`transformer_outputs[0]`) or attribute accessor (`transformer_outputs. hidden_states`), an exception is thrown, because the `OrderedDict` instance doesn't know how to handle that. 

## A simple solution

A simple solution is to replace the `OrderedDict` instantiation above with `od = x.__class__()`, thus:

```
        elif isinstance(x, OrderedDict):
            od = x.__class__()
            for key, value in x.items():
                od[key] = _apply(value)
            return od
```

In this way we keep on using the original class, with its extended means to access attributes, which is used in the HuggingFace library.

## To Reproduce

1. Use a HuggingFace model, e.g. `GPT2LMHeadModel`
```
from transformers import GPT2Config, GPT2LMHeadModel
```

2. Build the model 
```
        # You could also try, e.g. "gpt2-large"
        model_config = GPT2Config.from_pretrained("gpt2")
        self._model = GPT2LMHeadModel(model_config)

        self._model.to(self._device)
```

3. Auto wrap the model
```
                wrap_policy = functools.partial(default_auto_wrap_policy,
                                                min_num_params=1e8)

                fsdp_params = {
                    "mixed_precision": False,
                    "move_params_to_cpu": False
                }

                with enable_wrap(wrapper_cls=FullyShardedDDP, auto_wrap_policy=wrap_policy, **fsdp_params):
                    self._training_model = auto_wrap(self._model)

                # Finally wrap root module
                self._training_model = FullyShardedDDP(self._training_model, **fsdp_params)
```

4. Use this in a training loop

This results in errors similar to this one where the `OrderedDict` accessed in invalid ways:
```
... transformers/models/gpt2/modeling_gpt2.py", line 1057, in forward
    hidden_states = transformer_outputs[0]
KeyError: 0
```

## Expected behavior

No exception should occur, `FullyShardedDataParallel` should not break calling the HuggingFace model.

## Environment

```
$ pip show fairscale
Name: fairscale
Version: 0.4.3

$ pip show transformers
Name: transformers
Version: 4.12.5

$ pip show torch
Name: torch
Version: 1.10.0

$ python --version
Python 3.7.4
```

visionscaper:
## What does this PR do?
Fixes #876  .

## Before submitting

- [x] Did you have fun?

Yes

Not required

- [ ] Did you update the [changelog]
I don't think this is significant enough for an entry in the change log.

## PR review

### DIFF
--- a/fairscale/utils/containers.py
+++ b/fairscale/utils/containers.py
@@ -19,7 +19,8 @@ def apply_to_tensors(fn: Callable, container: Union[torch.Tensor, Dict, List, Tu
         if torch.is_tensor(x):
             return fn(x)
         elif isinstance(x, OrderedDict):
-            od = OrderedDict()
+            od = x.__class__()
+
             for key, value in x.items():
                 od[key] = _apply(value)
             return od
--- a/fairscale/utils/containers.py
+++ b/fairscale/utils/containers.py
@@ -20,7 +20,6 @@ def apply_to_tensors(fn: Callable, container: Union[torch.Tensor, Dict, List, Tu
             return fn(x)
         elif isinstance(x, OrderedDict):
             od = x.__class__()
-
             for key, value in x.items():
                 od[key] = _apply(value)
             return od
--- a/tests/ci_test_list_3.txt
+++ b/tests/ci_test_list_3.txt
@@ -22,3 +22,4 @@ tests/optim/test_oss.py
 tests/optim/test_oss_adascale.py
 tests/optim/test_ddp_adascale.py
 tests/experimental/nn/data_parallel/test_gossip.py
+tests/nn/data_parallel/test_fsdp_hf_transformer_eval.py
--- a/tests/nn/data_parallel/test_fsdp_hf_transformer_eval.py
+++ b/tests/nn/data_parallel/test_fsdp_hf_transformer_eval.py
@@ -52,7 +52,7 @@ class TransformerWithCustomOutput(nn.Transformer):  # type: ignore[name-defined]
         return TransformerOutput(output=output)
 
 
-class DecoderWithLMHead(nn.Module):
+class TransformerWithLMHead(nn.Module):
     def __init__(self, d_vocab=100, d_model=16):
         super().__init__()
         self.d_vocab = d_vocab
@@ -97,7 +97,7 @@ class TestHFTransformersAutoWrap(unittest.TestCase):
         self.device = torch.device("cuda")
 
         print("Build model ...")
-        self.model = DecoderWithLMHead()
+        self.model = TransformerWithLMHead()
         self.model.to(self.device)
 
     def test_auto_wrap_hf_model(self):
--- a/tests/nn/data_parallel/test_fsdp_hf_transformer_eval.py
+++ b/tests/nn/data_parallel/test_fsdp_hf_transformer_eval.py
@@ -1,32 +1,88 @@
+from collections import OrderedDict
+from dataclasses import dataclass
 import tempfile
 import unittest
 
 import torch
+from torch import nn
 
 from fairscale.nn import FullyShardedDataParallel as FSDP
 from fairscale.nn import auto_wrap, enable_wrap
 from fairscale.utils.testing import dist_init
 
-try:
-    from transformers import GPT2Config, GPT2LMHeadModel, GPT2Tokenizer
-    from transformers.models.gpt2.modeling_gpt2 import GPT2Model
-except ImportError:
-    raise unittest.SkipTest(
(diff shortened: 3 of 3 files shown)


visionscaper:
I have chosen to fix this simply by replacing `od = OrderedDict()` with `od = x.__class__()`.

Another option would have been to revert back to `od = OrderedDict()` if an exception is raised during construction, e.g. due to the presence of positional arguments:

```
try:
    od = x.__class__()
except:
    # backup solution
   od = OrderedDict()
```

However, my reasoning is that in case of an exception during construction, this exception should be raised in any case. This is because in that case it is also likely that using an `OrderedDict` will cause issues later on since the user is doing something with their class that we don't model with our `OrderedDict`.

visionscaper:
The test I wrote fails on the current FairScale `main` branch, with my fix the test passes.

visionscaper:
An important remark about the test : it requires the Huggingface `transformer` package installed. Since I don't know how you would like to deal with that, I have added code that skips the test when this package is not installed.

visionscaper:
I see that the tests fail on CircleCI because I didn't add my new test to a "test list". How does this need to be resolved?

min-xu-ai:
> I see that the tests fail on CircleCI because I didn't add my new test to a "test list". How does this need to be resolved?

Thanks for adding the test! Yes, please add your test to one of the list txt files. Perhaps end of 3rd list file will be good.

min-xu-ai (approved):
LGTM!

visionscaper:
Hi @min-xu-ai

anj-s:
> Hi @min-xu-ai, I will add the test to the list, but how do you want me to handle the Huggingface transformers package dependency? I could add it to ‘requirements-dev.txt’ …

@visionscaper Is there any way we could add a test that is not dependent on the HF library. Ideally FairScale should not have any dependencies other than PyTorch.

visionscaper:
Hi @min-xu-ai, if you even don't want it as a development dependency than we have to implement a minimal model for which a sub-module also uses a custom `OrderedDict` as output in the same way as Huggingface does.

I think in some of the other FSDP unit test files a test model is used; maybe I can use that, I will look in to it.

visionscaper:
Hi @min-xu-ai @tmarkstrum, I have adapted my test to be independent of the Huggingface transformer library and added it to the test list. As you can see all tests pass now. 🎉

Let me know if you have any questions.

min-xu-ai:
Wow. Awesome. Thanks so much!

anj-s:
> Hi @min-xu-ai @tmarkstrum, I have adapted my test to be independent of the Huggingface transformer library and added it to the test list. As you can see all tests pass now. 🎉
> 
> Let me know if you have any questions.

Thanks @visionscaper for adding the test! LGTM from my side.

anj-s (approved):

PR #882, opened 2021-12-03
WIP: Fsdp multistream overlap

tmarkstrum:
## What does this PR do?
Fixes the issue that all_gather and reduce_scatter NCCL operations do not overlap with other computation streams in FSDP backward propagation.
The FSDP profile before the fix
Screen Shot 2021-12-03 at 2 37 34 PM

The FSDP profile after the fix.
Screen Shot 2021-12-03 at 2 23 37 PM

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -283,7 +283,7 @@ class FullyShardedDataParallel(nn.Module):
     def __init__(
         self,
         module: nn.Module,
-        process_group: Optional[ProcessGroup] = None,
+        process_group: Optional[Dict[str, ProcessGroup]] = None,
```
min-xu-ai, line 286: As we have discussed a bit last week, I think we can make this backward compatible. We can either (1) keep this unchanged and add a new param for the new PG or (2) make this argument check whether the type given is a PG or a Dict. I think perhaps (1) is easier and you basically add a new optional PG argument for reduce scatter. Also, please update the docstring and explain the reason for adding the new PG.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -619,10 +619,15 @@ def clip_grad_norm_(
         local_norm = calc_grad_norm(params_with_grad, norm_type).cuda()
         if norm_type == inf:
             total_norm = local_norm
-            dist.all_reduce(total_norm, op=torch.distributed.ReduceOp.MAX, group=self.process_group)
+            dist.all_reduce(
```
min-xu-ai, line 622: The all_reduce here is after both fwd and bwd passes are done, right? We need to pick the right PG and async flags for it. I think what you picked here (i.e. using the scatter reduce PG and use async=False) makes sense. But please use profile and dump and trace and let's review the trace on this.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1879,10 +1886,12 @@ def update_p_data(custom_output_tensor: Optional[torch.Tensor] = None) -> None:
                     # Fill output_tensor with (p.data for each shard in self.world_size)
                     if hasattr(dist, "_all_gather_base") and enable_nccl_base_collectives:
                         # New version of PyTorch has all_gather_base, which is faster than chunk and then all_gather.
-                        dist._all_gather_base(output_tensor, p_data, group=self.process_group)
+                        dist._all_gather_base(
+                            output_tensor, p_data, group=self.process_group["all_gather_group"], async_op=False
```
min-xu-ai, line 1890: We can probably add 2 internal variables like self._async_all_gather and self._async_reduce_scatter and use those variable instead of True/False values directly in this line and line 1894. This is less likely to cause bugs where we change one to be async and leaving another to be not async.

fairscale/utils/parallel.py
```diff
@@ -58,7 +58,7 @@ def enable_pytorch_sync_bn(module: torch.nn.Module) -> None:
             layer._specify_ddp_gpu_num(1)  # type: ignore
 
 
-def get_process_group_cached(ranks: Optional[Sequence[int]] = None) -> ProcessGroup:
+def get_process_group_cached(ranks: Optional[Sequence[int]] = None) -> Dict[str, ProcessGroup]:
```
min-xu-ai, line 61: I suggest that we don't make this function be aware of specifics of all_gather or reduce_scatter. It should be able to support general and arbitrary keys.

In other words, I think the cache should change from using only `rank` as the key to a tuple `(name, rank)` as the key. This way, the caller just supply a `name` and a list of ranks and then, this will return either a new group or a cached group without caring about the actual value of the key.

fairscale/utils/parallel.py
```diff
@@ -25,7 +25,7 @@ def chunk_and_pad(tensor: torch.Tensor, num_chunks: int) -> List[torch.Tensor]:
     return chunks
 
 
-def validate_process_group(device: torch.device, process_group: ProcessGroup) -> None:
+def validate_process_group(device: torch.device, process_group: Dict[str, ProcessGroup]) -> None:
```
min-xu-ai, line 28: similar to the comment below, I think we can make this function not aware of the specifics of all_gather and reduce_scatter.

fairscale/utils/reduce_scatter_bucketer.py
```diff
@@ -35,11 +35,17 @@ def flush(self) -> None:
         # reduce-scatter bucket
         if hasattr(dist, "_reduce_scatter_base") and enable_nccl_base_collectives:
             dist._reduce_scatter_base(
-                self.output_shard[: self.offset], self.data[:, : self.offset].contiguous(), group=self.group
+                self.output_shard[: self.offset],
+                self.data[:, : self.offset].contiguous(),
+                group=self.group,
+                async_op=False,
```
min-xu-ai, line 41: similar to the comment above, let's use a variable instead of True/False values in multiple places in this file.

stubs/torch/distributed/__init__.pyi
```diff
@@ -62,8 +62,8 @@ def all_reduce(tensor: Tensor, op: ReduceOp = ReduceOp.SUM, group:Optional[Proce
 def all_gather(tensor_list: List[Tensor], tensor: Tensor, group:Optional[ProcessGroup] = None, async_op: bool = False): ...
 def reduce_scatter(tensor: Tensor, input_list: List[Tensor], op: ReduceOp = ReduceOp.SUM, group:Optional[ProcessGroup] = None, async_op: bool = False): ...
 # These two functions takes flatten tensors directly, avoiding internal buffer allocations overheads.
-def _all_gather_base(input_tensor: Tensor, output_tensor: Tensor, group:Optional[ProcessGroup] = None): ...
-def _reduce_scatter_base(output_tensor: Tensor, input_tensor: Tensor, group:Optional[ProcessGroup] = None): ...
+def _all_gather_base(input_tensor: Tensor, output_tensor: Tensor, group:Optional[ProcessGroup] = None, async_op: bool = False): ...
```
min-xu-ai, line 65: these 2 lines of updates are great. If you want, you can separate them into a PR and get reviewed and merge it in first.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: (1) sounds good.

tmarkstrum:
Closed this PR because #897 PR is a backward compatible solution.

PR #885, opened 2021-12-13
FullyShardedDataParallel: only return full state dict on rank 0

### LINKED ISSUE #844: FullyShardedDataParallel: only return full state dict on rank 0
## Motivation

Currently, `FullyShardedDataParallel.state_dict` returns full state dict on all ranks. If `state_dict_device` is CPU (https://github.com/facebookresearch/fairscale/blob/main/fairscale/nn/data_parallel/fully_sharded_data_parallel.py#L2221), the memory footprint of the operation will be `[SIZE_OF_FULL_STATE_DICT] * [NUM_DEVICES_PER_NODE]`. This can sometimes cause hosts without beefy RAM to OOM.

## Pitch

For the purpose of checkpointing, we only need full state dict on a single rank. The proposal is to make `FullyShardedDataParallel.state_dict` only return full state dict on rank 0, and return `None` or empty dict (whichever makes more sense) on non-rank 0. Doing so will allow `FullyShardedDataParallel` to skip the GPU -> CPU copy on non-rank 0 altogether.

four4fish:
## What does this PR do?
Fixes #844 
Updated pytorch_nightly dependences as 20211101 version is not available anymore
Available list is here: https://download.pytorch.org/whl/nightly/cu111/torch_nightly.html

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -2381,6 +2381,9 @@ def _post_state_dict_hook(
     # recursively, so we need to make sure that we only clone each tensor at
     # most once. Thus we add an attribute on the tensor called "_has_been_cloned"
     # which keeps track of tensors that are no longer at risk of being freed.
+    if dist.get_rank() != 0:
```
anj-s, line 2384: Will this be a backwards incompatible change?

four4fish: This is a bug fix to avoid OMM, it shouldn't impact correctness or user facing behavior.

anj-s: Sure, I understand the OOM issue. Let me rephrase my question :) Previously, we were returning a full state dict on all ranks. With this chance we will be returning an empty state dict on some of the ranks when users call `state_dict`.  Is that correct? Would something break if users execute something based on this assumption?

min-xu-ai:
It would really nice to add a unit test for this as well.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
four4fish: Thank you for the explanation. Please correct me if I'm wrong. If user call `_post_state_dict_hook` on a rank, this will be a breaking change. My understanding is that `_post_state_dict_hook` is for colon tensor and copy from rank 0 is sufficient?  Do we have internal meta use case or other oss use case that user calls `_post_state_dict_hook` on a rank?

anj-s: Can you explain this "colon tensor and copy from rank 0 is sufficient?" statement?  _post_state_dict_hook is called every time a user calls state_dict on the model. This is a post processing function in a sense. 
Can we add a flag to make this optional? We can then think of this being the default in the future?

four4fish: >Can you explain this "colon tensor and copy from rank 0 is sufficient?" statement?

It was from reading the comments of this function.  I was curious that if function's goal is to copies tensor, does only copy from rank 0 be enough?

>Can we add a flag to make this optional? We can then think of this being the default in the future?
Yeah make sense, will do!

anj-s:
@four4fish Do let me know when the PR is ready to review. Thanks!

four4fish:
> @four4fish Do let me know when the PR is ready to review. Thanks!

Thank you! I will try to add a unit test and open the PR. Still trying to figure out how to test different state_dict() result on different rank, any suggestions or examples?

min-xu-ai:
> > @four4fish Do let me know when the PR is ready to review. Thanks!
> 
> @anj-s Thank you! I will try to add a unit test and open the PR. Still trying to figure out how to test different state_dict() result on different rank, any suggestions or examples?

For testing, you can have the test code call state_dict() on all ranks and then test the results differently using if rank == x: .... Feel free to send the draft test and we can take a look.

four4fish:
@anj-s @min-xu-ai Thank you for the feedbacks, I have added tests and ready for review. Thanks

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -300,6 +300,7 @@ def __init__(
         force_input_to_fp32: bool = False,
         verbose: bool = False,
         cpu_offload: bool = False,
+        state_dict_on_rank_0_only: bool = False,
```
ananthsub, line 303: add a docstring as well?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -2373,14 +2380,21 @@ def alloc_storage_(data: torch.Tensor, size: torch.Size) -> None:
 
 
 def _post_state_dict_hook(
-    module: FullyShardedDataParallel, state_dict: "OrderedDict[str, torch.Tensor]", prefix: str, *args: Any
+    state_dict_on_rank_0_only: bool,
```
anj-s, line 2383: can we move this after prefix string so that we add to it after current args?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -2373,14 +2380,21 @@ def alloc_storage_(data: torch.Tensor, size: torch.Size) -> None:
...
+    prefix: str,
+    *args: Any,
 ) -> "OrderedDict[str, torch.Tensor]":
     # Assuming we are in a ``summon_full_params()`` context, we need to clone
     # each tensor so that it does not get freed (in-place) when the context
     # exits. At the same time, this hook can be called multiple times
     # recursively, so we need to make sure that we only clone each tensor at
     # most once. Thus we add an attribute on the tensor called "_has_been_cloned"
     # which keeps track of tensors that are no longer at risk of being freed.
+    if state_dict_on_rank_0_only and dist.get_rank() != 0:
```
anj-s, line 2395: Can we add a comment for why we need this? Also move the previous comment from line 2389-2394 below it since it is not relevant for this new addition.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -278,6 +278,11 @@ class FullyShardedDataParallel(nn.Module):
             The `OffloadConfig` object is used to specify the type of offload (i.e SSD, CPU) and
             other required knobs when offloading parameters from GPU. Currently the OffloadConfig
             only supports specifying SSD offload as an option. Note: This is an experimental feature.
+        state_dict_on_rank_0_only (bool):
+            When set to ``True``, ``model.state_dict()`` will only returns full state dict on
```
anj-s, line 282: nit: "return a full state dict on rank 0 and an empty state dict on all other ranks." Another thing to mention here as well as in the comments is that this is to prevent OOMs which was the original reason for this change if I remember correctly.

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -361,6 +368,13 @@ def test_delayed_reduce_scatter(self):
         test_fn = functools.partial(self._test_identical_outputs, model_fn, config)
         spawn_and_init(test_fn)
 
+    @parameterized.expand([[True], [False]], name_func=rename_test)
+    def test_state_dict_on_rank_0_only(self, state_dict_on_rank_0_only):
+        config = {"state_dict_on_rank_0_only": state_dict_on_rank_0_only}
+        model_fn = functools.partial(NestedWrappedModuleWithDelay)
```
anj-s, line 374: nit: you can use "TransformerWithSharedParams" which is more generic IMO.

anj-s:
> @anj-s @min-xu-ai Thank you for the feedbacks, I have added tests and ready for review. Thanks

Thank you @four4fish  for the PR! looks great. Can you also make an addition to the CHANGELOG.md file about this flag addition.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
four4fish: I don't think so.. 
functools.partial(func_x, a) is creating a new function func_y which doesn't need param a, that's what we need base on how the hook defined here https://github.com/pytorch/pytorch/blob/0460324b9bf7182dde9fa17bd498f0440e41c345/torch/nn/modules/module.py#L1246-L1255, which will take a method as input and the params pass into the method is positional argument *arg.
If we move the state_dict_on_rank_0_only to the last and define `self._register_state_dict_hook(functools.partial(_post_state_dict_hook, state_dict_on_rank_0_only=self.state_dict_on_rank_0_only))` we will get ` func() got multiple values for keyword argument state_dict_on_rank_0_only` error

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
four4fish: Added at the end :)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.4.5] - TBD
 
 ### Added
+- Added state_dict_on_rank_0_only flag allow user choose to return full state dict on rank 0 and return empty dict non-rank 0 to prevent OOM [#844]
```
anj-s, line 10: Thank you for updating the CHANGELOG. nit: can you add "[FSDP]" to the start of the sentence so that we know which API was modified?

four4fish: Done

anj-s (approved):
Thank you for the PR!

anj-s:
I'll merge the PR once the tests are green.

four4fish:
> 

Thank you! Do we know when the next release will be?

anj-s:
> > 
> 
> Thank you! Do we know when the next release will be?

We are planning to cut a release early next week. let me know if you need it to be done before for some reason.

four4fish:
> > > 
> > 
> > 
> > Thank you! Do we know when the next release will be?
> 
> We are planning to cut a release early next week. let me know if you need it to be done before for some reason.

Next week works, I wanted to get a ETA to know when will be a good time for us to update Pytorch Lightning fsdp plugin. Thanks

PR #887, opened 2021-12-15
Enabling ssd_offload training basic tests.

another-pjohnson:
## What does this PR do?
This PR completes the initial changes required to support training w/ ssd_offload enabled

 * Add SsdFlatParameter paralleling FlatParameter
 * Removing SsdBuffer and SsdTensorHandleView as SsdFlatParameter replaces the need for both of these.
 * Adding unit tests to fairscale/tests/experimental/nn/test_ssd_offload.py
 * Add full training test to fairscale/tests/nn/data_parallel/test_fsdp_offload.py

## Before submitting

- [ X] Did you have fun?
- [ X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/main/CONTRIBUTING.md)?
  - [X ] N/A
  - [ X] N/A
  - [X ] fairscale/tests/experimental/nn/test_ssd_offload.py
  - [ X] fairscale/tests/nn/data_parallel/test_fsdp_offload.py
  - [X ] N/A

## PR review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Fixed a corner case of FSDP init order and losing one of the flags [#880]
+- FSDP: Adding basic training support for SSD Offload, it now only supports flattened parameters. It remains an experimental feature.
 
 ## [0.4.3] - 2021-11-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Fixed a corner case of FSDP init order and losing one of the flags [#880]
-- FSDP: Adding basic training support for SSD Offload, it now only supports flattened parameters. It remains an experimental feature.
+- FSDP: Adding basic training support for SSD Offload, it now only supports flattened parameters. Renamed OffloadConfig.ssd_filepath_dir to more generic OffloadConfig.dir. SSD Offload remains an experimental feature. [#887]
 
 ## [0.4.3] - 2021-11-18
 
--- a/fairscale/experimental/nn/ssd_offload.py
+++ b/fairscale/experimental/nn/ssd_offload.py
@@ -9,12 +9,10 @@ from enum import Enum, auto
 from functools import reduce
 import io
 import os
-import pickle
-from typing import IO, Any, BinaryIO, Callable, Iterator, List, Optional, Sequence, Tuple, Type, Union
+from typing import Any, Iterator, List, Optional, Sequence, Tuple, Type
 
 import numpy as np
 import torch
-from torch.serialization import DEFAULT_PROTOCOL as DEFAULT_PROTOCOL
 
 try:
     from torch.utils._pytree import tree_map
@@ -240,38 +238,6 @@ class SsdTensorHandle(torch.Tensor):
         return r
 
 
-# Classes supporting torch.save/load
-class TorchSaver:
-    def __init__(self) -> None:
-        self.pickle_module = DisableMemoizationPicklerModule
-
-    def save(
-        self, obj: Any, f: Union[str, os.PathLike, BinaryIO, IO[bytes]], pickle_protocol: int = DEFAULT_PROTOCOL
-    ) -> None:
-        torch.serialization.save(
-            obj, f, self.pickle_module, pickle_protocol=pickle_protocol, _use_new_zipfile_serialization=False
-        )
-
-
-class SsdParameter(torch.nn.Parameter, SsdTensorHandle):
-    @classmethod
-    def from_tensor(cls: Type[SsdParameter], tensor: SsdTensorHandle) -> SsdParameter:  # type: ignore
-        r = cls(tensor.shape, tensor.dtype, tensor.requires_grad)
-        r.tensor = tensor
-        return r
-
-    @staticmethod
-    def __new__(
-        cls: SsdParameter, shape: Tuple[int, ...], dtype: torch.dtype, requires_grad: bool = True
-    ) -> SsdParameter:
-        r = SsdTensorHandle._make_wrapper_subclass(cls, shape, dtype=dtype, requires_grad=requires_grad)  # type: ignore
-
-        return r
-
-    def __init__(self, shape: Tuple[int, ...], dtype: torch.dtype, requires_grad: bool = True) -> None:
-        super(SsdParameter, self).__init__(shape, dtype, requires_grad)  # type: ignore
-
-
 class SsdFlatParameter(torch.nn.Parameter, SsdTensorHandle):
     """A parameter that is initialized from a list of parameters and can be
     turned into a list of views as needed.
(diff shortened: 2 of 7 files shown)


fairscale/experimental/nn/ssd_offload.py
```diff
@@ -229,92 +240,215 @@ def unwrap(e: Any) -> torch.Tensor:
...
+        p = pickle.Pickler(data_buf, protocol)
+        p.fast = True
+        return p
+
+    @classmethod
+    def dump(cls, obj: Any, f: io.BytesIO, protocol: int) -> None:
+        pickle.dump(obj, f, protocol)
+
+
+class FileChunkingIterator:
```
anj-s, line 382: Shall we add this in a followup PR since this and SsdTensor is not used anywhere?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -335,6 +336,9 @@ def __init__(
         if self.fp32_reduce_scatter and not self.mixed_precision:
             raise ValueError("fp32_reduce_scatter requires mixed_precision=True")
 
+        if self.ssd_offload and not self.flatten_parameters:
+            raise ValueError("ssd_offload requires flatten_parameters=True")
```
anj-s, line 340: Error message should say `OffloadConfig`'s offload type. since ssd_offload is specific to the implementation

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -109,9 +110,9 @@ class OffloadConfig:
     """Class for specifying all arguments related to offloading parameters."""
 
     # Offload type: currently only supports: "ssd_offload"
-    offload_type: str = None
+    offload_type: Optional[str] = None
```
anj-s, line 113: Should we specify a default as well if we are making this optional?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -109,9 +110,9 @@ class OffloadConfig:
...
     # Path to the directory for storing parameters offloaded to disk.
-    ssd_filepath_dir: str = None
+    ssd_directory: Optional[str] = None
```
anj-s, line 115: nit:just `dir` to avoid verbosity?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -358,13 +362,14 @@ def __init__(
         # TODO(anj): Should we conditionally do this only if we have params?
         # TODO(anj): Figure out if we can allocate the buffer during sharding.
         self.buffer_size = sum(p.numel() for p in params)
+        self.ssd_directory = ""
```
anj-s, line 365: move this below the boolean check for self.ssd_offload?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -358,13 +362,14 @@ def __init__(
...
         if self.ssd_offload:
             assert import_ssd_offload, "We need to import ssd_offload.py to enable the `ssd_offload` feature."
-            self.ssd_buffer_filepath_dir = (
-                offload_config.ssd_filepath_dir if offload_config.ssd_filepath_dir else tempfile.gettempdir()
+            self.ssd_directory = (
+                offload_config.ssd_directory
+                if (offload_config and offload_config.ssd_directory)
+                else tempfile.gettempdir()
```
anj-s, line 371: should we just make tempfile.gettempdir() the default value?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -676,13 +683,7 @@ def _shard_parameters_(self) -> None:
...
             if not p._is_sharded:
                 if self.ssd_offload:
-                    # Insert tensor into the SSD buffer and free parameter storage.
-                    p._is_sharded = False
-                    self.numel_padded_per_param.append(0)
-                    p._shard_size = p.data.size()  # type: ignore
-                    p._handle = self.ssd_buffer.insert(p.data)  # type: ignore
-                    free_storage_(p.data)
-                    continue
+                    pass
```
anj-s, line 686: maybe we can just check for `if not self.ssd_offload`.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1072,14 +1054,14 @@ def summon_full_params(self, recurse: bool = True, volatile: bool = False) -> Ge
...
                     if self.ssd_offload:
                         # Store tensors in the SSD buffer and free param storage.
                         for p in self.params:
-                            p._shard_size = p.data.size()  # type: ignore
-                            p._handle = self.ssd_buffer.insert(p.data)  # type: ignore
-                            free_storage_(p.data)
-                        self.ssd_buffer.to_disk()
+                            assert isinstance(p, SsdFlatParameter)
+                            p.to_file()
+                        # self._use_fp32_param_shard()
```
anj-s, line 1062: remove commented line?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1153,6 +1135,7 @@ def _init_param_attributes(self, p: Parameter) -> None:
             return
 
         # A single shard of the parameters in full precision.
+        # PJ this will cause memory leakage with ssd
```
anj-s, line 1138: Is there a followup TODO here or just a caution for readers? We use properties of p to set other attributes but don't deal with the data itself. Wondering what kind of guard we can add here to prevent errors. Maybe we should add a TODO to followup :)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1206,9 +1189,12 @@ def _init_param_attributes(self, p: Parameter) -> None:
             # shard in pinned memory so that we can do a non-blocking transfer.
```
anj-s, line 1189: maybe move the comment about the pinning to where we actually pin the memory.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1365,7 +1348,9 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
     @torch.no_grad()
     def _free_ssd_offload(self) -> None:
         if self.ssd_offload:
-            self.ssd_buffer.to_disk()
+            for p in self.params:
+                assert isinstance(p, SsdFlatParameter)
+                p.to_file(permit_when_tensor_none=True)
```
anj-s, line 1353: why do we need the permit_when_tensor_none bool? what condition requires this check?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1405,7 +1390,9 @@ def _pre_backward_hook(*unused: Any) -> None:
             # Note, both ``self._rebuild_full_params`` and ``self._use_full_params`` are
             # idempotent.  So in case they are called unnecessarily, they don't incur much
             # overhead.
-            if self.ssd_offload or self.reshard_after_forward:
+            # PJ TODO: Revisit if this is still needed for ssd_offload
+            # if self.ssd_offload or self.reshard_after_forward:
+            if self.reshard_after_forward:
```
anj-s, line 1395: I had always set it to delete the params in memory in the previous implementation. In the new one perhaps we don't need it. I remember fixing a particular test using this condition. Lets remove the self.ssd_offload for now and see what happens. Will try to remember what i fixed in the meantime.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1730,7 +1717,8 @@ def _finalize_parameters(fsdp_module: FullyShardedDataParallel) -> None:
         for m in self.modules():  # includes self
             if isinstance(m, FullyShardedDataParallel):
                 _finalize_parameters(m)
-                self._free_ssd_offload()
+                # PJ: I don't understand this, should it be m._free_ssd_offload?
```
anj-s, line 1720: Looks right to me. I might have had a nested loop of freeing memory previously.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: This is purely a mypy linter fix. Since we initialize it as None, and that doesn't match type (str) mypy complains.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: I can't do that because mypy will complain about ssd_directory being undefined on line 388:
```
 387         self._fsdp_wrapped_module: nn.Module = FlattenParamsWrapper(
 388             module, param_list=to_be_flatten_params, ssd_offload=self.ssd_offload, ssd_directory=self.ssd_directory
 389         )
```

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: My understanding is that 1139:
```
p._fp32_shard = p.data
```

Will invoke SsdFlatParam's torch_dispatch function, returning a tensor. If we save a ptr to this tensor (p._fp32_shard), then this tensor will never be freed by the GC. Even if we call p.to_file(), the refcount to the tensor will still be 1 because of p._fp32_shard will still be pointing to it.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: Moved around the 2 comments about pinned memory in my next commit so hopefully it's a bit easier to understand.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: In order to make _free_ssd_offload idempotent, we need to make sure that if this function is called when the SsdFlatParam is already offloaded to disk, no error occurs. 

Without `permit_when_tensor_none` set to True, calling p.to_file() when the SsdFlatParam is already offloaded to disk results in an error.

fairscale/experimental/nn/ssd_offload.py
anj-s: My feeling is that we should add the classes once we need it. I know it will be in a followup PR but would be preferable to add it then. We are not using SsdParameter are we? I thought we are only using SsdFlatParameter.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Fixed a corner case of FSDP init order and losing one of the flags [#880]
+- FSDP: Adding basic training support for SSD Offload, it now only supports flattened parameters. It remains an experimental feature.
```
anj-s, line 12: Can you add the PR number as well?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1153,21 +1129,23 @@ def _init_param_attributes(self, p: Parameter) -> None:
             return
 
         # A single shard of the parameters in full precision.
+        # TODO: PJ - I believe this will cause memory leakage with ssd
```
anj-s, line 1132: The convention for TODOs are using "TODO(username): ...." Can we reformat this to follow that convention? might be good to also add context on why you think this will cause a memory leak.

Shall we add this memory leak investigation to our task tracker as well? (maybe you already did but just checking)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -109,9 +110,9 @@ class OffloadConfig:
     """Class for specifying all arguments related to offloading parameters."""
 
     # Offload type: currently only supports: "ssd_offload"
-    offload_type: str = None
+    offload_type: Optional[str] = None
     # Path to the directory for storing parameters offloaded to disk.
-    ssd_filepath_dir: str = None
+    dir: Optional[str] = None
```
anj-s, line 115: This is a breaking change for existing users. Should we raise an exception if we see `ssd_filepath_dir` or maybe mention this in the Changelog entry?

anj-s (approved):
Thank you for the PR. Looks good! I would remove the classes from ssd_offload that we are not using currently.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
another-pjohnson: I will note this in the changelog.

PR #888, opened 2021-12-15
Release automation

anupambhatnagar:
## What does this PR do?
This PR adds a new github workflow which allows us to push releases with a click of a button. 

### Added
- `release.yml` - a new workflow called for publishing fairscale releases. **It requires a manual dispatch and the allowed input values are "major", "minor" and "patch"**
- `version.json` - this file will contain the fairscale version
- `release_utils.py` - contains some util functions used by the workflow

### Removed
- `build_wheels.yml` - a workflow we are not using 
- `manual-workflow.yml` - a mock workflow to see how manual dispatch works. no longer required

Minor edits in **setup.py**, **conf.py**, **__init__.py**

I created a mock repo to test the workflow and created a mock package that was pushed to my PyPI account. Successful runs for each of the different inputs can be seen [here](https://github.com/anupambhatnagar/package-deployer/actions/workflows/release.yml). 

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review


fairscale/__init__.py
```diff
@@ -3,10 +3,6 @@
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Please update the doc version in docs/source/conf.py as well.
-__version_tuple__ = (0, 4, 3)
-__version__ = ".".join([str(x) for x in __version_tuple__])
```
tmarkstrum, line 8: I think __version__ here is to provide __version__ attribute for developers to use at run time. 
https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

tmarkstrum (commented):
I like put version to a specific doc file. I think we should keep __version__ attribute at the same time in case developer need this attribute at run time.

fairscale/__init__.py
tmarkstrum: I propose assign a value to __version__  similar way as the set up file.

anupambhatnagar: good catch. fixed it.

anupambhatnagar: We need to update the value in the `__version_tuple__`. The cleanest way to do it is to move it into a separate file `version.py` containing only the `__version_tuple__` variable and update it when needed.

tmarkstrum (approved):
lgtm

fairscale/__init__.py
```diff
@@ -3,16 +3,13 @@
...
-
 ################################################################################
 # Import most common subpackages
 ################################################################################
 
 from typing import List
 
-from . import nn
```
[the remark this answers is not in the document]
anupambhatnagar: no, tests fails if this is removed.

PR #889, opened 2021-12-17
checkpoint code

sshleifer:


tmarkstrum:
Are all the unit tests passing?

anj-s:
> Are all the unit tests passing?

They won't pass since an underlying assumption that we are gather params in full precision has been set to False. Working on adding a flag so that we can toggle this feature as needed.

anj-s:
@sshleifer Shall we close this PR? I don't think this was meant to be checked in AFAIK. I will need to make this change and add tests before merging it to main since we do want to move away from using this branch.

PR #897, opened 2021-12-24
[FSDP] Enable FSDP reduce scatter overlap

tmarkstrum:
## What does this PR do?
Fixes the issue that all_gather and reduce_scatter NCCL operations do not overlap with other computation streams in FSDP backward propagation.
The FSDP profile before the fix

The FSDP profile after the fix.

The overlap is enabled by sending the reduce_scatter operations in a separate stream with a different process group. To disable the overlap, set the FSDP API parameter process_group_reduce_scatter to ProcessGroupName.all_gather. 
With this feature enabled, we see 20% throughput improvement on training a 13 B parameter transformer model. and 8.5% throughput improvement on training a 369.7M parameter model.
## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


fairscale/utils/parallel.py
```diff
@@ -58,7 +60,14 @@ def enable_pytorch_sync_bn(module: torch.nn.Module) -> None:
             layer._specify_ddp_gpu_num(1)  # type: ignore
 
 
-def get_process_group_cached(ranks: Optional[Sequence[int]] = None) -> ProcessGroup:
+class ProcessGroupName(str, Enum):
+    main = "main"
```
anj-s, line 64: nit: default

fairscale/utils/parallel.py
```diff
@@ -89,28 +102,30 @@ def get_process_group_cached(ranks: Optional[Sequence[int]] = None) -> ProcessGr
...
-        default_pg = dist.group.WORLD
-        if type(default_pg) == object:
-            # For PyTorch 1.6 and 1.7, dist.group.WORLD is an object, not a world process group, like that in 1.8 and 1.9.
-            default_pg = dist.new_group()
-        cache[None] = default_pg
-        cache[frozenset(list(range(dist.get_world_size())))] = default_pg
+
+        main_pg = dist.new_group(ranks=ranks)
+        cache[None] = main_pg
+        cache[(ProcessGroupName.main, None)] = main_pg
```
anj-s, line 119: curious but do you know why do we need three types of keys in the cache?

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -484,6 +494,8 @@ def _one_step(self, model, group):
         for m in model.modules():
             if isinstance(m, FullyShardedDataParallel):
                 m.process_group = group
+                group_rs = torch.distributed.new_group()
+                m.process_group_reduce_scatter = group_rs
```
anj-s, line 498: inline the previous line with this one.

anj-s:
can you add a unit test for this change that checks the time required for 1 training iteration with and without this flag being true? You'll need a nested FSDP module to really see the benefit.

fairscale/utils/parallel.py
tmarkstrum: When the rank is not specified, it goes to None.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -189,6 +190,14 @@ class FullyShardedDataParallel(nn.Module):
             module to be wrapped with FSDP.
         process_group (Optional):
             process group for sharding
+        process_group_reduce_scatter (Optional):
+            process group for reduce scatter
```
min-xu-ai, line 194: add default value and what does it mean when it is left at the default value.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -189,6 +190,14 @@ class FullyShardedDataParallel(nn.Module):
...
+        enable_reduce_scatter_overlap (bool):
```
min-xu-ai, line 195: can this be merged with previous option so that when the process group is some special value, then this is not enabled? The reason I prefer that is that we can reduce the number of params to FSDP.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -305,6 +316,12 @@ def __init__(
         init_start = time.time()
         super().__init__()
         self.process_group = process_group or get_process_group_cached()
+        self.enable_reduce_scatter_overlap = enable_reduce_scatter_overlap
+        if enable_reduce_scatter_overlap:
+            self.process_group_reduce_scatter = process_group_reduce_scatter or get_process_group_cached(
+                ProcessGroupName.reduce_scatter
+            )
+        self.enable_reduce_scatter_overlap = enable_reduce_scatter_overlap
```
min-xu-ai, line 324: this is a duplicated line with line 319?

I like how the function `get_process_group_cached` is modified with support multiple names. Thanks for doing that!

To reduce the number of parameters, perhaps the new option can take three values:

None: means using the default group
a PG: means using the giving PG
ProcessGroupName.all_gather: means using the same PG as all_gather

This way, we don't need the enabled option, right?

fairscale/utils/parallel.py
min-xu-ai: or name it all_gather_and_compute?

min-xu-ai (commented):
Very nice. Only main comment is around the additional option you are adding here. I think it could be simplified. Once that's addressed, it is good to go from my point of view.

min-xu-ai:
With this #882 is no longer needed?

min-xu-ai:
Two more things:

1. Didn't you have a unit test that tests the overlapping with different streams. Can you add that as well? (OK to be a separate PR)
2. Can you also include more performance numbers in the PR so that we can refer back to this PR in the future?

BTW, you should also update the changelog file. I almost forgot about it. It is really easy to forget.

fairscale/utils/parallel.py
tmarkstrum: I will switch to default because if the parameter for this feature is turned off, all the communications will fall back the this group.

tests/nn/data_parallel/test_fsdp.py
tmarkstrum: what does it mean?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: do you mean that the feature is disabled if the value is PG.all_gather?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: I've been discussing this with @tmarkstrum offline for the same reason you mentioned. I was suggesting that since the 2 PG solution is set to be True by default and will be enabled for everyone, the only reason we would want to turn it off is to rollback. Hence it does not belong in the API list and instead can be an env var for rollback purposes. Using another special value might also be fine if all we want to do is turn off the feature.

fairscale/utils/parallel.py
tmarkstrum: done

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: done. Default to None.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: When it is ParameterGroupName.all_gather, the feature is not enabled.

anj-s:
@tmarkstrum  can we update the PR title to properly reflect this change?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -189,6 +190,12 @@ class FullyShardedDataParallel(nn.Module):
             module to be wrapped with FSDP.
         process_group (Optional):
             process group for sharding
+        process_group_reduce_scatter (Optional):
+            process group for reduce scatter
+            if the process group is specified as ProcessGroupName.all_gather, the reduce_scatter
```
anj-s, line 195: This is not very clear TBH. Can you reword this to say 1)what is the default value and what that means 2) what happens if a user passes in a valid process group and 3) what is the rollback option for this feature.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -305,6 +313,12 @@ def __init__(
         init_start = time.time()
         super().__init__()
         self.process_group = process_group or get_process_group_cached()
+        if process_group_reduce_scatter == ProcessGroupName.all_gather:
```
anj-s, line 316: lets add a comment about what this is doing.

fairscale/utils/parallel.py
```diff
@@ -80,6 +90,10 @@ def get_process_group_cached(ranks: Optional[Sequence[int]] = None) -> ProcessGr
     Extra process groups can also reduce training speed (observed on VISSL models).
 
     Args:
+        name ProcessGroupName:
+            There are two process groups when reduce_scatter overlap is enabled. The "default" process group is the
```
anj-s, line 94: what about the "all_gather" process group in the enum list above?

tests/nn/data_parallel/test_fsdp.py
```diff
@@ -467,6 +475,8 @@ def _test_pickle(self, rank, group, config):
     def _test_multiprocessing(self, rank, group, config):
         mp = torch.multiprocessing.Pool(1)
         dummy_group = DummyProcessGroup(rank=group.rank(), size=group.size())
+        dummy_group_reduce_scatter = DummyProcessGroup(rank=group.rank(), size=group.size())
+        config["process_group_reduce_scatter"] = dummy_group_reduce_scatter
```
anj-s, line 479: can you inline 478 here?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
anj-s: why not pass ProcessGroupName.default in case we want to turn off the feature?

tmarkstrum: That should work as well.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: done

fairscale/utils/parallel.py
tmarkstrum: it is removed.

tests/nn/data_parallel/test_fsdp.py
tmarkstrum: Done

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -190,6 +191,13 @@ class FullyShardedDataParallel(nn.Module):
             module to be wrapped with FSDP.
         process_group (Optional):
             process group for sharding
+        process_group_reduce_scatter (Optional):
+            process group for reduce scatter
+            it defaults to None. A seperate process group is initialized and assigned to the reduce_scatter operation. And the
```
min-xu-ai, line 196: instead of default to None, why not default to ProcessGroupName.reduce_scatter? That way, the type of this is Union[str, PG] and it not an optional anymore.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -290,6 +298,7 @@ def __init__(
         self,
         module: nn.Module,
         process_group: Optional[ProcessGroup] = None,
+        process_group_reduce_scatter: Optional[Any] = None,
```
min-xu-ai, line 301: Any -> Union[str, PG]

min-xu-ai (approved):
Nice! Only a minor comment.

anj-s:
> can you add a unit test for this change that checks the time required for 1 training iteration with and without this flag being true? You'll need a nested FSDP module to really see the benefit.

Discussed offline with @tmarkstrum. this will be added in a follow up PR.

anj-s (approved):

anj-s:
Great work! excited to this perf improvement :)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: Good idea.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
tmarkstrum: Done.

PR #912, opened 2022-01-15
FSDP: better traceback for dtype assertion

sshleifer:


suchenzang (approved):
Whoa dejavu - I totally thought this was added at some point.  Good work for putting it in! 🚀

sshleifer:
nightly CI failures are spurious

PR #913, opened 2022-01-19
[FSDP] Add FairScale FSDP adoptions logging

zhaojuanmao:
Add logging to track FairScale FSDP adoptions internally

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -321,6 +321,12 @@ class FullyShardedDataParallel(nn.Module):
         offload_config: Optional[OffloadConfig] = None,
         state_dict_on_rank_0_only: bool = False,
     ):
+        try:
+            import torch._C
+            torch._C._log_api_usage_once("fairscale.fsdp")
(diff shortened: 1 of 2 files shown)


anj-s:
Thanks for adding this! Can you share how we can view these metrics?

anj-s (approved):

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -321,6 +321,7 @@ def __init__(
         offload_config: Optional[OffloadConfig] = None,
         state_dict_on_rank_0_only: bool = False,
     ):
+        torch._C._log_api_usage_once("fairscale.fsdp")
```
min-xu-ai, line 324: Check torch version before calling this?

zhaojuanmao: Thank you @min-xu-ai , I will do try-except and only logging if _C module is available

zhaojuanmao:
> Thanks for adding this! Can you share how we can view these metrics?

Sure! we logged the metrics to internal scuba table, we will share the results once logging is done

Issue #908, opened 2022-01-12
Flaky Tests in CircleCI


anupambhatnagar:
Here is a list of flaky tests that we should fix in our next fix-a-thon.

- test_shared_weight_mevo[optim_state-flat]
- test_regnet[pytorch-flatten-mixed]
- test_shared_weight_mevo[train-none]
- test_shared_weight_mevo[train-nonflat]
- test_fsdp_memory[fsdp_amp_default-ckpt]; fails due to timeout.
- test1[no_flatten-full] - fails due to timeout.

min-xu-ai:
also layertracker test seems to be flaky

tests.experimental.tooling.test_layer_memory_tracker

```
>       assert summary.total_forward_allocations >= summary.total_activation_allocations
E       assert 77056000 >= 77070864
E        +  where 77056000 = LayerwiseMemoryTrackerSummary(max_memory_allocated=104022528, max_memory_cached=134217728, total_activation_allocation... is_forward=True, all_gathered=0, cumul_all_gathered=0, event=TraceForwardEvent(memory_diff=0, memory_activations=0))]).total_forward_allocations
E        +  and   77070864 = LayerwiseMemoryTrackerSummary(max_memory_allocated=104022528, max_memory_cached=134217728, total_activation_allocation... is_forward=True, all_gathered=0, cumul_all_gathered=0, event=TraceForwardEvent(memory_diff=0, memory_activations=0))]).total_activation_allocations

tests/experimental/tooling/test_layer_memory_tracker.py:65: AssertionError
```

anj-s:
Another flaky test for the list:
tests/nn/data_parallel/test_fsdp.py: TestSerialization

min-xu-ai:
Remember to address todo in https://github.com/facebookresearch/fairscale/pull/933/files/86d45de0e14c8273916c3c68db8c297bc3bb59a8#diff-bc534f971a86e11cc30be248c89989c8024c941a855d4716025da461fcd29047

thanks to Anjali for suggesting recording this.

min-xu-ai:
Another one: https://app.circleci.com/pipelines/github/facebookresearch/fairscale/4277/workflows/73e769c1-510d-4c83-bb25-0eae414f7897/jobs/47992

I will disable it.

PR #923, opened 2022-02-01
[Fix][FSDP] Don't remove post backward hooks for multiple backward fix

### LINKED ISSUE #918: FSDP fails with multiple forward and then multiple backward calls
I have a use case where I need to do multiple forward calls and then backwards call instead of usual pattern but FSDP throws following error when calling multiple backwards:

```
Asserting FSDP instance is: FullyShardedDataParallel(
  world_size=2, flatten_parameters=True, mixed_precision=False,
  (_fsdp_wrapped_module_test): FlattenParamsWrapper(
    (_fpw_module): FFN(
      (fc1): Linear(in_features=10, out_features=10, bias=True)
      (fc2): Linear(in_features=10, out_features=10, bias=True)
      (relu): ReLU()
    )
  )
)
ERROR: expected to be in states [<TrainingState.BACKWARD_POST: 4>] but current state is TrainingState.BACKWARD_PRE
```

Code to repro error: 
```
import torch
import torch.nn as nn
import argparse
from fairscale.nn.data_parallel import FullyShardedDataParallel
from fairscale.nn import enable_wrap
from fairscale.nn import wrap


class FFN(nn.Module):
    def __init__(self):
        super().__init__()
        self.fc1 = nn.Linear(10, 10)
        self.fc2 = nn.Linear(10, 10)
        self.relu = nn.ReLU()
    
    def forward(self, x):
        return self.fc2(self.relu(self.fc1(x)))


def main(i, args):
    torch.manual_seed(0)
    torch.cuda.manual_seed(0)
    print(i)
    torch.cuda.set_device(i)
    torch.distributed.init_process_group(
        backend='nccl',
        init_method='tcp://localhost:22345',
        world_size=args.world_size,
        rank=i,
    )
    ffn = FFN()

    with enable_wrap(
        wrapper_cls=FullyShardedDataParallel
    ):
        model = wrap(
            ffn,
            process_group=torch.distributed.new_group(),
            flatten_parameters=True,
            compute_dtype=torch.float16
        ).half()

    model = model.train()

    x1 = torch.rand((10, 10)).cuda().half()
    out1 = model(x1)
    loss1 = out1.sum()
    
    x2 = torch.rand((10, 10)).cuda().half()
    out2 = model(x2)
    loss2 = out2.sum()
    
    with model.no_sync():
        loss1.backward()
        loss2.backward()

if __name__ == '__main__':
    parser.add_argument('--world-size', '--ws', type=int)
    args = parser.parse_args()
    torch.multiprocessing.spawn(
        fn=main, 
        args=(args, ),
        nprocs=min(
            torch.cuda.device_count(),
            args.world_size,
        ),
        join=True,
    )
```

ngoyal2707:
fixes #918 

I am quite confident, that we dont need to remove backward hooks even after finalizing. They will be automatically removed if the leaf variables go out of context and cuda autograd graph cleans up.

Mots tests were succeeding, apart from one related to cpu offload locally, will debug that

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1543,8 +1543,6 @@ class FullyShardedDataParallel(nn.Module):
             return  # don't register grad hooks if grad isn't enabled
         for p in self.params:
             if p.requires_grad:
-                if hasattr(p, "_shard_bwd_hook"):
-                    continue
                 # Register a hook on the first call, empirically, autograd
                 # fires it at the end for this param, which makes sense.
                 p_tmp = p.expand_as(p)  # Get a grad_fn on p_tmp.
@@ -1751,8 +1749,6 @@ class FullyShardedDataParallel(nn.Module):
                     continue
                 if hasattr(p, "_shard_bwd_hook"):
                     p_assert(len(p._shard_bwd_hook) == 2, f"WFPB: incorrect hook num: {len(p._shard_bwd_hook)}")
-                    p._shard_bwd_hook[1].remove()
-                    delattr(p, "_shard_bwd_hook")
 
                 # Leave the gradient accumulation state as-is if not synchronizing this pass. This ensures p.grad
                 # remains the unsharded gradient accumulated from prior no-sync passes, and p._saved_grad_shard


ngoyal2707:
cc: @myleott

min-xu-ai:
I remember if we don't remove it bad things will happen. :-) Exactly which bad things is the good question. I suspect that some backward hooks will fire excessively or some memory will leak. But if we can be confident that this is indeed save to do, it would be really nice! Less code is always better!

ngoyal2707:
> I remember if we don't remove it bad things will happen. :-) Exactly which bad things is the good question. I suspect that some backward hooks will fire excessively or some memory will leak. But if we can be confident that this is indeed save to do, it would be really nice! Less code is always better!

hmm is there any way to test "some backward hooks will fire excessively or some memory will leak"? or any doc or any past issues about this part? or who's the best POC about this? Myle?

min-xu-ai:
> hmm is there any way to test "some backward hooks will fire excessively or some memory will leak"? or any doc or any past issues about this part? or who's the best POC about this? Myle?

I am the POC. :-) I remember touch this part of the code last time. I don't remember anyone else touched it after I did.

I wish I remember the details of potential issues with removing it. I think the best way for now is to use the unit tests. After unit tests are good if I can't thinking of any potential issues that are not covered by the unit test, we can certainly go ahead remove these line.

ngoyal2707:
Updated PR, local tests are still running but I am hoping now they will all pass 🤞

ngoyal2707:
@min-xu-ai looks like all but MeVo tests are passing, which should be unrelated to this change

min-xu-ai:
> @min-xu-ai looks like all but MeVo tests are passing, which should be unrelated to this change

hmm, are you sure? I clicked on this first one, it seems the serialization test failed and then it timed out:

https://app.circleci.com/pipelines/github/facebookresearch/fairscale/4061/workflows/c33d045c-fbe1-4364-8bc7-7a44335cb7d3/jobs/44795

ngoyal2707:
> > @min-xu-ai looks like all but MeVo tests are passing, which should be unrelated to this change
> 
> hmm, are you sure? I clicked on this first one, it seems the serialization test failed and then it timed out:
> 
> https://app.circleci.com/pipelines/github/facebookresearch/fairscale/4061/workflows/c33d045c-fbe1-4364-8bc7-7a44335cb7d3/jobs/44795

that has to be flaky tests 🤷‍♂️ cause all are passing locally and the change has no relation at all to serialization

ngoyal2707:
```
(fairseq-20211123-py39-meg-v2.6) namangoyal@learnfair7614:~/src/fairscale$ pytest tests/nn/data_parallel/test_fsdp.py::TestSerialization
=========================================================================================================================================================================== test session starts ============================================================================================================================================================================
platform linux -- Python 3.9.7, pytest-5.4.1, py-1.11.0, pluggy-0.13.1 -- /private/home/namangoyal/.conda/envs/fairseq-20211123-py39-meg-v2.6/bin/python
cachedir: .pytest_cache
rootdir: /private/home/namangoyal/src/fairscale, inifile: setup.cfg
plugins: hydra-core-1.0.7, timeout-1.4.2, cov-2.10.0
collected 8 items

tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_multiprocessing__False_False_ PASSED                                                                                                                                                                                                                                                                    [ 12%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_multiprocessing__False_True_ PASSED                                                                                                                                                                                                                                                                     [ 25%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_multiprocessing__True_False_ PASSED                                                                                                                                                                                                                                                                     [ 37%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_multiprocessing__True_True_ PASSED                                                                                                                                                                                                                                                                      [ 50%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_pickle__False_False_ PASSED                                                                                                                                                                                                                                                                             [ 62%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_pickle__False_True_ PASSED                                                                                                                                                                                                                                                                              [ 75%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_pickle__True_False_ PASSED                                                                                                                                                                                                                                                                              [ 87%]
tests/nn/data_parallel/test_fsdp.py::TestSerialization::test_pickle__True_True_ PASSED                                                                                                                                                                                                                                                                               [100%]

======================================================================================================================================================================= 8 passed in 77.47s (0:01:17) =======================================================================================================================================================================
```

min-xu-ai:
We do have flaky tests. We try to track them here: https://github.com/facebookresearch/fairscale/issues/908

However, I just briefly looked at this error:
image

I have actually never seen it before. But I could be wrong. Let me trigger some re-runs on CI.

min-xu-ai:
@zhaojuanmao, can you take a look at this diff as well? Do you think we could remove the code that removes the hooks?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1543,8 +1543,6 @@ def _register_post_backward_hooks(self) -> None:
             return  # don't register grad hooks if grad isn't enabled
         for p in self.params:
             if p.requires_grad:
-                if hasattr(p, "_shard_bwd_hook"):
```
anj-s, line 1546: Do the comments above need to updated because it seems like we don't need to remove the hook at the end of the BW?

anj-s:
@ngoyal2707  Can you share the use case that this supports? I don't think I understood the context from the PR title.

ngoyal2707:
> @ngoyal2707 Can you share the use case that this supports? I don't think I understood the context from the PR title.

@anj-s its a fix for this: https://github.com/facebookresearch/fairscale/issues/918

anj-s:
> > @ngoyal2707 Can you share the use case that this supports? I don't think I understood the context from the PR title.
> 
> @anj-s its a fix for this: #918

Can you add a unit test for this using the smaller repro you have?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
ngoyal2707: ohh yeah definitely, will do that

ngoyal2707:
> > > @ngoyal2707 Can you share the use case that this supports? I don't think I understood the context from the PR title.
> > 
> > 
> > @anj-s its a fix for this: #918
> 
> Can you add a unit test for this using the smaller repro you have? Great that you figured out the solution!

yes, definitely, will add unit test as well

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
zhaojuanmao: if the two lines are removed, will the hooks fire multiple times for multiple forwards cases (e.g. multiple activation checkpointing)?

zhaojuanmao: even for single activation checkpointing case like Checkpoint(FSDP(module)), if there is forward recomputation in the backward pass, the hooks will be registered twice and fired twice unexpectedly?

min-xu-ai:
This is very strange but I have triggered the re-run twice and both times + the first time failed with exact same set of errors. This could be a CI infra issue. But it doesn't seem to be due to flakiness at this point. :-(

ngoyal2707:
> ime failed with exact same set of errors. This could be a CI infra issue. But it doesn't seem to be due to flaki

hmm thats weird, doesn't reproduces on locally, so unfortunately no idea. how to fix it.

min-xu-ai:
> Is there a way to trigger the serialization unit test in the main branch? This can help to identify if the failure is due to the CI infra or the code.

good point. let me trigger a rerun on main

min-xu-ai:
OK. It seems that there is some infra issue with CI. The main branch is failing with timeouts as well

image

anj-s:
@ngoyal2707 Friendly ping on this PR! Do we still want to check in this fix?

min-xu-ai:
Thanks for this PR! I have taken it over in https://github.com/facebookresearch/fairscale/pull/1079, with the new test from you as well. Will close this and merge that one in once all tests are done.

Issue #791, opened 2021-09-10
Allreduce modules that are ignored by auto_wrap


ilmarkov:
## ❓ Questions and Help
`auto_wrap`  seems to ignore modules that have less than 1e8 elements by default - the default parameter of `default_auto_wrap_policy` (that would be nice to be mentioned in the documentation).

The question is what happens with the ignored modules in the case of FSDP. If they are not wrapped by FSDP, do they follow the usual DataParallel, i.e. they are all_reduced every step?

If those modules are not all_reduced, could you suggest how to use FSDP for the wrapped modules and `torch.nn.parallel.DistributedDataParallel` for the others?

fairscale: 0.4.0

min-xu-ai:
Oh, sorry for the confusion. The non-wrap module should be wrapped by an outer layer of FSDP in the end. No, DDP is not needed and it is not advised to mix DDP and FSDP.

I.e. with pseudo code:

```
model = Model()
with enable_wrap:
    model = auto_wrap(model)
model = FSDP(model)
```
Maybe auto_wrap really should be renamed to "auto_wrap_inner_layers" or something like that. cc: @zhaojuanmao

anj-s:
@min-xu-ai  why not wrap the entire model as well when we return from auto_wrap if the outer model's params meet the conditions? Are there use cases where we don't want to do this?

min-xu-ai:
Good idea. Maybe it was intended but somehow not doing it due to a bug? Perhaps a boolean argument for this would be nice.

anj-s:
Maybe we can repurpose this issue to track this effort or do you think we should open a new issue? This seems like a nice onboarding issue for someone to take on :)

min-xu-ai:
Of course, I am fine either way. Feel free to do organize it as you see fit.

ilmarkov:
@min-xu-ai Thank you for the answer. It is still unclear for me why user can't wrap the entire model with FSDP to have all child modules synchronized with the same API as `torch.nn.parallel.DistributedDataParallel`. Are there cases when user wraps not the entire model but only large parts of them without need to synchronize the others?

Could you also explain the difference between wrapping a module with FSDP and wrapping the module's children with FSDP? The documentation only says that there is performance difference but it's unclear how the framework behaves in these cases.

min-xu-ai:
> these

[...]

PR #930, opened 2022-02-09
[fix] Add option to wrap root module in auto_wrap

ruanslv:
## What does this PR do?
Fixes # [791](https://github.com/facebookresearch/fairscale/issues/791).

The issue is that the root/outer module sometimes is not wrapped, depending on whether there are enough params left after children are covered by recursive wrapping: https://github.com/facebookresearch/fairscale/blob/main/fairscale/nn/wrap/auto_wrap.py#L299

So I've added a new input param to `default_auto_wrap_policy` to tell us if the current module is the root, and a new flag that allows us to wrap the root regardless of number of unwrapped params (flag is enabled by default).

Adjusted test-cases to show more clearly behavior with and without the new flag when there are not enough params left for root module.

### DIFF
--- a/fairscale/nn/wrap/auto_wrap.py
+++ b/fairscale/nn/wrap/auto_wrap.py
@@ -13,18 +13,20 @@ def default_auto_wrap_policy(
     module: nn.Module,
     recurse: bool,
     unwrapped_params: int,
+    module_is_root: bool,
     # These are customizable for this default policy function.
     min_num_params: int = int(1e8),
     force_leaf_modules: Optional[Set[Type[nn.Module]]] = None,
     exclude_wrap_modules: Optional[Set[Type[nn.Module]]] = None,
+    skip_params_check_for_root: bool = True,
 ) -> bool:
     """Default policy function for :func:`auto_wrap`.
 
        Return if a module should be wrapped during :func:`auto_wrap`.
 
-       The first three parameters are used by :func:`auto_wrap`. If
+       The first four parameters are used by :func:`auto_wrap`. If
        you write a custom version of this policy function, your version
-       needs to at least accept the first three parameters and free
+       needs to at least accept the first four parameters and free
        to do whatever you want in the function.
 
     Args:
@@ -37,6 +39,8 @@ def default_auto_wrap_policy(
            on whether we should wrap the said module.
        unwrapped_params (int):
            The number of parameters yet to be wrapped in this module.
+       module_is_root (bool):
+           Indicates if current module is the root.
 
        min_num_params (int):
            Customizable policy input. It controls the size threshold
@@ -45,6 +49,9 @@ def default_auto_wrap_policy(
            keep as leaves, i.e., their children will never be wrapped.
        exclude_wrap_modules (Set[Type[nn.Module]]):
            Customizable set of module types to be excluded in wrapping.
+       skip_params_check_for_root (bool):
+           If module_is_root is True, then this includes the root in
+           wrapping regardless of their number of unwrapped params.
     """
     force_leaf_modules = (
         default_auto_wrap_policy.FORCE_LEAF_MODULES  # type: ignore
@@ -63,7 +70,9 @@ def default_auto_wrap_policy(
         return is_large and not isinstance(module, tuple(force_leaf_modules))
     else:
         # If we are not recursing, determine if we should wrap.
-        return is_large and not isinstance(module, tuple(exclude_wrap_modules))
+        return ((module_is_root and skip_params_check_for_root) or is_large) and not isinstance(
+            module, tuple(exclude_wrap_modules)
+        )
 
 
 # Set those defaults to the default_auto_wrap_policy function. Make them easy to be imported.
@@ -75,6 +84,7 @@ def config_auto_wrap_policy(
     module: nn.Module,
     recurse: bool,
     unwrapped_params: int,
+    module_is_root: bool,
 ) -> bool:
     """Config based policy function for :func:`auto_wrap`.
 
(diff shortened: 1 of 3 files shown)


tests/nn/wrap/test_wrap.py
```diff
@@ -62,6 +62,43 @@ def test_auto_wrap(self):
...
 
+    def test_auto_wrap_all_modules(self):
+        """
+        Setting a low min_num_params so that all modules are wrapped.
+        """
+        with enable_wrap(wrapper_cls=FSDP, process_group=self.process_group, flatten_parameters=False):
+            sequential = nn.Sequential(
+                nn.Linear(5, 5), nn.Linear(5, 5), nn.Sequential(nn.Linear(5, 5), nn.Linear(5, 5))
+            )
+            my_auto_wrap_policy = functools.partial(default_auto_wrap_policy, min_num_params=5)
```
min-xu-ai, line 73: Maybe you have considered this option, but what if we add an option to the default_auto_wrap_policy for this behavior? Adding this option to auto_wrap would affect all policy functions that users may write for auto_wrap? Does that make policy functions harder to write because of this option?

Again, thank you so much for the PR and awesome test cases. My general feeling is that if we can limit options in auto_wrap and try to let the policy function control wrapping, that'd be better. What do you think @ruanslv ?

min-xu-ai (approved):
This is so cool! Thanks for taking the time to implement this approach! We can merge once all tests are green.

min-xu-ai:
I triggered a rerun of the tests since the failures seems to be due to a single flaky test.

pyproject.toml
```diff
@@ -27,4 +27,4 @@ use_parentheses = true
 skip_glob = ["build/*", "stubs/*"]
 # Don't split "import" and "from".
 force_sort_within_sections = true
-known_third_party = ["benchmark_dataset", "datasets", "golden_configs", "models", "numpy", "parameterized", "pytest", "recommonmark", "setuptools", "sklearn", "torch", "torchtext", "torchvision", "utils"]
+known_third_party = ["benchmark_dataset", "datasets", "distutils", "golden_configs", "models", "numpy", "parameterized", "pytest", "recommonmark", "setuptools", "sklearn", "torch", "torchtext", "torchvision", "utils"]
```
min-xu-ai, line 30: I think you need to undo this change or the precommit check fails.

ruanslv: https://github.com/ruanslv/fairscale/blob/auto_wrap_fix/CONTRIBUTING.md#pre-commit-recommended

These changes came from the pre-commit checks, I'm not managing to do a commit without them. Should I just bypass them? Any idea how to do that?

min-xu-ai: Yeah. Need to run precommit uninstall to bypass it for now.

anj-s: You can also remove the change, run `git commit -m "" --no-verify` to enable pushing to the branch without running a local pre-commit check. There is some mismatch between the local and remove pre-commit checks that i am looking into.

fairscale/nn/wrap/auto_wrap.py
```diff
@@ -13,18 +13,20 @@ def default_auto_wrap_policy(
     module: nn.Module,
     recurse: bool,
     unwrapped_params: int,
+    module_is_root: bool,
     # These are customizable for this default policy function.
     min_num_params: int = int(1e8),
     force_leaf_modules: Optional[Set[Type[nn.Module]]] = None,
     exclude_wrap_modules: Optional[Set[Type[nn.Module]]] = None,
+    skip_params_check_for_root: bool = True,
```
anj-s, line 21: Won't this change behavior is users are currently using this policy? Wondering if users will end up doing a double wrap without knowing we've already wrapped the root module.

min-xu-ai: good catch. It is worth checking. If so, perhaps default it the False? Also, I can make a separate change where we disable double wrapping in FSDP's init function.

anj-s: Makes sense. Let's make the default behavior as  False for now and once we have the PR that disables double wrapping in FSDP's init function, we can make the default behavior as being True. I was thinking if there might be any repercussions for nested wrapping etc. but I think not wrapping if the module has already been wrapped is a fair check in any condition.

min-xu-ai:
I have a PR #933 that will improve the test stability. Maybe you can merge with main again once that PR is merged?

pyproject.toml
ruanslv: Thanks for the command pointer. I've removed the changes to this file.

anj-s:
@ruanslv Does this need to merge from main before being submitting? Maybe you already did this since I don't see a conflict  above. One last thing is to add an entry to the  [CHANGELOG](https://github.com/facebookresearch/fairscale/blob/main/CHANGELOG.md). Thanks!

ruanslv:
@anj-s I think I'm missing "write access"? Sending screenshot of what I'm seeing attached, I don't see the option to merge. There are also some test failures that seem unrelated to my change, not sure whether they are blocking the merge. Sorry, not super familiar with github UI.
Screenshot 2022-02-15 at 21 16 41

anj-s:
> @anj-s I think I'm missing "write access"? Sending screenshot of what I'm seeing attached, I don't see the option to merge. There are also some test failures that seem unrelated to my change, not sure whether they are blocking the merge. Sorry, not super familiar with github UI. Screenshot 2022-02-15 at 21 16 41


Ah I meant merge from main branch so that there are no conflicts but I don't see any. I am going merge the PR now myself and figure out why you may not have access.

ruanslv:
Oh right, yes I don't think there were any conflicts. Merge worked fine:
```
$ git pull
Already up to date.
ruanrms-mbp:fairscale ruanrms$ git merge auto_wrap_fix
Updating 67bf5bf..8ff6fab
Fast-forward
 fairscale/nn/wrap/auto_wrap.py | 35 +++++++++++++++++++++++++++--------
 tests/nn/wrap/test_wrap.py     | 30 +++++++++++++++++++++++-------
 2 files changed, 50 insertions(+), 15 deletions(-)
```

Thanks for merging the PR. Here are the changelog changes: https://github.com/facebookresearch/fairscale/pull/935

PR #939, opened 2022-02-24
Adding ELI5 video to Fairscale

dmitryvinn:
## Summary
Here at Meta Open Source, we've been creating short-form videos to explain Meta OSS projects in the simplest terms possible called ELI5s. We have also been working on React Labs series that we would like to add to the site. We started embedding these videos into project pages as you can see in [Docusaurus home page](https://docusaurus.io/), [Recoil](https://recoiljs.org/), [Litho](https://fblitho.com/), [Jest](https://jestjs.io/), [React Native](https://reactnative.dev/),[Hydra](https://hydra.cc/), and over 20 other projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ FairScale was designed with the following values in mind:
 
 * **Performance** - FairScale APIs provide the best performance in terms of scaling and efficiency.
 
+## Watch Introductory Video
+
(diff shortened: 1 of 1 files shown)


min-xu-ai (approved):
super nice! Thanks!

PR #964, opened 2022-03-23
[FSDP] ssd_offload: Support pickling/unpickling of SsdTensorHandle based tensors

another-pjohnson:
## What does this PR do?
Resolves what should be one of the last remaining blockers from making ssd_offload a fully supported feature. I verified that raw SsdTensorHandles (and SsdParameter/SsdFlatParameters) can be torch.saved and torch.loaded successfully (using unit tests in test_ssd_offload.py). In addition, I verified that creating a checkpoint from an FSDP wrapped model using torch.save(model.state_dict()) and then loading it via model.load_state_dict() works correctly (using a unit test in test_fsdp_offload.py).

Do to some intricacies from the flatten params wrapper, the unpickle/pickle methods added not not actually used due to the way state_dict calls summon_full_params. But they are still needed for serializing arbitrary objects containing SsdTensorHandle based tensors.

## Before submitting

- [ X] Did you have fun?
- [X ] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/main/CONTRIBUTING.md)?
  - [X ] N/A
- [ X] Did you make sure to update the docs?
- [X ] Did you update the [changelog](https://github.com/facebookresearch/fairscale/blob/main/CHANGELOG.md)? (if needed)

## PR review

### DIFF
--- a/fairscale/experimental/nn/ssd_offload.py
+++ b/fairscale/experimental/nn/ssd_offload.py
@@ -9,10 +9,13 @@ from enum import Enum, auto
 from functools import reduce
 import io
 import os
-from typing import Any, Iterator, List, Optional, Sequence, Tuple, Type
+import pickle
+from types import TracebackType
+from typing import IO, Any, BinaryIO, Iterator, List, Optional, Sequence, Tuple, Type, Union
 
 import numpy as np
 import torch
+from torch.serialization import DEFAULT_PROTOCOL as DEFAULT_PROTOCOL
 
 try:
     from torch.utils._pytree import tree_map
@@ -21,7 +24,7 @@ except ImportError:
     pass
 
 
-DEFAULT_CHUNK_SIZE = 1024 * 1024
+DEFAULT_CHUNK_SIZE = 2048 * 2048
 
 
 def _get_num_chunks(input_tensor: torch.Tensor, chunk_size_bytes: int = DEFAULT_CHUNK_SIZE) -> int:
@@ -89,8 +92,12 @@ class SsdTensorHandle(torch.Tensor):
     data into the tensor that is an attribute of the SsdTensorHandle object or write the tensor to file. At any
     point in time the Tensor may be in memory or on disk.
 
+    Class Variables:
+        override_directory_path: This variable is used by CheckpointPathContextManager to modify the path to any
+            SsdTensorHandles that are saved to a checkpoint via pickling (e.g. torch.save)
+
     Args:
-        shape Tuple[int, ...]: Shape of the tensor that is represented by the handle.
+        shape torch.Size: Shape of the tensor that is represented by the handle.
         dtype: torch.dtype: Dtype of the tensor that is represented by the handle.
         requires_grad: bool: Property of the tensor that is represeneted by the handle.
 
@@ -98,14 +105,18 @@ class SsdTensorHandle(torch.Tensor):
         A SSDTensorHandle object representing a Tensor.
     """
 
+    override_directory_path: Optional[str] = None
+
     @staticmethod
     def __new__(
-        cls: SsdTensorHandle, shape: Tuple[int, ...], dtype: torch.dtype, requires_grad: bool = False
+        cls: Type[SsdTensorHandle], shape: torch.Size, dtype: torch.dtype, requires_grad: bool = False
     ) -> SsdTensorHandle:
-        r = torch.Tensor._make_wrapper_subclass(cls, shape, dtype=dtype, requires_grad=requires_grad)  # type: ignore
+        r = super(SsdTensorHandle, cls)._make_wrapper_subclass(cls, shape, dtype=dtype, requires_grad=requires_grad)  # type: ignore
         return r
 
-    def __init__(self, shape: Tuple[int, ...], dtype: torch.dtype, requires_grad: bool) -> None:
+    def __init__(self, shape: torch.Size, dtype: torch.dtype, requires_grad: bool) -> None:
+        self._unpickle_f: Optional[Union[BinaryIO, IO[bytes]]] = None
+
         self._shape = shape
         if len(shape) == 0:
             self._numel = 0
@@ -122,20 +133,18 @@ class SsdTensorHandle(torch.Tensor):
 
     @classmethod
     def from_file(
-        cls, shape: Tuple[int, ...], dtype: torch.dtype, filename: str, requires_grad: bool = False
+        cls, shape: torch.Size, dtype: torch.dtype, filename: str, offset: int = 0, requires_grad: bool = False
     ) -> SsdTensorHandle:
         """Returns a new SsdTensorHandle from a file."""
         handle = cls(shape=shape, dtype=dtype, requires_grad=requires_grad)
-        handle.filename = filename
-        handle.storage_state = StorageState.ON_DISK
+        handle.point_to_file(filename, offset=offset)
         return handle
 
     @classmethod
     def from_tensor(cls: Type[SsdTensorHandle], tensor: torch.Tensor) -> SsdTensorHandle:
         """Returns a new SsdTensorHandle from a tensor."""
         handle = cls(shape=tensor.shape, dtype=tensor.dtype, requires_grad=tensor.requires_grad)
-        handle.tensor = tensor
-        handle.storage_state = StorageState.ON_CPU
+        handle.point_to_tensor(tensor)
         return handle
 
     def is_available(self) -> bool:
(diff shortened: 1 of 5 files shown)


fairscale/experimental/nn/ssd_offload.py
```diff
@@ -21,7 +24,8 @@
     pass
 
 
-DEFAULT_CHUNK_SIZE = 1024 * 1024
+# DEFAULT_CHUNK_SIZE = 1024 * 1024
```
anj-s, line 27: remove?

another-pjohnson: Yes I will remove that in the next update to this PR

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -9,10 +9,13 @@
...
 import io
 import os
-from typing import Any, Iterator, List, Optional, Sequence, Tuple, Type
+import pickle
+from types import TracebackType
+from typing import IO, Any, BinaryIO, Iterator, List, Optional, Sequence, Tuple, Type, Union
 
 import numpy as np
 import torch
+from torch.serialization import DEFAULT_PROTOCOL as DEFAULT_PROTOCOL
```
anj-s, line 18: do we need the "as" since DEFAULT_PROTOCOL is the same? maybe i am missing something about Python?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -90,22 +94,26 @@ class SsdTensorHandle(torch.Tensor):
...
 
+    override_directory_path: Optional[str] = None
+
     @staticmethod
     def __new__(
-        cls: SsdTensorHandle, shape: Tuple[int, ...], dtype: torch.dtype, requires_grad: bool = False
-    ) -> SsdTensorHandle:
-        r = torch.Tensor._make_wrapper_subclass(cls, shape, dtype=dtype, requires_grad=requires_grad)  # type: ignore
+        cls: Type[SsdTensorHandle], shape: torch.Size, dtype: torch.dtype, requires_grad: bool = False
+    ) -> "SsdTensorHandle":
```
anj-s, line 110: why do we need quotes here for specifying the return type?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -90,22 +94,26 @@ class SsdTensorHandle(torch.Tensor):
...
-        shape Tuple[int, ...]: Shape of the tensor that is represented by the handle.
+        shape torch.Size: Shape of the tensor that is represented by the handle.
         dtype: torch.dtype: Dtype of the tensor that is represented by the handle.
         requires_grad: bool: Property of the tensor that is represeneted by the handle.
 
     Returns:
         A SSDTensorHandle object representing a Tensor.
     """
 
+    override_directory_path: Optional[str] = None
```
anj-s, line 105: can you add a comment about why we need this class variable?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -237,61 +245,129 @@ def unwrap(e: Any) -> torch.Tensor:
...
+        result._unpickle_f = io.open(result.filename, "wb")
+        return result
+
+    def __reduce_ex__(self, proto: int) -> Tuple[Any, Any, Any, Any]:
+        byte_iter = None
+        filename = self.filename
+        if self.override_directory_path is not None:
+            head, tail = os.path.split(self.filename)
+            filename = os.path.join(self.override_directory_path, tail)
+            print(f"overridding path: {os.path.join(head, tail)} to {filename}")
```
anj-s, line 263: should this be a logging debug statement instead of print?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -237,61 +245,129 @@ def unwrap(e: Any) -> torch.Tensor:
...
+            byte_iter,
+        )
+
+    def append(self, item: bytes) -> None:
+        assert self._unpickle_f
+        self._unpickle_f.write(item)
+
+    def extend(self, items: List[bytes]) -> None:
+        for i in items:
+            self.append(i)
```
anj-s, line 284: why does extend not need to write to the file?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -237,61 +245,129 @@ def unwrap(e: Any) -> torch.Tensor:
...
+
+
+class CheckpointPathContextManager:
```
anj-s, line 287: Can you add a doc string and comment about why we need this? I see it only used in tests but maybe its useful more widely?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -237,61 +245,129 @@ def unwrap(e: Any) -> torch.Tensor:
...
+
+
+class SsdParameter(SsdTensorHandle, torch.nn.Parameter):
+    @classmethod
+    def from_tensor(cls: Type[SsdParameter], tensor: SsdTensorHandle) -> SsdParameter:  # type: ignore
+        r = cls(tensor.shape, tensor.dtype, tensor.requires_grad)
+        r.point_to_tensor(tensor)
+        return r
 
-class SsdFlatParameter(torch.nn.Parameter, SsdTensorHandle):
```
anj-s, line 241: why can't we still use this inheritance?

fairscale/experimental/nn/ssd_offload.py
```diff
@@ -237,61 +245,129 @@ def unwrap(e: Any) -> torch.Tensor:
...
+        cls: Type[SsdParameter], shape: torch.Size, dtype: torch.dtype, requires_grad: bool = True
+    ) -> SsdParameter:
+        r = super(SsdParameter, cls).__new__(cls, shape, dtype=dtype, requires_grad=requires_grad)
+        return r  # type: ignore
+
+    def __init__(self, shape: torch.Size, dtype: torch.dtype, requires_grad: bool = True) -> None:
+        super(SsdParameter, self).__init__(shape, dtype, requires_grad)
+
+
+class SsdFlatParameter(SsdParameter):
```
anj-s, line 335: just wondering if we can always use SsdFlatParameter since it is a generic version of the FlatParameter. Thats probably a TODO for the future.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: I remember that mypy had some issues with the function signature of this class before. Some things have changed (including mypy version that I developed this code on, and it doesn't seem to be having problems with the return type not being inclosed in quotes. I will remove them in the next version of this PR.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: Done in next version of PR.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: The "as" allows us to use "DEFAULT_PROTOCOL" instead of "torch.serialization.DEFAULT_PROTOCOL" later in the file.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: I will remove the print. That was left in accidentally when I was implementing the feature.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: extend calls append, which writes to self._unpickle_f.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: Done in next version.

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: I see SsdFlatParameter eventually living inside fairscale/nn/misc/flatten_params_wrapper.py where SsdTensorHandle and SsdParameter would live inside fairscale/experimental/nn/ssd_offload.py (which would be moved somehwere once we promote it to a non-experimental feature).

fairscale/experimental/nn/ssd_offload.py
another-pjohnson: I'm not sure, right now SsdFlatParameters must have their file_params (filename and offset). In addition the SsdTensorHandle hook will automatically write back the tensor to disk whenever there is a modification. I don't think we want that default behavior when FlatParameters are stored inside of GPU/CPU memory.

anj-s (approved):

PR #972, opened 2022-04-11
fix for high gpu reserved memory

ngoyal2707:
Before this change:

Training 175b on 1024 A100s:
```
"ppl": "4217.39", "wps": "105123", "ups": "0.05", "wpb": "2.09715e+06", "bsz": "1024", "num_updates": "26", "lr": "1.09091e-06", "gnorm": "40.827", "clip": "100", "train_wall": "20", "max_cuda_gb_allocated": "13.4", "max_cuda_gb_reserved": "75.4", "current_cuda_gb_allocated": "1.9", "current_cuda_gb_reserved": "75.4", "cuda_gb_free": "65.9"
```

After the change:

```
"ppl": "4217.39", "wps": "107317", "ups": "0.05", "wpb": "2.09715e+06", "bsz": "1024", "num_updates": "26", "lr": "1.09091e-06", "gnorm": "40.827", "clip": "100", "train_wall": "20", "max_cuda_gb_allocated": "13.8", "max_cuda_gb_reserved": "22.1", "current_cuda_gb_allocated": "1.9", "current_cuda_gb_reserved": "22.1", "cuda_gb_free": "65.5"
```

No idea if the approach is perfect or not, but for above use case, I am getting slightly better WPS, much better cuda gb reserved (which I can almost explain now, where its being used) and ppl matches.

plus I can train much much larger models now.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1156,6 +1156,8 @@ class FullyShardedDataParallel(nn.Module):
         self._is_root: Optional[bool] = None
         self._streams: Dict[str, torch.cuda.Stream] = {}
         self._reducer: Optional[ReduceScatterBucketer] = None
+        self._fsdp_forward_ordering: List[nn.Module] = []
+        self._my_fsdp_instance_idx: Optional[int] = None
         for p in self.params:
             if hasattr(p, "_fp32_shard"):
                 del p._fp32_shard  # reset _init_param_attributes
@@ -1327,6 +1329,7 @@ class FullyShardedDataParallel(nn.Module):
                 m.no_broadcast_optim_state = m.no_broadcast_optim_state or (
                     (m.world_size == 1) and (m.world_size < self.world_size) and (m.process_group != self.process_group)
                 )
+                m._fsdp_forward_ordering = self._fsdp_forward_ordering
 
     def _setup_streams(self) -> None:
         """Create streams to overlap data transfer and computation."""
@@ -1386,6 +1389,10 @@ class FullyShardedDataParallel(nn.Module):
         if self._is_root and self.mixed_precision:
             args, kwargs = cast_floats_to_right_precision(True, True, *args, **kwargs)
 
+        if self not in self._fsdp_forward_ordering:
+            self._my_fsdp_instance_idx = len(self._fsdp_forward_ordering)
+            self._fsdp_forward_ordering.append(self)
+
         # If enabled, convert the input to FP32 if we are in full precision.
         # no_grad is not used because the input might be for a non-root instance,
         # which mean autograd needs to go through the conversion.
@@ -1395,6 +1402,14 @@ class FullyShardedDataParallel(nn.Module):
         # All-gather full parameters. This will also transfer FP32 parameters to
         # ``self.compute_dtype`` (e.g., FP16 if *mixed_precision* is ``True``).
         self._rebuild_full_params()
+
+        if (
+            self._fsdp_forward_ordering is not None
+            and self._my_fsdp_instance_idx is not None and self._my_fsdp_instance_idx < len(self._fsdp_forward_ordering) - 1
+        ):
+            self._fsdp_forward_ordering[self._my_fsdp_instance_idx + 1]._rebuild_full_params(
+                wait_for_all_gather=False
+            )
 
         # Register backward hooks to reshard params and reduce-scatter grads.
         # These need to be re-registered every forward pass.
@@ -1484,6 +1499,12 @@ class FullyShardedDataParallel(nn.Module):
             # overhead.
             if self.reshard_after_forward:
                 self._rebuild_full_params()
+                if (
+                    self.reshard_after_forward
+                    and self._fsdp_forward_ordering is not None
+                    and self._my_fsdp_instance_idx is not None and self._my_fsdp_instance_idx > 0
+                ):
+                    self._fsdp_forward_ordering[self._my_fsdp_instance_idx - 1]._rebuild_full_params(wait_for_all_gather=False)
             else:
                 self._use_full_params()
 
@@ -1847,7 +1868,7 @@ class FullyShardedDataParallel(nn.Module):
                     self._output_pre_backward_hook_registered.clear()
 
     @torch.no_grad()
-    def _rebuild_full_params(self, force_full_precision: bool = False) -> Optional[List[Tuple[torch.Tensor, bool]]]:
+    def _rebuild_full_params(self, force_full_precision: bool = False, wait_for_all_gather = True) -> Optional[List[Tuple[torch.Tensor, bool]]]:
         """
         Gather all shards of params.
 
@@ -1916,6 +1937,8 @@ class FullyShardedDataParallel(nn.Module):
 
         # Early exit if we already have full params and don't need full precision.
         if self.has_full_params and not force_full_precision:
+            if wait_for_all_gather:
+                torch.cuda.current_stream().wait_stream(self._streams["all_gather"])
             for p in self.params:
                 update_p_data()
             return output_tensors
@@ -1978,8 +2001,8 @@ class FullyShardedDataParallel(nn.Module):
 
                     if self.move_params_to_cpu and (self.params[0].dtype == self.compute_dtype):
                         self._free_fp16_param_shard([p])
-
-        torch.cuda.current_stream().wait_stream(self._streams["all_gather"])
+        if wait_for_all_gather:
+            torch.cuda.current_stream().wait_stream(self._streams["all_gather"])
         return output_tensors
 
     @torch.no_grad()
@@ -2047,6 +2070,7 @@ class FullyShardedDataParallel(nn.Module):
             # Storage object and unshard it in-place. For now, just resize
             # the Storage to 0 to save memory.
             free_storage_(p._full_param_padded)
+            torch.cuda.current_stream().synchronize()
 
     def local_metadata_dict(self) -> Dict[str, Any]:
         """


anj-s:
Thanks Naman! Looks great and we should definitely incorporate. Couple of questions:
1. This still uses the prefetch mechanism right? Can you explain the approach a bit.
2. Curious why the cuda_gb_free does not increase even when the reserved value decreases so drastically? Maybe we are logging something else but I would expect the value of free to be higher.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1978,8 +2001,8 @@ def update_p_data(custom_output_tensor: Optional[torch.Tensor] = None) -> None:
 
                     if self.move_params_to_cpu and (self.params[0].dtype == self.compute_dtype):
                         self._free_fp16_param_shard([p])
-
-        torch.cuda.current_stream().wait_stream(self._streams["all_gather"])
```
anj-s, line 1982: curious about this change? Why do we condition on the boolean?

ngoyal2707:
1) Yes it is using prefetch approach, there could maybe a way to achieve above without prefetch, but I thought hard about it and couldn't see it 

2) that is just by how we log it: https://github.com/fairinternal/fairseq-py/blob/main/fairseq/trainer.py#L1060

min-xu-ai:
cc @zhaojuanmao

zhaojuanmao:
@ngoyal2707 are you disabling multiple process groups for reduce_scattter and all_gather in the backward, instead, using the backward prefetch approach here? 

PyTorch version of FSDP is doing the backward prefetch right now. This will better avoid over prefetching more layers in the backward pass using multiple process group approach

zhaojuanmao:
but will it be conflict with multi process group approach? will the backward prefetch be a config?

stephenroller:
Tried this on AWS. Find it doubles roughly WPS for my particular workload. (1700 -> 3200)

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -2047,6 +2070,7 @@ def _free_full_params(self, params: Optional[List[Parameter]] = None) -> None:
             # Storage object and unshard it in-place. For now, just resize
             # the Storage to 0 to save memory.
             free_storage_(p._full_param_padded)
+            torch.cuda.current_stream().synchronize()
```
zhaojuanmao, line 2073: @ngoyal2707 I guess this helped releasing GPU reserved memory...

For @stephenroller models, were your training slowed down by high reserved memory and slow CUDA cache allocator? if so, maybe this change helped a lot...

zhaojuanmao: but I'm wondering whether this will result in regression for other models which are not suffered from high reserved memory and slow CUDA cache allocator issue

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1396,6 +1403,14 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
         self._rebuild_full_params()
 
+        if (
+            self._fsdp_forward_ordering is not None
+            and self._my_fsdp_instance_idx is not None and self._my_fsdp_instance_idx < len(self._fsdp_forward_ordering) - 1
+        ):
+            self._fsdp_forward_ordering[self._my_fsdp_instance_idx + 1]._rebuild_full_params(
+                wait_for_all_gather=False
+            )
+
```
zhaojuanmao, line 1413: nice to add this forward prefetching!

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
stephenroller: Yeah my jobs are fine tuned of a large pre trained model. We consistently saw on both 40gb and 80gb nodes that the memory reserved was always 99%

stephenroller:
Let's push this through

anj-s:
> Let's push this through

To push this through we need an owner who can:
1. Evaluate if we should move to the prefetch solution or if it will work with the existing 2 PG solution.
2. Fix unit tests and make sure it works for FSDP in general and not just this use case.

zhaojuanmao:
@anj-s and @stephenroller, I suspected the speedup is possibly mostly from this single sentence change "torch.cuda.current_stream().synchronize()" for @stephenroller's use case, possibly it is good to verify that. e..g, @stephenroller, just use FairScale master branch + this change "torch.cuda.current_stream().synchronize()" to your use case, and see whether there is a speed up. 

But this change is better to be optional, not be default, otherwise other FSDP use cases that are not suffered from high GPU reserved memory issue possibly will get regression.

anj-s:
> @anj-s and @stephenroller, I suspected the speedup is possibly mostly from this single sentence change "torch.cuda.current_stream().synchronize()" for @stephenroller's use case, possibly it is good to verify that. e..g, @stephenroller, just use FairScale master branch + this change "torch.cuda.current_stream().synchronize()" to your use case, and see whether there is a speed up.
> 
> But this change is better to be optional, not be default, otherwise other FSDP use cases that are not suffered from high GPU reserved memory issue possibly will get regression.

Can you explain a little bit about why you think there might be a regression?

stephenroller:
I'm not entirely sure about this. We also have large scale, not latency bound pretraining, and this change increased WPS there--when ideal reserved memory is only 25% of the node. What's clear in both cases was we were thrashing the allocator aggressively.

This could maybe be a regression for places where the full state and activations fit in memory, and the network is able to fetch faster than the compute can utilize the state. But that case is better addressed with zero2 anyway, no?

stephenroller:
@ngoyal2707 is clearly the owner and should push this through.

zhaojuanmao:
"What's clear in both cases was we were thrashing the allocator aggressively."   

==> that mostly means "torch.cuda.current_stream().synchronize()" helps here, because CUDA allocation is a async operation, it could try to allocate a large storage before completely free full parameters here. So it is trashing.  Adding this sync for your use case basically makes sure full parameters are freed before allocating. This helped reducing CUDA allocator trash. 


For the use cases that are not suffering from slow cudaAllocator, this blocking operation will wait for all pending async operations and reduce parallelizations, and thus could result in regression, no?

zhaojuanmao:
So for unblocking this effort, suggestion here is: 
1. root cause whether prefetching or cuda sync really helps here first
2. if add cuda sync, make it optional

ngoyal2707:
So I agree with everyone here. I should be the owner but honestly I still don't know if this is the best approach here and I want opinion of people who understands cuda and pytorch caching allocator better

The core reason of the issue is due to unbounded queue of launching kernels combined with multiple cuda streams.

One very simple pytorch example of this is following:

```
import torch
torch.cuda._sleep(100000000000)
for i in range(100):

    x=torch.zeros(4 * 1024 * 1024, device="cuda")
    del x
print(torch.cuda.memory_allocated())
print(torch.cuda.memory_reserved())
# print(torch.cuda.memory_snapshot())
```

above code outputs at end:
```
0
16777216
```
That means, we only reserve 16MB of memory for the 4M fp32 tensor cause its only been used by single stream, so pytorch caching allocator can just assign every time the same memory block as operations within stream are sequential.


comparing above to following:
```
import torch
s1 = torch.cuda.Stream()

torch.cuda._sleep(100000000000)

with torch.cuda.stream(s1):
     torch.cuda._sleep(100000000000)

for i in range(100):
     x=torch.zeros(4 * 1024 * 1024, device="cuda")
     x.record_stream(s1)
     torch.cuda.current_stream().wait_stream(s1)
     del x

print(torch.cuda.memory_allocated())
print(torch.cuda.memory_reserved())
# print(torch.cuda.memory_snapshot())
```
This outputs:
```
0
1677721600
```

i.e. pytorch reserves 16MB tensor for every new allocation cause each memory block is being used by both current stream and `s1` stream.


So in FSDP case, `s1` is the `all_gather` stream, which is of course running slower than CPU thread. What I don't know is, if there is any better solution than just making CPU thread wait. We can change how it waits i.e. instead of just waiting on `torch.cuda.current_stream().synchronize()`, we can change to always launch new `all_gather` once previous one is done.


So I think there possibly can be some regression, specially to very small models cause GPU can be idle while CPU thread is launching new kernels.

qingliu1111:
Want to add a data point about the performance improvement on AWS brought by this PR : ) 
I'm running fine-tuning flow for 175B model on AWS with 16 nodes (128GPU). 
* Without this PR: wps is ~ 2817.1 
* With this PR:      wps is ~ 8817.1 (almost tripled the speed!)

More context: With similar training setup we were getting wps ~12000 on Azure. For 175B model fine-tuning, this PR helps us close a lot the speed gap between Azure and AWS.

zhaojuanmao:
adding @mrshenli who is looking into pluggable cudaCacheAllocator, and hopefully can resolve this issue in general

min-xu-ai:
We probably have 2 options to move forward:

1. put this under a boolean flag
2. only make the cpu thread wait if the free memory drops under some threshold (but perhaps with a timed wait, not cuda.synchronize)

Either way, the tests need to pass. I am proposing to Ruan and Crutcher (both are new team members) to see if they can help out or not.

glample:
Hi, also wanted to report that I observe 2x training speed when using this branch over the main one (800 -> 1600 words per second)

min-xu-ai:
> Hi, also wanted to report that I observe a 2x training speed increase when using this branch over the main one (800 -> 1600 words per second)

Thanks! What's the model size and # of GPUs? Also, are you using fairseq?

glample:
Hi,

Model size is 10B parameters (48 layers, dim 4096). We are not using fairseq but a small custom implementation.

stephenroller:
I feel hiding this under a Boolean flag could be user hostile at least if defaulting to false. It can be frustrating when libraries don't do the right thing by without setting many flags.

ruanslv:
Happy to take a look at this (if no one else is actively working on it). It does seem important to find a way to merge this PR

zhaojuanmao:
> So for unblocking this effort, suggestion here is:
> 
> 1. root cause whether prefetching or cuda sync really helps here first
> 2. if add cuda sync, make it optional

@ngoyal2707 @stephenroller @min-xu-ai , just want to put another data point here from PTD FSDP experiments, we narrowed down that it is 'torch.cuda.current_stream().synchronize() right after _free_storage() ' actually helped the reserved memory reduction here. Prefetching codes do not help too much here.   

e.g. adding this line 'torch.cuda.current_stream().synchronize() right after _free_storage()' reduced high reserved GPU memory from 95% to 47% of total memory in our HF T5 8B model experiments. 

I remembered Fair Protein team also mentioned they tried this sync after _free_storage() and get reserved memory reduced, but did not understand why...

So I'm planning to push this change in the PTD FSDP version. Before making this change in PTD FSDP version, we are planning to do more experiments and see whether it has regression for other models in different setups. Meanwhile we will seek feedback from core team about implication of this change (will add you in the discussions). 

Thank you!

ruanslv:
@zhaojuanmao could you loop me into any discussions too pls? I have been looking at this PR from the FairScale side.

Unless forward() for the PTD version is different from FairScale, I would expect that simply adding synchronize() without the forward prefetching, while indeed helping to reduce reserved memory, would cause regressions in throughput.

The reason I see is that synchronize alone basically disables any kind of parallelism on forward() within a given shard, no? The code would start doing sequentially:

All-gather -> forward computation -> cuda synchronize() -> all-gather -> forward computation -> synchronize() -> ...

While currently, we kick off the next all-gather as soon as the first one is done.

Similar logic applies to backward pass I'm guessing. Any harm in keeping the prefetching?

I've been debating across two options:
(1) Keeping current prefetching, which is like a best-effort to kick off the next all-gather but may still cause some regressions if we get the order wrong and/or still under-schedule future all-gathers.
(2) Min's proposal of doing some kind of wait (timed wait or synchronize()) based on amount of reserved memory. This is more generic but finding appropriate memory thresholds could be difficult.

Tks!

zhaojuanmao:
@ruanslv no harm in keeping prefetching:) and your understanding of forward prefetching benefit sounds correct! PTD version already had forward prefetching and backward prefetching actually.... regarding PTD forward prefetching, we only allow to prefetch one all gather ahead. will add you into the discussions, thank you!

regarding your two options: 
1. seems it is important to make scheduling aligned with module execution order, possibly PTD non-recursive wrapping can help here...
2. sounds a good idea! it is better than blindly apply synchronization point, right? @mrshenli

ruanslv:
Oh I see. If PTD already has prefetching then indeed nothing else to do apart from adding synchronize() calls in order to mimic this PR's behaviour. Thanks for the clarification!

> seems it is important to make scheduling aligned with module execution order, possibly PTD non-recursive wrapping can help here...

@zhaojuanmao I would love to learn in more details the prefetching done on PTD side. I don't know what you meant by "PTD non-recursive wrapping".

A few potential flaws with the implementation in this PR, which I'm not sure are real concerns or not and, if they are, how PTD handles them:

1/ We assume the module execution order is the same across different (forward/backward) passes, given we instantiate the module list only during the first forward pass. Perhaps execution order for forward passes is always the same and based on the code, but what about backward pass? Is the computation order deterministic across different passes?

2/ Even within a single forward/backward pass, does the execution order for the backward pass actually match the one for the forward pass? Given here we are simply prefetching the previous module using the forward pass ordering.

> sounds a good idea! it is better than blindly apply synchronization point, right? 

Tricky part is finding a general memory threshold we are comfortable with and that is general enough. I'm trying some local experiments here to get a better intuition behind this, will share if I find a more concrete proposal.

ruanslv:
Second attempt at this in https://github.com/facebookresearch/fairscale/pull/1052

PR #994, opened 2022-05-26
[minor] Remove false tuples

crutcher:
## What does this PR do?
This PR removes a number of false tuples, the construction `(X)` does not create a singleton tuple in python, it just resolves to `X`; the presence of false tuples in the code is misleading to readers, and can lead to difficult debugging situations. True singleton tuples can be constructed via `(X,)`; but I've just removed these because (apparently) the tests work with the singleton values.

## Before submitting

- [X] Did you have fun?
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/main/CONTRIBUTING.md)?
  - [X] N/A
  - [X] N/A
  - [X] N/A
  - [X] N/A

## PR review

### DIFF
--- a/tests/experimental/nn/test_ssd_offload.py
+++ b/tests/experimental/nn/test_ssd_offload.py
@@ -33,7 +33,7 @@ def test_write_read():
     _init()
 
     with tempfile.NamedTemporaryFile() as f:
-        ref_tensor = torch.rand((128), dtype=torch.float32)
+        ref_tensor = torch.rand(128, dtype=torch.float32)
         test_tensor = torch.zeros_like(ref_tensor)
         assert not torch.equal(ref_tensor, test_tensor)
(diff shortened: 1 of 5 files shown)


min-xu-ai (approved):
nice! if this check basically is 100% correct all the time, then I don't see why not it can be in pre-commit.

crutcher:
It's messy to scan for pre-commit without looking at the AST, so a quick-and-dirty scan is probably not what we want here; The regex `(\d+)` occurs too often in other contexts, particularly in documentation.

I am interested in bringing AST scanning (via LibCST) to bear on facebookresearch code; but that's a more involved project.

I stumbled upon this by accident, did a manual scan with grep, and fixed these up by hand, skipping a lot of false matches.

PR #995, opened 2022-05-26
[minor] .gitignore data/ cached by tests

crutcher:
## What does this PR do?
.gitignore `data/` directory produced by running tests.

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review


.gitignore
```diff
@@ -32,3 +32,6 @@ env.bak/
 venv.bak/
 .vscode/*
 *.DS_Store
+
+# Data generated by tests
+data/
```
min-xu-ai, line 37: interesting. I didn't see this before. Do you happen to know which test generated this dir? This is super generic name too. Perhaps renaming it would be better.

crutcher: it's the standard path of the torchvision.datasets lib,
```
$ ls data/
cifar-10-batches-py  cifar-10-python.tar.gz

$ grep -iR cifar tests

tests/optim/test_layerwise_gradient_scaler.py:        train_ds = torchvision.datasets.CIFAR10(root="./data", train=True, download=True, transform=transform)
```

min-xu-ai: I didn't know that we have a test that download a dataset. I though only benchmark runs do that with data being cached. @anupambhatnagar, can we avoid doing this? This is going to make our test repeatedly downloading and can get blacklisted.

If that is changed, then, we don't really need this gitignore.

anupambhatnagar: The cifar data is only downloaded once. If the directory is found in the root of the project, downloading is skipped. This is the behavior on devfair and on CircleCI. The attached screenshot shows is from my devfair box and in this  case the files had already been downloaded by the `test_vision_model` test in a prior run.

In the case that CircleCI uses a new container to run the tests, then the dataset will need to be downloaded on that container. Additionally, given the download rates on in the fair cluster we have never seen the test take over our timeout limit of 70 seconds.

Adding /data to .gitignore is a no-op in my opinion.

 Screen Shot 2022-05-26 at 5 18 25 PM

min-xu-ai: We have seen that a web server black list our test machine IPs if they download too frequently. They may treat this as a DDos attack. E.g. If we have N dev branches and 15 test machines for each branch we are testing, the downloading can easily be large, like 15*N or more. It is not timeout of download speed.

We can wait until it happens (if ever) to fix this. In that case, I am fine with the gitignore. I'd suggest a name like test_layerwise_gradient_scaler_data rather than data.

anupambhatnagar: Agree on both points :)

crutcher: Since this is a dataroot for common datasets, I'd suggest we use a name that is scoped to that purpose, rather than a particular test that is using the dataset.

Calling this directory `data/` is pretty common practice; but if that is too broad, maybe `cached_datasets/` ?

min-xu-ai: 👍

tests/optim/test_layerwise_gradient_scaler.py
```diff
@@ -19,6 +19,7 @@
 
 from fairscale.optim.layerwise_gradient_scaler import LayerwiseGradientScaler
 from fairscale.utils.testing import skip_a_test_if_in_CI
+from tests.common_paths import DATASET_CACHE_ROOT
```
min-xu-ai, line 22: See my comment from the not-a-pr PR. I suggest we don't import code from `tests` dir since (1) that top level module name is very common and keeping all code under fairscale make things easier to grep etc. How about adding this to fairscale/testing dir?

min-xu-ai (commented):
Thanks for fixing this! What do you think about my suggestion below?

.gitignore
crutcher: Updated, @min-xu-ai please approve?

tests/optim/test_layerwise_gradient_scaler.py
crutcher: this is a deeper question, about what we want to export to library users, vrs the amount of machinery we want to include in the testing stack (and the development package requirements vrs the user package requirements).

as a general pattern, the current tests don't use project-specific testing libraries, so the robustness of those libs, when dropped into `fairscale/testing` is limited to what we wish to expose to experimental users; which has a cooling effect on them.

another path, `fairscale_testlibs/` could exist, to hold this machinery, and distinguish it from `fairscale/` and `tests/`

I would like to add that this pr (make `.gitignore` play nice with current behavior) is growing in scope to (fix common test data paths as a pattern); but I'm happy to finish this.

as a concrete proposal, I suggest `fair_testlibs/` as a path, and if we later extract that to a supplemental dev repo, we don't have to change the module paths

min-xu-ai: Your plan sounds great to me

min-xu-ai (approved):
Lgtm

tests/optim/test_layerwise_gradient_scaler.py
crutcher: Updated, but used `fair_dev/` to cover other potential dev-only machinery.

PR #999, opened 2022-05-30
[discussion] crutcher's codebase notes

crutcher:
This is not a PR, it is a collection of code notes I'm working on while reading through the fairscale codebase, lifted to a draft pr for purposes of collaborative discussion. This will likely become a source of PRs.

Please comment, if you feel like it


fairscale/nn/data_parallel/__init__.py
```diff
@@ -8,4 +8,5 @@
 from .fully_sharded_data_parallel import FullyShardedDataParallel, OffloadConfig, TrainingState, auto_wrap_bn
 from .sharded_ddp import ShardedDataParallel
 
+# FIXME: why the empty __all__ list with explicit imports?
```
min-xu-ai, line 11: oh, yeah, I remember now. I think we debated a bit on whether we want to support:

```
from foo import *
```

I think the `__all__`, only controls that case. See this bullet point in CONTRIBUTING.md to see the rational behind it:
image


With empty __all__, import * won't accidentally import symbols that covers other symbols with the same name. Also, it lower the overhead for devs to maintain 2 list in a consistent way.

fairscale/nn/data_parallel/fsdp_optim_utils.py
```diff
@@ -243,6 +243,7 @@ def build_unflat_state_dict(
     return unflat_optim_state_dict
 
 
+# FIXME: `util/tensor/shape.py` ? lift to pytorch?
```
min-xu-ai, line 246: good idea.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -305,6 +305,8 @@ class FullyShardedDataParallel(nn.Module):
             Default: False
     """
 
+    # FIXME: class annotations?
```
min-xu-ai, line 308: what is "class annotations"?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -358,6 +360,7 @@ def __init__(
                 if not hasattr(process_group_reduce_scatter, "allgather") and hasattr(
                     process_group_reduce_scatter, "rank"
                 ):
+                    # FIXME: common test func for dummy PG.
```
min-xu-ai, line 363: I think the comment on the next line means that this case is caused by the usage of a dummy PG in our unit test, i.e. this case is not a production use case.

fairscale/nn/data_parallel/sharded_ddp.py
```diff
@@ -26,6 +26,7 @@
 from fairscale.utils.params import Workhandle, get_global_rank
 
 
+# FIXME? `util/tensor/grad.py`?
```
min-xu-ai, line 29: The fate of sharded_ddp should be first determined. We should see if we still need it because it provides something unique that FSDP doesn't already have. For OSS, I think we should keep it since it shard optimizer state differently from FSDP. I am less clear to SDP. cc @blefaudeux

fairscale/nn/model_parallel/__init__.py
```diff
@@ -23,4 +23,5 @@
 from .mappings import copy_to_model_parallel_region, gather_from_model_parallel_region
 from .random import get_cuda_rng_tracker, model_parallel_cuda_manual_seed
 
+# FIXME: why block re-export?
```
min-xu-ai, line 26: again, see contributing.md

fairscale/nn/model_parallel/cross_entropy.py
```diff
@@ -27,6 +27,7 @@
 
 
 class _VocabParallelCrossEntropy(torch.autograd.Function):
+    # FIXME: docs
```
min-xu-ai, line 30: good idea. the whole model_parallel dir is an import from nvidia megatron code if I am not mistaken. We again need to determine how to keep it. Maybe we want to keep it to make sure a version of it work seamlessly with fsdp or pipeline, or for other reasons. I don't really know.

fairscale/nn/model_parallel/cross_entropy.py
```diff
@@ -102,5 +103,6 @@ def backward(ctx, grad_output):  # type: ignore
 
 
 def vocab_parallel_cross_entropy(vocab_parallel_logits: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+    # FIXME: why does this exist?
```
min-xu-ai, line 106: beats me

fairscale/nn/model_parallel/initialize.py
```diff
@@ -28,6 +28,8 @@
 
 from .utils import ensure_divisibility
 
+# FIXME: this assumes exactly one distributed process.
```
min-xu-ai, line 31: I suggest reading the megatron paper first if you haven't already. I do think it is an approach that does model parallelism in a way that's layer-specific, not a general approach like ZERO 1/2/3.

min-xu-ai: i.e. for different layers, vocab, embedding, etc., different model parallel versions will need to be written specifically.

fairscale/utils/containers.py
```diff
@@ -12,6 +12,9 @@
 """Useful functions to deal with tensor types with other python container types."""
 
 
+# FIXME: the semantics and definition of this are more flexible
```
min-xu-ai, line 15: Will have to look at the code new version when you have it to have a concrete opinion on it. on the surface sounds nice.

fairscale/utils/containers.py
```diff
@@ -24,6 +27,17 @@ def _apply(x: Union[torch.Tensor, Dict, List, Tuple, Set]) -> Any:
                 od[key] = _apply(value)
             return od
         elif isinstance(x, PackedSequence):
+            # FIXME: undocumented out-of-bound usage.
```
min-xu-ai, line 30: interesting. I think we need a big picture description of what are the use cases of this class and what's the algorithm does it use. This will help us have full unit test coverage and not over engineering it.

fairscale/utils/golden_testing_data.py
```diff
@@ -6,6 +6,7 @@
 
 """ Golden data used in unit tests. """
 
+# FIXME: why is this in the main lib, rather than the unittests directory?
```
min-xu-ai, line 9: we don't import anything from `tests` and we didn't create a `fairscale.testing` directory. I think we should do that. We only have `fairscale.utils.testing.py`. perhaps we should have a fairscale.testing dir and put all testing related code, like this one, there. what do you think?

fairscale/utils/object.py
```diff
@@ -9,6 +9,7 @@
 import torch
 
 
+# FIXME: both of these mechanisms at _least_ double allocate and double copy.
```
min-xu-ai, line 12: I think this is related to some all_gather call for objects before pytorch supported it. Newer pytorch version might already support it. If all the supported pytorch version has this, we can remove it. We support pytorch LTS, stable, and also test against a nightly build.

fairscale/utils/parallel.py
```diff
@@ -15,6 +15,8 @@
 import torch.nn.functional as F
 
 
+# FIXME: this seems like a candidate for pytorch.
```
min-xu-ai, line 18: see above. since we are separated from pytorch, we develop a bit differently from pytorch team's module.

* for pytorch team's module, they can assume a single version of pytorch, which is the "current" version they are running with.
* for us, we support multiple versions of pytorch (this make user's life a bit easier since they don't have to get new version of pytorch to get new features from us)
* on the other hand, we have a smaller code base and fewer tests, we can potentially develop faster since we don't have to checkin our code into a big codebase and run large number of tests just for our module.

fairscale/utils/parallel.py
```diff
@@ -33,6 +35,11 @@ def validate_process_group(device: torch.device, process_group: ProcessGroup) ->
     the CPU.
     """
     if not hasattr(process_group, "allgather"):
+        # FIXME: consider dummy having an explicit marker or base class,
```
min-xu-ai, line 38: I see. nice idea.

fairscale/utils/parallel.py
```diff
@@ -104,12 +111,19 @@ def get_process_group_cached(
...
         if name == ProcessGroupName.reduce_scatter and "pytest" in sys.modules:
+            # FIXME: consider dummy having an explicit marker or base class,
+            # and having an explicit `is_dummy_process_group(pg)` test,
+            # rather than assuming that malformed pgs are dummy pgs,
+            # and doing the test differently in each of these places.
             return None
         else:
             raise RuntimeError("torch.distributed is not yet initialized but process group is requested.")
 
+    # FIXME: it is _very_ odd to hang this cache off a function attribute,
```
min-xu-ai, line 122: True. This is like a hack. We have different needs in deals with PGs. They impact GPU memory as well as overlapping of compute & communication. They also impact usability and flexibility in terms of letting users control PG or letting us do it. The caching method here is by no means the best possible method.

fairscale/utils/testing_memory.py
```diff
@@ -11,6 +11,8 @@
 import torch
 
 
+# FIXME: write a test for this.
```
min-xu-ai, line 14: Sure. It is a test for a test function though. :-)

This is like "it-works" method for getting the test done. By no means the best. Some times, "perfect is an enemy of good". :-)

fairscale/utils/version.py
```diff
@@ -24,6 +24,7 @@ def torch_version(version: str = torch.__version__) -> Tuple[int, ...]:
 
         # Assuming that we're interested in the second use-case more than the first,
         # return the pre-release or dev numbering
+        # FIXME: log this ... everytime we call `torch_version()`? This is pretty spammy.
```
min-xu-ai, line 27: yeah, totally

tests/utils/test_containers.py
```diff
@@ -99,6 +99,10 @@ def test_split_unpack():
     x = torch.Tensor([1])
     y = torch.Tensor([2])
 
+    tensors, packed_non_tensors = split_non_tensors(x)
```
min-xu-ai, line 102: nice addition to the test

min-xu-ai (commented):
I added things that I know to the best of my knowledge. This is a great way to share info on the code base as you read through it! As you dig deeper I may not have time and knowledge to help as much as you'd like. But I will certainly try my best.

fairscale/nn/checkpoint/__init__.py
```diff
@@ -7,4 +7,5 @@
 
 from .checkpoint_activations import checkpoint_wrapper, is_checkpointing, is_recomputing
 
+# FIXME: what is the goal of hiding the above imports from the export?
```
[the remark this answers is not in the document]
crutcher: `__all__` exists solely to constrain what `from x import *` does.

On the one hand, I'm a big fan of not using `*` imports; on the other, forcing that on callers by voiding out `__all__` is a bit weird, and if that's the choice here, we should have a comment stating that explictly, a policy describing that at the top level, and should enforce this throughout the codebase (with a pre-commit hook)

fairscale/utils/testing_memory.py
crutcher: always write tests for your test functions. you want to _trust_ your test functions, particularly when they do something clever like this :)

tests/utils/test_containers.py
crutcher: i pulled this into a PR: https://github.com/facebookresearch/fairscale/pull/1000

fairscale/nn/checkpoint/__init__.py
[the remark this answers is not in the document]
min-xu-ai: Precommit hook sounds great. Don’t know how to do it properly though. 😅

fairscale/utils/testing_memory.py
min-xu-ai: Very true

min-xu-ai:
clean up non-commiting PR

PR #1001, opened 2022-05-31
wgit functionalities and skeleton, move to subparsers and addition of Repo Class

riohib:
## What does this PR do?
Fixes # Adds many of the functionalities for the WeiGit class initializations, a json file creation, and creation of the `.git` repo within `.wgit` repo among others. Also, adds the skeleton for many of the planned WeiGit methods along with their CLI integration and tests inside the `./tests/experimental/wgit/cli` directory for verification. 

Morever, adds a `Repo` class that identifies a weigit repo with it's path. Finally, all the cli implementation have been changed and now uses `subparsers` instead, with subcommands e.g  `wgit commit -m <message>` and `wgit log -f <filename>`. 

Adds testing in separate directory, and also checks for exceptions while creating `.wgit` directory files. Repo class path  now uses `os.path.realpath` to strip symlinks if present. Also, exit codes have been centralized in `experimental/wgit/utils.py`. Also, discarded inf while loop checking.

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] Yes
  - [ ] N/A

## PR review


experimental/wgit/cli.py
```diff
@@ -5,18 +5,92 @@
...
+
+    if args.command == "commit":
+        weigit_api.WeiGit.commit(args.message)
+
+    if args.command == "checkout":
+        weigit_api.WeiGit.checkout()
+
+    if args.command == "version":
+        print(wgit.__version__)
+        exit(0)
```
min-xu-ai, line 93: this is not needed? by default, returns here will end up have an exit code of 0.

experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,141 @@
...
+            3. Initialize a .git directory within the .wgit using `git init`.
+            4. add a .gitignore within the .wgit directory, so that the git repo within will ignore `sha1_ref_count.json`
+        """
+
+        # Make .wgit directory
+        try:
+            os.mkdir(".wgit")
+        except FileExistsError:
+            sys.stderr.write("WeiGit has been already Initialized \n")
+            sys.exit(0)
```
min-xu-ai, line 30: exit with error code, say 1? feel free to define all exit code in a single place, like an enum.

experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # if no .wgit dir then initialize the following
+        SHA1_store()
+
+        # create sha1_ref_count
+        with open(".wgit/sha1_ref_count.json", "w") as f:
+            print("The sha1_ref_count.json file has been created!")
```
min-xu-ai, line 37: this is for debugging, right?

experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Make the .wgit a git repo, add a gitignore and add SHA1_ref
+        pygit2.init_repository(".wgit/.git", False)
```
min-xu-ai, line 40: try & except around this in case it fails? in general, during error handling, we need to figure out how to recover, like undo things on disk, etc.

experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,141 @@
...
+        with open("./.wgit/.gitignore", "w") as f:
```
min-xu-ai, line 41: there should be a check that this file is not already there?

experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,141 @@
...
+            """
+            is_wgit_in_curr = pathlib.Path(os.path.join(check_dir, ".wgit")).exists()
+            is_refcount_in_wgit = pathlib.Path(os.path.join(check_dir, ".wgit/sha1_ref_count.json")).exists()
+            is_git_in_wgit = pathlib.Path(os.path.join(check_dir, ".wgit/.git")).exists()
+            return is_wgit_in_curr and is_refcount_in_wgit and is_git_in_wgit
+
+        if weigit_repo_exists(self.check_dir):
+            self.repo_path = os.path.join(self.check_dir, ".wgit")
+        else:
+            max_dir_depth = 100
```
min-xu-ai, line 122: add a comment on the need of this? max_dir_depth is used to cache cycles in the symbolic links? Have you tried os.path.realpath() to get the real path for a given path?

tests/experimental/wgit/cli/test_cli.py
```diff
@@ -3,11 +3,64 @@
...
+
+import pathlib
+import shutil
+
 import experimental.wgit.cli as cli
 
 
+def setup_module(module):
+    try:
+        shutil.rmtree(".wgit")
```
min-xu-ai, line 15: can we run the test inside a dir like `wgit_test_dir`, instead of the top level fairscale dir?

min-xu-ai (commented):
Cool! Almost ready, only minor comments below.

experimental/wgit/cli.py
riohib: You're correct. reflecting the changes

experimental/wgit/weigit_api.py
riohib: The `max_dir_depth` is simply so that we do not get into an infinite while loop. The while loop basically starts from the input directory and keeps looping one dir up till it reaches the root dir. Could you suggest a better way to handle the loop or will a comment suffice for this magic number? We can probably error out as well.

In terms of os.path.realpath, are you suggesting parsing the constructor input path with os realpath at the start in the constructor to strip of any sym links?

experimental/wgit/weigit_api.py
riohib: yes, kept it for the skeleton. Probably better to remove it now!

experimental/wgit/weigit_api.py
riohib: Yes, good idea!

experimental/wgit/weigit_api.py
riohib: I think for this one, this code snippet only executes if `.wgit` doesn't exist (if .wgit exists then we error out in line 30). Should still checking be better?

tests/experimental/wgit/cli/test_cli.py
riohib: Could you clarify this a little bit?

min-xu-ai: I assume that you run the test within the current dir during development. That means, say your fariscale is at

/home/name/fairscale

then, the test will create

/home/name/fairscale/.wgit

I am wondering can you make the test run within a dir like /home/name/fairscale/wgit_unittest or something like that.

Am I missing something?

experimental/wgit/weigit_api.py
min-xu-ai: I am wondering when will we get into an infinite loop? Have you run into it? there shouldn't be any cycle in the path, which means it should always go to root. I was hoping that either infinite loop is either not possible or can be dealt with realpath.

experimental/wgit/weigit_api.py
min-xu-ai: True. There could be a rare race condition that after the first check, this file shows up. But you are right, we don't need the check here. Can you please add a comment here?

tests/experimental/wgit/cli/test_cli.py
riohib: Oh got it. Yeah, we can definitely do that.

experimental/wgit/weigit_api.py
riohib: To be honest, we will probably not get into an infinite loop as you imply. We can probably discard it.  Also, since we are going up the directory linearly, I think we can simply get the realpath of the input dir path at the start. Making these changes.

experimental/wgit/weigit_api.py
riohib: Yes adding a comment to reflect it.

riohib: Adding a try and except, in case we face the race condition.

min-xu-ai:
I will take one last look once you are done to approve. Then, please merge.

riohib:
Added testing in separate directory, and also now checks for exceptions while creating `.wgit` directory files. Repo class path now uses `os.path.realpath` to strip symlinks if present. Moreover, exit codes have been centralized in `experimental/wgit/utils.py` and discarded inf while loop checking.

.gitignore
```diff
@@ -12,6 +12,10 @@
 *.egg-info/
 .testmondata
 
+# experimental
```
min-xu-ai, line 15: Improve this comment?

min-xu-ai (approved):
Looks great! Thanks for the patience for addressing my comments!

riohib:
Thanks for pointing these out at as well. Learning a lot of the details, Appreciate it :))

PR #1002, opened 2022-06-06
Addition of wgit add and wgit commit functionalities. Includes refactors and new classes.

riohib:
## What does this PR do?
Fixes # (issue).
1. Implements the `wgit add` and `wgit commit` functionalities. 
2. Redoes the `.wgit/file.pt` (now `.wgit/checkpoint.pt`) to reflect the meta_data to be commited by git (instead of appending). Moreover, the actual `sha1` is no longer the key within this json data. Adds, future option for inclusion of layerwise`sha1s`.
3. Adds reference tracking in `sha1_refs.json`.  
4. Added some error handling to `add` and failure clean up and handling.
5. Added a new `PyGit` class to find and wrap the `.wgit/.git` repo. Handles the `git add` and `git commit` of the metadata file `checkpoint.pt`. PyGit2 is being used for these functionalities.
6. Refactored code. Discarded the `staticmethod` based implementation in favor of a object instance based one, with `Repo`, `PyGit` and `WeiGit` classes.

[Update:]
  1. Combines the functionalities of the WeiGit and Repo class into a single WeiGitRepo class.
  2. Classes are now separated into their own modules.
  3. Moved some functions and staticmethod to utils.
  4. Adds a range of tests for add and commit functionalities of weigit.

[Update]:
 1. Moved wgit within `fairscale/experimental`. Made changes to other codes to support installation.
 2. Added type checking.
 3. Additional checks for some cli functionalities.

[Update]:
1. Changes weigit `cli` slightly to handle errors and some cases. 
2. Correspondingly changes the `weigit_api.WeiGitRepo` class `__init__` method to reflect the `cli` changes. 

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


experimental/wgit/cli.py
```diff
@@ -74,19 +74,24 @@ def main(argv=None):
         weigit = weigit_api.WeiGit()
 
     if args.command == "add":
-        weigit_api.WeiGit.add(args.add)
+        weigit = weigit_api.WeiGit()
```
min-xu-ai, line 77: maybe rename "weigit" variable into "repo" to denote that it is a repo?

min-xu-ai: same for the rest in this file obviously.

experimental/wgit/weigit_api.py
```diff
@@ -112,8 +201,65 @@ class SHA1_store:
...
+            # Try
+            with open(ref_file, "w", encoding="utf-8") as f:
+                json.dump(ref_data, f, ensure_ascii=False, indent=4)
+            # except: clean up TODO: Make the whole add method in try except block cleaning up the SHA1 directory ## TODO
+
+    def read(SHA1):
+        pass
+
+    @staticmethod
+    def get_sha1_hash(file_path):
```
min-xu-ai, line 251: this should be a helper function in the utils.py?

experimental/wgit/weigit_api.py
```diff
@@ -14,92 +16,179 @@
 
 
 class WeiGit:
```
min-xu-ai, line 18: I am thinking that we should combine the WeiGit class and the Repo class. Just use the repo class for everything that WeiGit class is doing. I.e. 

repo = Repo()
repo.add()
repo.status()

etc.

What's the reason to have 2 separate classes?

min-xu-ai (commented):
Added some high level comment. Let's sync offline.

fairscale/experimental/wgit/utils.py
```diff
@@ -0,0 +1,71 @@
...
+            sha1.update(data)
+    return sha1.hexdigest()
+
+
+def weigit_repo_exists(check_dir: Path) -> bool:
+    wgit_exists, sha1_refs, git_exists, gitignore_exists = weigit_repo_status(check_dir)
+    return wgit_exists and sha1_refs and git_exists and gitignore_exists
+
+
+def weigit_repo_status(check_dir: Path) -> tuple:
```
min-xu-ai, line 55: status can be confused with status command, rename to `weight_repo_file_check`?

fairscale/experimental/wgit/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    gitignore_exists = check_dir.joinpath(".wgit/.gitignore").exists()
+    return wgit_exists, sha1_refs, git_exists, gitignore_exists
+
+
+class ExitCode(Enum):
+    CLEAN = 0
+    FILE_EXISTS_ERROR = 1
+    FILE_DOES_NOT_EXIST_ERROR = 2
+
+    ERROR = -1  # unknown errors
```
min-xu-ai, line 71: is this actually used?

fairscale/experimental/wgit/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    returns the state of the weigit repo and checks if all the required files are present.
+    """
+    wgit_exists = check_dir.joinpath(".wgit").exists()
+    sha1_refs = check_dir.joinpath(".wgit/sha1_refs.json").exists()
+    git_exists = check_dir.joinpath(".wgit/.git").exists()
+    gitignore_exists = check_dir.joinpath(".wgit/.gitignore").exists()
+    return wgit_exists, sha1_refs, git_exists, gitignore_exists
+
+
+class ExitCode(Enum):
```
min-xu-ai, line 66: docstring

fairscale/experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,172 @@
...
+
+import json
+import os
+import pathlib
+from pathlib import Path
+import shutil
+import sys
+from typing import Union
+
+import pygit2
```
min-xu-ai, line 14: is this still needed? Shouldn't we be using PyGit class not this directly?

fairscale/experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,172 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
min-xu-ai, line 1: like I mentioned, this file can be renamed to repo.py

fairscale/experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,172 @@
...
+from pathlib import Path
+import shutil
+import sys
+from typing import Union
+
+import pygit2
+
+from fairscale.experimental.wgit.pygit import PyGit
+from fairscale.experimental.wgit.sha1_store import SHA1_store
+from fairscale.experimental.wgit.utils import ExitCode, create_dir, create_file, get_sha1_hash, write_to_file
```
min-xu-ai, line 18: all these can be:

from .pygit import ...
from .sha1_store import ...
from .utils import ...

fairscale/experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class WeiGitRepo:
```
min-xu-ai, line 21: can be simply "Repo" class.

fairscale/experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,172 @@
...
+        if not self._exists() and init:
+            # No weigit repo exists and is being initialized with init=True
+            # Make .wgit directory, create sha1_refs and metadata file
+            create_dir(dir_path=self._wgit_dir, exception_msg="An exception occured: WeiGit already Initialized")
+            create_file(file_path=self._metadata_file)
+            create_file(file_path=self._sha1_ref)
+
+            # Make the .wgit a git repo
+            try:
+                pygit2.init_repository(str(self._wgit_git_path), False)
```
min-xu-ai, line 52: This probably should be part of the PyGit class and should handle self.user and self.email's init.

tests/experimental/wgit/test_api.py
```diff
@@ -0,0 +1,109 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
min-xu-ai, line 1: Thanks for adding those top level (integration) tests!

I think we should also include unit test for each class, like test_repo.py, test_sha1_store.py, test_pygit.py etc. Basically, with unit test, we can test each APIs in separation and make bugs and regression much easier to catch.

fairscale/experimental/wgit/weigit_api.py
```diff
@@ -0,0 +1,172 @@
...
+            # use the sha1_has to create a directory with first2 sha naming convention
+            try:
+                repo_fdir = os.path.join(self.path, "sha1_store", sha1_hash[:2]) + "/"
+                os.makedirs(repo_fdir)
+            except FileExistsError as error:
+                sys.stderr.write(f"An exception occured: {repr(error)}\n")
+                sys.exit(ExitCode.FILE_EXISTS_ERROR)
+            try:
+                # First transfer the file to the internal sha1_store
+                repo_fpath = os.path.join(repo_fdir, sha1_hash[2:])
```
min-xu-ai, line 88: this should probably be part of sha1_store.add() function.

min-xu-ai (commented):
This is getting much better! Thanks for all the re-structuring and tests.

I have mostly smaller and local comments. These may still require good effort to address them all. Thanks for the patience. We want to get the basic structure as good as possible and it will speed up our future development.

Again, we are not increasing any scope of this PR. We just want to limit it the existing goals but doing the existing things in the right way.

fairscale/experimental/wgit/pygit.py
```diff
@@ -0,0 +1,59 @@
...
+        if pygit_parent_p != self.wgit_path:
+            self.exists = False
+            self.repo = None
+            raise pygit2.GitError("No git repo exists within .wgit. Reinitialize weigit!")
+        else:
+            self.exists = True
+            self.repo = pygit2.Repository(str(self.path))
+            self.index = self.repo.index
+            self.name = "user"  # Commit metadata
+            self.email = "user@email.com"
```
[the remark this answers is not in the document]
min-xu-ai: Yeah, I can see this is tricky. I think the init() needs to take those as optional arguments which can be used by test in case the user's config isn't set. For normal cases when user's .gitconfig is already set, the options won't be needed.

fairscale/experimental/wgit/sha1_store.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self) -> None:
+        try:
+            sha1_store_path = Path.cwd().joinpath(".wgit", "sha1_store")
+            if not sha1_store_path.exists():
+                Path.mkdir(sha1_store_path, parents=False, exist_ok=False)
+        except FileExistsError as error:
+            sys.stderr.write(f"An exception occured while creating Sha1_store: {repr(error)}\n")
+            sys.exit(ExitCode.FILE_EXISTS_ERROR)
+
+    def add_ref(self, current_sha1_hash: str) -> None:
```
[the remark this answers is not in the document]
min-xu-ai: Just to make sure we don't have any misunderstanding, the sha1_store.add() function does more than just checksuming, right? It also copy the file or store the data into a destination file inside sha1_store's internal dir so that it can be read back in the future, right?

fairscale/experimental/wgit/utils.py
riohib: The last one? I don't think that is being used now.

fairscale/experimental/wgit/weigit_api.py
riohib: Ah I see, this makes is more reasonable. Need to update the design doc a bit as well to reflect this.

.gitignore
```diff
@@ -13,7 +13,7 @@
 .testmondata
 
 # experimental weigit
-experimental/.gitignore
+fairscale/experimental/wgit/dev
```
min-xu-ai, line 16: comment on what this dev dir is used for?

fairscale/experimental/wgit/__init__.py
```diff
@@ -3,7 +3,7 @@
 # This source code is licensed under the BSD license found in the
 # LICENSE file in the root directory of this source tree.
 
-from .cli import main
-from .weigit_api import WeiGit
+from .repo import Repo
+from .version import __version_tuple__
 
-__version__ = "0.0.1"
+__version__ = ".".join([str(x) for x in __version_tuple__])
```
min-xu-ai, line 9: after this line, perhaps add

__all__ = []

so that `import *` won't import anything by default to avoid polluting the name space. That's a convention we use in fairscale.

fairscale/experimental/wgit/cli.py
```diff
@@ -71,25 +72,30 @@ def main(argv=None):
...
+        repo.commit(args.message)
 
     if args.command == "checkout":
-        weigit_api.WeiGit.checkout()
+        repo = Repo(Path.cwd())
+        repo.checkout(args.checkout)
 
     if args.command == "version":
-        print(wgit.__version__)
+        print(".".join([str(x) for x in version.__version_tuple__]))
```
min-xu-ai, line 98: it seems that you can just use __version__ directly not duplicate the code here?

fairscale/experimental/wgit/pygit.py
```diff
@@ -0,0 +1,100 @@
...
+        parent_path: Path,
+        gitignore: List = list(),
+        name: str = "user",
+        email: str = "user@email.com",
+    ) -> None:
+        """
+        PyGit class to wrap the wgit/.git repo and interact with the git repo.
+
+        Args:
+        parent_path: Has to be the full path of the parent!
```
min-xu-ai, line 24: pls describe the meaning of this arg in addition of the important note here.

also document other args like gitignore, name, email.

maybe no need to have default values for name and email? Let caller always pass them explicitly?

fairscale/experimental/wgit/pygit.py
```diff
@@ -0,0 +1,100 @@
...
+                self.path = self._parent_path.joinpath(".git")
+            else:
+                # if the parent is not a .wgit repo,
+                # then the found-repo is a different git repo. Init new .wgit/.git
+                self._init_wgit_git(gitignore)
+        else:
+            # no git repo found, make .wgit a git repo
+            self._init_wgit_git(gitignore)
+
+    def _init_wgit_git(self, gitignore: List) -> None:
```
min-xu-ai, line 52: document the gitignore arg. you can use "see above" since it is the same as that in the `__init__`.

fairscale/experimental/wgit/pygit.py
```diff
@@ -0,0 +1,100 @@
...
+    def _exists(self) -> bool:
+        """returns True if wgit is a git repository"""
+        return self._parent_path == Path(self.repo.path).parent
+
+    @property
+    def _path(self) -> str:
+        """returns the path of the git repository PyGit is wrapped around"""
+        return self.repo.path
+
+    def status(self) -> None:
```
min-xu-ai, line 98: maybe this should return the status, not print it? Let caller decide whether print or not?

fairscale/experimental/wgit/repo.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class Repo:
+    def __init__(self, parent_dir: Path, init: bool = False) -> None:
+        """Features:
+        1. Create the wgit directory if it does not exist.
+        2. SHA1Store.init()
+        3. Create SHA1 .wgit/sha1_refs.json
+        3. Initialize a .git directory within the .wgit using `git init`.
+        4. add a .gitignore within the .wgit directory, so that the git repo within will ignore `sha1_refs.json`
```
min-xu-ai, line 22: document the args pls.

fairscale/experimental/wgit/repo.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        # If repo does not exist, creates a new wgit repo object with self.repo.path pointing to the path of repo
+        # and notes all the internal files.
+        # else, if repo already exists: create a pygit object from the .wgit/.git.
+        self.wgit_parent = parent_dir
+        self._repo_path: Union[None, Path] = None
+        self._wgit_dir = Path(".wgit")
+        self._metadata_file = Path(".wgit/checkpoint.pt")
+        self._sha1_ref = Path(".wgit/sha1_refs.json")
```
min-xu-ai, line 31: should this be part of sha1 store class? It is for sha1 store for ref counting for delete, right?

fairscale/experimental/wgit/repo.py
```diff
@@ -0,0 +1,131 @@
...
+        3. Initialize a .git directory within the .wgit using `git init`.
+        4. add a .gitignore within the .wgit directory, so that the git repo within will ignore `sha1_refs.json`
+        """
+        # If repo does not exist, creates a new wgit repo object with self.repo.path pointing to the path of repo
+        # and notes all the internal files.
+        # else, if repo already exists: create a pygit object from the .wgit/.git.
+        self.wgit_parent = parent_dir
+        self._repo_path: Union[None, Path] = None
+        self._wgit_dir = Path(".wgit")
+        self._metadata_file = Path(".wgit/checkpoint.pt")
```
min-xu-ai, line 30: I am still confused by the need of this file. I think there shouldn't be a single hardcode named metadata file like this.

Instead, we will have metadata file PER weight file users ask us to track.

Say, user have a file name `cifar.pt`, and he does:

wgit add cifar.pt

then,

inside .wgit/git, we should create/modify:

.wgit/git/cifar.pt

which is a metadata file. Am I missing anything that you are thinking otherwise?

fairscale/experimental/wgit/repo.py
```diff
@@ -0,0 +1,131 @@
...
+        if self._exists():
+            print("wgit checkout: sha1")
+
+    def compression(self) -> None:
+        """Not Implemented: Compression functionalities"""
+        print("Not Implemented!")
+
+    def checkout_by_steps(self) -> None:
+        """Not Implemented: Checkout by steps"""
+        print("Not Implemented!")
```
min-xu-ai, line 105: you could `raise NotImplementedError()`. I remember this is a standard python exception for it.

min-xu-ai (approved):
Great improvement! I have some minor comments and 1 major one.

The major one is about hardcoded checkpoint.pt. See my comment below. I think it needs to be changed to PER user-supplied file name. You can address this before merging or merge this and immediately address this issue since this is a pretty fundamental thing.

riohib:
I think I will fix the major issue you mentioned and then merge this. Thanks a lot for pointing it out.

fairscale/experimental/wgit/repo.py
riohib: I think this makes sense for a more general use case. I think I was operating under the impression that we would be making users track only one checkpoint file per weigit repo. Will go ahead and include these. Thanks!!

riohib:
Final Notes: 
> I think I will fix the major issue you mentioned and then merge this. Thanks a lot for pointing it out. :)

Upon further work, fixing the highlighted major issue needs a cascading range of changes in the sha1 reference tracking that will significantly change the scope of this PR. Hence, I am merging PR and will open a new PR with the new fixes and modifications that deals with all the comments in this PR after the last commit.

PR #1023, opened 2022-06-30
support param groups (fairseq composite optimizer

alexeib:


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1971,6 +2003,9 @@ def update_p_data(custom_output_tensor: Optional[torch.Tensor] = None) -> None:
                             alloc_storage_(p._full_param_padded, size=p_size)
                         output_tensor = p._full_param_padded
 
+                    if output_tensor.dtype != p_data.dtype:
```
min-xu-ai, line 2006: @alexeib, do you have more details on what triggers this bug?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -453,15 +453,47 @@ def __init__(
...
+            # We put the params to three groups. weight, bias and the rest of parameters
+            to_be_flatten_params = []
             non_flatten_params = []
-            param_name_groups = [param_names]
+            del param_name_groups
+            group_to_index = {}
+            param_name_groups = []
+
+            for i, n in enumerate(param_names):
+                pg = getattr(params[i], "param_group", "default")
```
min-xu-ai, line 469: If I understood it correctly, the original module's param get tagged with an attr named "param_group", which has the str type, right? Then, FSDP uses this tag to put params in groups and all untagged ones fall into the default flatten group?

If so, how do user express a need to leave params unflattened? Also, is the tagging process easy and not error-prune for the users / caller code? Can you point to me where is the caller code for this in your branch?

I hope we can explore a few options for this API to find one that is easiest and less error prone yet most flexible for the users. :-) Thanks for this initial idea though. I do think it could be the one we end up using. Just wanted to carefully consider a few alternatives.

min-xu-ai (commented):
Thanks and see a few comments below.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
alexeib: the composite optimizer in fairseq works by groupping params into different groups based on this "param_group" attribute. if its absent, the param gets put into a "default" group.

keeping params unflattened i guess would be accomplished by setting "flatten_parameters" to false, right ?

the tagging process boils down to just setting props on params in model constructor. e.g. see: https://github.com/facebookresearch/fairseq/blob/main/examples/wav2vec/unsupervised/models/wav2vec_u.py#L458

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
alexeib: to be honest i forgot, but i think it had to do with having fp16 params but storing fp32 grads

min-xu-ai: haha, no worries. We will figure it out.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
min-xu-ai: I see. Will read in more details.
 
> keeping params unflattened i guess would be accomplished by setting "flatten_parameters" to false, right ?
I see, it is a combination of the tag and the arg flag. Then, if user wants to completely disable flattening, he/she needs to BOTH remove the tags (p.param_group) and set the arg flag to False? That could be a bit error-prune because I can see some users just flip and flag and forgot to remove the tags. ;-)

PR #1034, opened 2022-07-20
[Feat] dense to sst implementation

riohib:
1. Implementation of dense to sst function.
2. calculating the threshold for both the cases of top-k-element and top-k-percentage (fraction)

## What does this PR do?
Fixes # (issue).

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -4,7 +4,9 @@
 # LICENSE file in the root directory of this source tree.
 
 from enum import Enum
+from typing import Optional
 
+import torch
 from torch import Tensor
 
 
@@ -52,8 +54,25 @@ class SignalSparsity:
             conv.weight.data = 3d_sparser.get_sst_dst_weight(conv.weight.data)
     """
 
-    def __init__(self) -> None:
-        pass
+    def __init__(
+        self,
+        algo: Algo = Algo.FFT,
+        sst_top_k_dim: Optional[int] = -1,
+        sst_top_k_element: Optional[int] = None,
+        sst_top_k_percent: Optional[float] = None,
+        dst_top_k_dim: Optional[int] = -1,
+        dst_top_k_element: Optional[int] = None,
+        dst_top_k_percent: Optional[float] = None,
+    ) -> None:
+
+        self._algo = algo
+        self._sst_top_k_dim = sst_top_k_dim
+        self._sst_top_k_element = sst_top_k_element
+        self._sst_top_k_percent = sst_top_k_percent
+        self._dst_top_k_dim = dst_top_k_dim
+        self._dst_top_k_element = dst_top_k_element
+        self._dst_top_k_percent = dst_top_k_percent
+
         self._validate_conf()
 
     def _validate_conf(self) -> None:
@@ -63,7 +82,13 @@ class SignalSparsity:
 
         this should assert fail if checking fails.
         """
-        pass
+        # assert that both top_k_elements and top_k_percent aren't set for sst and dst
+        assert (
+            self._sst_top_k_element is None or self._sst_top_k_percent is None
+        ), "Both top-k element and top-k percent has been provided as argument"
+        assert (
+            self._dst_top_k_element is None or self._dst_top_k_percent is None
+        ), "Both top-k element and top-k percent has been provided as argument"
 
     def dense_to_sst(self, dense: Tensor) -> Tensor:
         """Get SST from a tensor
@@ -73,7 +98,14 @@ class SignalSparsity:
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
-        pass
+        w_fft = torch.fft.fft(dense)  # type: ignore
+        w_fft_abs = torch.abs(w_fft)  # get absolute FFT values
+
+        # Next, use the normalized real values for getting the topk mask
+        threshold = self.get_threshold(w_fft_abs, self._sst_top_k_element, self._sst_top_k_percent, self._sst_top_k_dim)
+        sps_mask = w_fft_abs > threshold
+        w_sst = w_fft * sps_mask  # but mask the actual complex FFT values topk
+        return w_sst
 
     def dense_sst_to_dst(self, dense: Tensor, sst: Tensor) -> Tensor:
         """From dense and SST to a DST
@@ -111,6 +143,35 @@ class SignalSparsity:
             A dense tensor in real number domain from the SST.
         """
         pass
+
+    def get_threshold(
+        self,
+        in_tensor: Tensor,
+        top_k_element: Optional[int] = None,
+        top_k_percent: Optional[float] = None,
+        dim: int = -1,
+    ) -> Tensor:
+        """Get a mask for a tensor corresponding to a certain sparsity level.
+        Args:
+            in_tensor (Tensor)
+                input torch tensor for which sparse mask is generated.
+            sparsity (float)
+                target sparsity of the tensor for mask generation.
+        """
+        abs_tensor = torch.abs(in_tensor)
+
+        if self._dst_top_k_percent or self._sst_top_k_percent:
+            if top_k_percent == 0.0:  # if sparsity is zero, we want a mask with all 1's
+                threshold = torch.tensor(float("-Inf"), device=in_tensor.device).unsqueeze(dim)
+            else:
+                threshold = torch.quantile(abs_tensor, top_k_percent, dim).unsqueeze(dim)  # type: ignore
+
+        elif self._dst_top_k_element or self._sst_top_k_element:
+            # top-k along only some dimension dim
+            v, _ = torch.topk(in_tensor, top_k_element, dim)
+            min_v, _ = torch.min(v, dim)
+            threshold = min_v.unsqueeze(dim)
+        return threshold
 
     def sst_or_dst_to_mask(self) -> None:
         # we shouldn't need this function since going from SST/DST to mask should be a
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -14,14 +14,12 @@ from torch import Tensor
 def _get_k_for_topk(topk_percent: Optional[float], top_k_element: Optional[int], top_k_total_size: int) -> int:
     """Converts the top_k_percent to top_k_element when top_k_percent is provided
     as the criterion for top-k calculation. When, top_k_element is used as the criterion,
-    simply returns the value for k. Also, ensures k is never 0 to avoid all sparse tensors.
+    simply returns the value for k. Also, ensures k is never 0 to avoid all-zero tensors.
     """
-    if top_k_element is not None and top_k_element > top_k_total_size:
-        raise ValueError("top_k_element for sst or dst is larger than max number of elements along top_k_dim")
-
     if top_k_element is None:
         top_k_element = int(top_k_total_size * topk_percent / 100.0)
-
+    elif top_k_element > top_k_total_size:
+        raise ValueError("top_k_element for sst or dst is larger than max number of elements along top_k_dim")
     # ensure we never have 100% sparsity in tensor and always have 1 surviving element!
     return max(1, top_k_element)
 
@@ -47,25 +45,26 @@ def _scatter_topk_to_sparse_tensor(
             dimension for top-k
 
     Returns:
-        Returns a sparse_tensor with the same shape as the top_k_tensor and to_be_sparsify_tensor,
-        and populated with the values of the to_be_sparsify_tensor at the indices corresponding
-        to the top-k indices of the source tensor.
+        (Tensor)
+            Returns a sparse_tensor with the same shape as the top_k_tensor and to_be_sparsify_tensor,
+            and populated with the values of the to_be_sparsify_tensor at the indices corresponding
+            to the top-k indices of the source tensor.
     """
     assert (
         top_k_tensor.shape == to_be_sparsify_tensor.shape
     ), "top_k_tensor and to_be_sparsify_tensor have different shapes!"
 
     sparse_tensor = torch.zeros_like(to_be_sparsify_tensor)
-    if dim is None and len(top_k_tensor.shape) > 1:
-        orig_shape = sparse_tensor.shape
+    shape = top_k_tensor.shape
+    orig_shape = sparse_tensor.shape
+    if dim is None and len(shape) > 1:
         sparse_tensor = sparse_tensor.reshape(-1)
-        _, idx = top_k_tensor.flatten().topk(k)
-        sparse_tensor[idx] = to_be_sparsify_tensor.flatten()[idx]
-        sparse_tensor = sparse_tensor.reshape(orig_shape)
-    else:
-        _, idx = top_k_tensor.topk(k, dim=dim)
-        # gather the values from the to_be_sparsify_tensor and scatter to a tensor of zeros
-        sparse_tensor = sparse_tensor.scatter_(dim, idx, to_be_sparsify_tensor.gather(dim, idx))
+        to_be_sparsify_tensor = to_be_sparsify_tensor.reshape(-1)
+        top_k_tensor = top_k_tensor.reshape(-1)
+        dim = -1
+
+    _, i = top_k_tensor.topk(k, dim=dim)
+    sparse_tensor = sparse_tensor.scatter(dim, i, to_be_sparsify_tensor.gather(dim, i)).reshape(orig_shape)
     return sparse_tensor
 
 
@@ -73,7 +72,9 @@ def _top_k_total_size(tensor: Tensor, topk_dim: Optional[int]) -> int:
     """Get the total size of the input tensor along the topk_dim dimension. When, the
     dimension is None, get the number of elements in the tensor.
     """
-    return tensor.numel() if topk_dim is None else tensor.shape[topk_dim]
+    top_k_total_size = tensor.numel() if topk_dim is None else tensor.shape[topk_dim]
+    assert top_k_total_size > 0, "Total size of input tensor along the topk_dim has to be greater than 0."
+    return top_k_total_size
 
 
 def _dct_transform(dense: Tensor) -> Tensor:
@@ -83,7 +84,8 @@ def _dct_transform(dense: Tensor) -> Tensor:
         dense (Tensor):
             Input dense tensor (no zeros).
     Returns:
-        transformed dense tensor DCT components
+        (Tensor)
+            transformed dense tensor DCT components
     """
     raise NotImplementedError("Support for DCT has not been implemented yet!")
 
@@ -155,7 +157,7 @@ class SignalSparsity:
         self._dst_top_k_percent = dst_top_k_percent
 
         self._validate_conf()
-        # TODO: [Min] Type checking for the following
+        # TODO (Min): Type checking for the following
         self._transform = torch.fft.fft if algo is Algo.FFT else _dct_transform  # type: ignore
 
     def _validate_conf(self) -> None:
@@ -165,7 +167,9 @@ class SignalSparsity:
         1. validating that one and only one of top_k_element and top_k_percent is set.
         2. Asserting that both element and percentage are in valid ranges.
 
-        This asserts fail if the checking fails.
+        Throws:
+            ValueError:
+                If validation fails.
         """
         # assert that both top_k_elements and top_k_percent aren't set for sst and dst
         def one_and_only(a: Optional[int], b: Optional[float]) -> bool:
@@ -177,7 +181,9 @@ class SignalSparsity:
         ):
             raise ValueError(
                 "One and only one of top_k_element and top_k_percent for "
-                "each of sst and dst must be provided as an argument."
+                "each of sst and dst must be provided as an argument.\n"
+                f"Input values are: sst element={self._sst_top_k_element}, sst percent={self._sst_top_k_percent}, "
+                f"dst element={self._dst_top_k_element}, dst percent={self._dst_top_k_percent}"
             )
 
         # assert that, if top_k_percent is not None, it is a valid number for a percentage.
@@ -185,13 +191,20 @@ class SignalSparsity:
             return a is None or (0.0 < a <= 100.0)
 
         if not (none_or_in_range(self._sst_top_k_percent) and none_or_in_range(self._dst_top_k_percent)):
-            raise ValueError("top_k_percent values for sst and dst has to be in the interval (0, 100]")
+            raise ValueError(
+                "top_k_percent values for sst and dst has to be in the interval (0, 100].\n"
+                f"Input values are: sst percent={self._sst_top_k_percent}, dst percent={self._dst_top_k_percent}"
+            )
 
         def none_or_greater_0(a: Optional[int]) -> bool:
             return a is None or (0 < a)
 
         if not (none_or_greater_0(self._sst_top_k_element) and none_or_greater_0(self._dst_top_k_element)):
-            raise ValueError("top_k_element values for sst and dst has to be greater than 0.")
+            raise ValueError(
+                "top_k_element values for sst and dst has to be greater than 0.\n"
+                f"Input values are: sst element={self._sst_top_k_element} "
+                f"and dst element={self._dst_top_k_element}"
+            )
 
     def dense_to_sst(self, dense: Tensor) -> Tensor:
         """Get Signal Sparse Tensor (SST) from a dense tensor
@@ -207,8 +220,9 @@ class SignalSparsity:
                 Input dense tensor (no zeros).
 
         Returns:
-            Same shaped tensor as the input dense tensor, still in dense format but in frequency
-            domain (complex valued) and has zeros.
+            (Tensor)
+                Same shaped tensor as the input dense tensor, still in dense format but in frequency
+                domain (complex valued) and has zeros.
         """
         top_k_total_size = _top_k_total_size(dense, self._sst_top_k_dim)
         k = _get_k_for_topk(self._sst_top_k_percent, self._sst_top_k_element, top_k_total_size)
@@ -235,7 +249,8 @@ class SignalSparsity:
                 Input SST tensor (has zeros).
 
         Returns:
-            Same shaped tensor, still dense format but has zeros. Non-zeros are top-k delta values.
+            (Tensor)
+                Same shaped tensor, still dense format but has zeros. Non-zeros are top-k delta values.
         """
         pass
 
@@ -254,7 +269,8 @@ class SignalSparsity:
                 Delta sparse tensor, optional.
 
         Returns:
-            A dense tensor in real number domain from the SST.
+            (Tensor)
+                A dense tensor in real number domain from the SST.
         """
         pass
 
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -95,21 +95,26 @@ class SignalSparsity:
 
         Dense -> fft -> top-k -> results.
 
+        For sparsification (or getting the topk) of the tensor in the transformed domain, the sparse
+        mask needs to be generated from the absolute values of the transformed tensors. However, using
+        the mask, we mask out the top_k values from the transformed tensor `w_fft` itself instead.
+        Note: the transformed tensor `w_fft` is a tensor of complex values, whereas `w_fft_abs` is a
+        tensor with real values from which we generate the sparse mask.
+
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
         if self._sst_top_k_element is not None:
-            assert (
+            assert self._sst_top_k_dim is None or (
                 self._sst_top_k_element <= dense.shape[self._sst_top_k_dim]
             ), f"top_k_element along dim {self._sst_top_k_dim} out of range for tensor of shape {dense.shape}."
 
-        w_fft = torch.fft.fft(dense)  # type: ignore
+        w_fft = self._transform(dense)
         w_fft_abs = torch.abs(w_fft)  # get absolute FFT values
 
         # Use the normalized real values for getting the topk mask. The threshold function handles both the cases
         # of when top_k_element or top_k_percent is proviced by the caller.
-        threshold = self.get_threshold(w_fft_abs, self._sst_top_k_element, self._sst_top_k_percent, self._sst_top_k_dim)
-        sps_mask = w_fft_abs >= threshold
+        sps_mask = self._sparse_mask(w_fft_abs, self._sst_top_k_element, self._sst_top_k_percent, self._sst_top_k_dim)
         w_sst = w_fft * sps_mask  # but mask the actual complex FFT values topk
         return w_sst
 
@@ -150,35 +155,69 @@ class SignalSparsity:
         """
         pass
 
-    def get_threshold(
+    def _transform(self, dense: Tensor) -> Tensor:
+        """
+        Applies the transformation algorithm chosen by the caller to the input dense tensor.
+        Returns the transformed tensor.
+
+        Args:
+            dense (Tensor):
+                Input dense tensor.
+        """
+        if self._algo is Algo.FFT:
+            transform = torch.fft.fft  # type: ignore
+        elif self._algo is Algo.DCT:
+            raise NotImplementedError
+        return transform(dense)
+
+    def _sparse_mask(
         self,
         in_tensor: Tensor,
         top_k_element: Optional[int] = None,
         top_k_percent: Optional[float] = None,
         dim: int = -1,
     ) -> Tensor:
-        """Get a mask for a tensor corresponding to a certain sparsity level.
+        """Returns a mask for a tensor with a certain sparsity level corresponding to either
+        the top_k_percent or the top_k_element of the tensor.
+
         Args:
             in_tensor (Tensor)
                 input torch tensor for which sparse mask is generated.
-            sparsity (float)
-                target sparsity of the tensor for mask generation.
+            top_k_element (float, optional)
+                Number of top-k elements to retain in the dense tensor corresponding to the mask.
+            top_k_percent (float, optional)
+                Percent of top-k elements to retain in the dense tensor corresponding to the mask.
+            dim (int, optional):
+                The dimension on which the top-k is done..
+                E.g. -1 is the last dim. None means flatten and top-k on all dims.
+                There is no way to specify multiple dims other than None.
         """
         abs_tensor = torch.abs(in_tensor)
 
+        # when the criterion is top_k_percent
         if self._dst_top_k_percent is not None or self._sst_top_k_percent is not None:
             sparsity = 1 - top_k_percent  # target sparsity in the unit interval [0,1]
             if sparsity == 0.0:  # if sparsity is zero, we want a mask with all 1's
                 threshold = torch.tensor(float("-Inf"), device=in_tensor.device).unsqueeze(dim)
             else:
                 threshold = torch.quantile(abs_tensor, sparsity, dim).unsqueeze(dim)  # type: ignore
+            sps_mask = in_tensor >= threshold
 
+        # when the criterion is top_k_elements
         elif self._dst_top_k_element is not None or self._sst_top_k_element is not None:
-            # top-k along only some dimension dim
-            v, _ = torch.topk(in_tensor, top_k_element, dim)
-            min_v, _ = torch.min(v, dim)
-            threshold = min_v.unsqueeze(dim)
-        return threshold
+            # If dim is None, we flatten the tensor and get the topk for the whole tensor.
+            if dim is None:
+                vals, idx = in_tensor.flatten().topk(top_k_element)
+                flat_tensor = torch.zeros_like(in_tensor.flatten())
+                flat_tensor[idx] = vals
+                sps_mask = flat_tensor.reshape(in_tensor.shape) > 0.0
+            else:
+                # top-k along some dimension dim only
+                vals, idx = in_tensor.topk(top_k_element, dim=dim)
+                topk = torch.zeros_like(in_tensor)
+                sps_mask = topk.scatter(dim, idx, vals) > 0.0
+
+        return sps_mask
 
     def sst_or_dst_to_mask(self) -> None:
         # we shouldn't need this function since going from SST/DST to mask should be a
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -59,10 +59,10 @@ class SignalSparsity:
         algo: Algo = Algo.FFT,
         sst_top_k_dim: Optional[int] = -1,
         sst_top_k_element: Optional[int] = None,
-        sst_top_k_percent: Optional[float] = None,
+        sst_top_k_percent: Optional[float] = 0.1,
         dst_top_k_dim: Optional[int] = -1,
         dst_top_k_element: Optional[int] = None,
-        dst_top_k_percent: Optional[float] = None,
+        dst_top_k_percent: Optional[float] = 0.1,
     ) -> None:
 
         self._algo = algo
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -17,7 +17,7 @@ def _get_k_for_topk(topk_percent: Optional[float], top_k_element: Optional[int],
     simply returns the value for k. Also, ensures k is never 0 to avoid all sparse tensors.
     """
     if top_k_element is not None and top_k_element > top_k_total_size:
-        raise ValueError("top_k_element for sst or dst is larger than max number of" "elements along top_k_dim")
+        raise ValueError("top_k_element for sst or dst is larger than max number of elements along top_k_dim")
 
     if top_k_element is None:
         top_k_element = int(top_k_total_size * topk_percent / 100.0)
@@ -176,7 +176,7 @@ class SignalSparsity:
             and one_and_only(self._dst_top_k_element, self._dst_top_k_percent)
         ):
             raise ValueError(
-                "One and only one of top_k_element and top_k_percent for"
+                "One and only one of top_k_element and top_k_percent for "
                 "each of sst and dst must be provided as an argument."
             )
 
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -98,12 +98,18 @@ class SignalSparsity:
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
+        if self._sst_top_k_element is not None:
+            assert (
+                self._sst_top_k_element <= dense.shape[self._sst_top_k_dim]
+            ), f"top_k_element along dim {self._sst_top_k_dim} out of range for tensor of shape {dense.shape}."
+
         w_fft = torch.fft.fft(dense)  # type: ignore
         w_fft_abs = torch.abs(w_fft)  # get absolute FFT values
 
-        # Next, use the normalized real values for getting the topk mask
+        # Use the normalized real values for getting the topk mask. The threshold function handles both the cases
+        # of when top_k_element or top_k_percent is proviced by the caller.
         threshold = self.get_threshold(w_fft_abs, self._sst_top_k_element, self._sst_top_k_percent, self._sst_top_k_dim)
-        sps_mask = w_fft_abs > threshold
+        sps_mask = w_fft_abs >= threshold
         w_sst = w_fft * sps_mask  # but mask the actual complex FFT values topk
         return w_sst
 
@@ -160,13 +166,14 @@ class SignalSparsity:
         """
         abs_tensor = torch.abs(in_tensor)
 
-        if self._dst_top_k_percent or self._sst_top_k_percent:
-            if top_k_percent == 0.0:  # if sparsity is zero, we want a mask with all 1's
+        if self._dst_top_k_percent is not None or self._sst_top_k_percent is not None:
+            sparsity = 1 - top_k_percent  # target sparsity in the unit interval [0,1]
+            if sparsity == 0.0:  # if sparsity is zero, we want a mask with all 1's
                 threshold = torch.tensor(float("-Inf"), device=in_tensor.device).unsqueeze(dim)
             else:
-                threshold = torch.quantile(abs_tensor, top_k_percent, dim).unsqueeze(dim)  # type: ignore
+                threshold = torch.quantile(abs_tensor, sparsity, dim).unsqueeze(dim)  # type: ignore
 
-        elif self._dst_top_k_element or self._sst_top_k_element:
+        elif self._dst_top_k_element is not None or self._sst_top_k_element is not None:
             # top-k along only some dimension dim
             v, _ = torch.topk(in_tensor, top_k_element, dim)
             min_v, _ = torch.min(v, dim)
--- a/fairscale/experimental/wgit/signal_sparsity.py
+++ b/fairscale/experimental/wgit/signal_sparsity.py
@@ -10,6 +10,60 @@ import torch
 from torch import Tensor
 
 
+# Helper Functions
+def get_k_for_topk(
+    tensor: Tensor, topk_percent: Optional[float], top_k_element: Optional[int], topk_dim: Optional[int]
+) -> int:
+    """Converts the top_k_percent to top_k_element when top_k_percent is provided
+    as the criterion for top-k calculation. When, top_k_element is used as the criterion,
+    simply returns the value for k. Also, ensures k is never 0 to avoid all sparse tensors.
+
+    Args:
+        tensor (Tensor):
+            Input tensor.
+        topk_percent (float, optional)
+            Percent of top-k elements to retain for SST. Default: 0.1
+
+    Returns:
+        k (int)
+            the value of k for use as top-k
+    """
+    if top_k_element is None:
+        dim_numel = tensor.numel() if topk_dim is None else tensor.shape[topk_dim]
+        k = int(dim_numel * topk_percent / 100.0)
+    else:
+        k = top_k_element
+
+    # ensure we never have 100% sparsity in tensor and always have 1 surviving element!
+    k = k if k != 0 else 1
+    return k
+
+
+def scatter_topk_to_sparse_tensor(src_tensor: Tensor, dest_tensor: Tensor, k: int, dim: Optional[int]) -> Tensor:
+    """Scatter the topk values of the dest_tensor to a zero tensor of the same shape
+    at the top-k indices of the src_tensor.
+
+    Args:
+        src_tensor (Tensor)
+            The source tensor whose top-k "indices" are taken and used to extract
+            the corresponding "values" from the dest_tensor.
+        dest_tensor (Tensor)
+            The tensor whose values are gathered according to the top-k indices
+            of the src_tensor, and a zero tensor of same shape is populated with these
+            values at those indices and creates the sparse_dest tensor.
+
+    Returns:
+        Returns a sparse_dest tensor with the values of the dest_tensor corresponding
+        to the top-k indices of the source tensor.
+
+    """
+    _, idx = src_tensor.topk(k, dim=dim)
+    sparse_dest = torch.zeros_like(dest_tensor)
+    # gather the values from the dest_tensor and scatter to a tensor of zeros
+    sparse_dest = sparse_dest.scatter_(dim, idx, dest_tensor.gather(dim, idx))
+    return sparse_dest
+
+
 class Algo(Enum):
     FFT = 0
     DCT = 1
@@ -59,13 +113,12 @@ class SignalSparsity:
         algo: Algo = Algo.FFT,
         sst_top_k_dim: Optional[int] = -1,
         sst_top_k_element: Optional[int] = None,
-        sst_top_k_percent: Optional[float] = 0.1,
+        sst_top_k_percent: Optional[float] = None,
         dst_top_k_dim: Optional[int] = -1,
         dst_top_k_element: Optional[int] = None,
-        dst_top_k_percent: Optional[float] = 0.1,
+        dst_top_k_percent: Optional[float] = None,
     ) -> None:
 
-        self._algo = algo
         self._sst_top_k_dim = sst_top_k_dim
         self._sst_top_k_element = sst_top_k_element
         self._sst_top_k_percent = sst_top_k_percent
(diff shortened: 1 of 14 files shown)


fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -63,17 +82,41 @@ def _validate_conf(self) -> None:
...
+        the mask, we mask out the top_k values from the transformed tensor `w_fft` itself instead.
+        Note: the transformed tensor `w_fft` is a tensor of complex values, whereas `w_fft_abs` is a
+        tensor with real values from which we generate the sparse mask.
+
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
-        pass
+        if self._sst_top_k_element is not None:
+            assert self._sst_top_k_dim is None or (
```
min-xu-ai, line 108: this check can be in the validate_conf? this doesn't depend on the actually size of the dimension of the tensor, thus, it can be checked earlier?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -63,17 +82,41 @@ def _validate_conf(self) -> None:
...
+                self._sst_top_k_element <= dense.shape[self._sst_top_k_dim]
+            ), f"top_k_element along dim {self._sst_top_k_dim} out of range for tensor of shape {dense.shape}."
+
+        w_fft = self._transform(dense)
+        w_fft_abs = torch.abs(w_fft)  # get absolute FFT values
```
min-xu-ai, line 113: you can use w_fft.abs()? Also, for a complex number, what's the abs() doing to it? The comment just repeats what the code is doing, but the comment does explain why abs() is OK here.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -63,17 +82,41 @@ def _validate_conf(self) -> None:
...
+        Note: the transformed tensor `w_fft` is a tensor of complex values, whereas `w_fft_abs` is a
+        tensor with real values from which we generate the sparse mask.
+
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
-        pass
+        if self._sst_top_k_element is not None:
+            assert self._sst_top_k_dim is None or (
+                self._sst_top_k_element <= dense.shape[self._sst_top_k_dim]
```
min-xu-ai, line 109: should this check be moved into the _sparse_mark() function below?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -63,17 +82,41 @@ def _validate_conf(self) -> None:
...
+            ), f"top_k_element along dim {self._sst_top_k_dim} out of range for tensor of shape {dense.shape}."
+
+        w_fft = self._transform(dense)
+        w_fft_abs = torch.abs(w_fft)  # get absolute FFT values
+
+        # Use the normalized real values for getting the topk mask. The threshold function handles both the cases
```
min-xu-ai, line 115: "normalized real value" means w_fft.abs() above?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -63,17 +82,41 @@ def _validate_conf(self) -> None:
...
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
-        pass
+        if self._sst_top_k_element is not None:
+            assert self._sst_top_k_dim is None or (
+                self._sst_top_k_element <= dense.shape[self._sst_top_k_dim]
+            ), f"top_k_element along dim {self._sst_top_k_dim} out of range for tensor of shape {dense.shape}."
+
+        w_fft = self._transform(dense)
```
min-xu-ai, line 112: perhaps rename w_fft into dense_freq in case the algo is not FFT?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -112,6 +155,70 @@ def sst_dst_to_dense(self, sst: Tensor, dst: Tensor = None) -> Tensor:
...
+        Returns the transformed tensor.
+
+        Args:
+            dense (Tensor):
+                Input dense tensor.
+        """
+        if self._algo is Algo.FFT:
+            transform = torch.fft.fft  # type: ignore
+        elif self._algo is Algo.DCT:
+            raise NotImplementedError
```
min-xu-ai, line 170: NotImplementedError -> NotImplementedError()

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -112,6 +155,70 @@ def sst_dst_to_dense(self, sst: Tensor, dst: Tensor = None) -> Tensor:
         """
         pass
 
+    def _transform(self, dense: Tensor) -> Tensor:
+        """
+        Applies the transformation algorithm chosen by the caller to the input dense tensor.
+        Returns the transformed tensor.
```
min-xu-ai, line 161: move this into a "Returns:" section. Check other code in repo.py or sha1_store.py see what it may look like.

Also, add info like: transformed tensor with same shape/dtype as the input tensor, only transformed into frequency domain. Comments should be useful to the reader, should not state what's already obvious from the code.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -112,6 +155,70 @@ def sst_dst_to_dense(self, sst: Tensor, dst: Tensor = None) -> Tensor:
...
+
+        Args:
+            dense (Tensor):
+                Input dense tensor.
+        """
+        if self._algo is Algo.FFT:
+            transform = torch.fft.fft  # type: ignore
```
min-xu-ai, line 168: if this function is just a one line call, why not do that during init?

I.e. in the init:

self._transform = torch.fft.fft if algo == FFT else _dct_placeholder

Then, you don't need this one line function. Just call self._transform directly would be OK? You can define a _dct_placeholder function just to raise the NotImplementedError(), right?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -112,6 +155,70 @@ def sst_dst_to_dense(self, sst: Tensor, dst: Tensor = None) -> Tensor:
...
+        self,
+        in_tensor: Tensor,
+        top_k_element: Optional[int] = None,
+        top_k_percent: Optional[float] = None,
+        dim: int = -1,
+    ) -> Tensor:
+        """Returns a mask for a tensor with a certain sparsity level corresponding to either
+        the top_k_percent or the top_k_element of the tensor.
+
+        Args:
```
min-xu-ai, line 183: Add "Returns:" section too.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -112,6 +155,70 @@ def sst_dst_to_dense(self, sst: Tensor, dst: Tensor = None) -> Tensor:
...
+                Percent of top-k elements to retain in the dense tensor corresponding to the mask.
+            dim (int, optional):
+                The dimension on which the top-k is done..
+                E.g. -1 is the last dim. None means flatten and top-k on all dims.
+                There is no way to specify multiple dims other than None.
+        """
+        abs_tensor = torch.abs(in_tensor)
+
+        # when the criterion is top_k_percent
+        if self._dst_top_k_percent is not None or self._sst_top_k_percent is not None:
```
min-xu-ai, line 198: it is very strange that you access those attributes from self within this function.

In fact, this function can be make into a helper that's outside of this class, right?

Does it really need any info from "self" class or it can operate independently?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -112,6 +155,70 @@ def sst_dst_to_dense(self, sst: Tensor, dst: Tensor = None) -> Tensor:
...
+            sparsity = 1 - top_k_percent  # target sparsity in the unit interval [0,1]
+            if sparsity == 0.0:  # if sparsity is zero, we want a mask with all 1's
+                threshold = torch.tensor(float("-Inf"), device=in_tensor.device).unsqueeze(dim)
+            else:
+                threshold = torch.quantile(abs_tensor, sparsity, dim).unsqueeze(dim)  # type: ignore
+            sps_mask = in_tensor >= threshold
+
+        # when the criterion is top_k_elements
+        elif self._dst_top_k_element is not None or self._sst_top_k_element is not None:
```
min-xu-ai, line 207: Let's chat in VC. I don't know why you need 2 cases and a threshold above. I was thinking we can reduce the percent into an element count and always do a top-K. I actually don't know why a mask is needed. Maybe you are trying to implement value based percentage, not count based percentage?

min-xu-ai (commented):
Thanks for the PR! I have many questions on the code so I skipped the test for now. Let's chat tomorrow in more details.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,8 +54,25 @@ class SignalSparsity:
...
     """
 
-    def __init__(self) -> None:
-        pass
+    def __init__(
+        self,
+        algo: Algo = Algo.FFT,
+        sst_top_k_dim: Optional[int] = -1,
+        sst_top_k_element: Optional[int] = None,
+        sst_top_k_percent: Optional[float] = 0.1,
```
min-xu-ai, line 62: BTW, on line 36 and 45, I meant percent value in [0, 100]. It should be checked. Default value could be 99.9, which means 0.1% non-zeros. I got sparsity definition reversed when I wrote 0.1 above in the comment. Sorry about that.

Also, we should make sure top_k_element > 0 and top_k_percent * dim > 0. I.e. no matter what, we should at least have 1 non-zero element in the resulting sparse vector.

fairscale/experimental/wgit/signal_sparsity.py
riohib: I presume this check cannot be moved into the validate_conf, since we need the input tensor dim. However, it can be moved to the `_sparse_mask` method as you suggest in the later comment. I have moved it there for the next commit.

min-xu-ai: What I meant here is the check of "is None", not the next line after the "or" part. The none check does not need the input tensor dim, right?

fairscale/experimental/wgit/signal_sparsity.py
riohib: the abs is calculating the magnitude of the complex values, similar to the magnitude of 2D vectors. I am adding an explanatory comment on this for clarification. Also thanks for suggesting the `w_fft.abs()` change.

fairscale/experimental/wgit/signal_sparsity.py
riohib: Yes, moving to `_sparse_mask`

fairscale/experimental/wgit/signal_sparsity.py
riohib: The none check is used together with the rest of the assetion, which checks `_sst_topk_element <= in_tensor.shape[self._sst_top_k_dim]`. If `self._sst_top_k_dim`, then that would be an error for the tensor.shape method. We are either checking if we have one of dim=None, or the rest of the assert.

min-xu-ai: Right, I miss read it. I thought you are check top_k_percent not None here. Sorry.

Where do you check one and only one of (element, percent) is not None? I see validate_conf only checking they not both not None, right?

fairscale/experimental/wgit/signal_sparsity.py
riohib: @min-xu-ai We can boil down ` top_k_percent * dim > 0`  to just ` top_k_percent > 0` right? Otherwise, we start needing the tensor at config validation.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,8 +54,25 @@ class SignalSparsity:
...
+        dst_top_k_dim: Optional[int] = -1,
+        dst_top_k_element: Optional[int] = None,
+        dst_top_k_percent: Optional[float] = 0.1,
```
riohib, line 65: @min-xu-ai For both the `sst_top_k_percent` and `dst_top_k_percent`, does it seem that starting with the default as `None` might be more appropriate? Since, we only want one of `top_k_element` and `top_k_percent` to have a value, having a default value forces the caller to proactively choose `dst/sst_top_k_percent = None.` 

E.g. `sparser_2d = SignalSparsity(sst_top_k_element=100, sst_top_k_percent=None, sst_top_k_dim=1, .....)` 
Having all `topk_k_percent` and `topk_k_element` arguments as `None` alleviates this problem.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: We can do this:

top_k_percent * dim -> top_k_element -> top_k_element > 0

this way, both element and percent should share most of the code.

The >0 part is not a validation. It is a low bound. Which means if user specified 0 or small percent, we will never do topk=0, instead, we will always do topk=1 in those cases. Therefore, it is not a validation but a runtime behavior. the reason for this is that I can't think of any valid use case you want topk =0.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: I thought the default is (element=None, percent=0.1), which is a common case since user may want to common do a percent based topk? Are you saying you want to make the common case with percent=None and element=x? What's the x value by default?

riohib: No really, I am saying (element=None, percent=None) seems like a better choice to me. 

1. For example, with (element=None, percent=0.1) default, say if we want to instantiate with (element = 10), we actually need to instantiate it with (element=10, percent=None), because we cannot allow (element = 10, percent=0.1) implicitly.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,66 @@
...
+    else:
+        k = top_k_element
+
+    # ensure we never have 100% sparsity in tensor and always have 1 surviving element!
+    k = k if k != 0 else 1
+    return k
+
+
+def scatter_topk_to_sparse_tensor(src_tensor: Tensor, dest_tensor: Tensor, k: int, dim: Optional[int]) -> Tensor:
+    """Scatter the topk values of the dest_tensor to a zero tensor of the same shape
```
min-xu-ai, line 43: let's rename src_tensor -> top_k_tensor, dest_tensor -> to_be_sparsify_tensor.

should comment why we are doing this. The reason is that the top-k computation might be done with a derived tensor from to_be_sparsified_tensor.

Also, assert these 2 tensors have the same shapes.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +108,81 @@ class SignalSparsity:
...
     """
 
-    def __init__(self) -> None:
-        pass
+    def __init__(
+        self,
+        algo: Algo = Algo.FFT,
+        sst_top_k_dim: Optional[int] = -1,
+        sst_top_k_element: Optional[int] = None,
+        sst_top_k_percent: Optional[float] = None,
```
min-xu-ai, line 116: having both none is a good default?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +108,81 @@ class SignalSparsity:
...
+
+        self._sst_top_k_dim = sst_top_k_dim
+        self._sst_top_k_element = sst_top_k_element
+        self._sst_top_k_percent = sst_top_k_percent
+        self._dst_top_k_dim = dst_top_k_dim
+        self._dst_top_k_element = dst_top_k_element
+        self._dst_top_k_percent = dst_top_k_percent
+
         self._validate_conf()
+        self._transform = torch.fft.fft if algo == Algo.FFT else self._dct_transform  # type: ignore
```
min-xu-ai, line 130: _dct_transform can be a helper? No need to be a class func?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +108,81 @@ class SignalSparsity:
...
+        ), "One and only one of dst_top_k_element and dst_top_k_percent has to be provided as argument"
+
+        # assert, if top_k_percent is not None, it is a valid number for percentage.
+        assert self._sst_top_k_percent is None or (
+            0.0 < self._sst_top_k_percent <= 100.0
+        ), "The value for the argument sst_top_k_percent has to be in the interval (0, 100]"
+
+        assert self._dst_top_k_percent is None or (
+            0.0 < self._dst_top_k_percent <= 100.0
+        ), "The value for the argument dst_top_k_percent has to be in the interval (0, 100]"
```
min-xu-ai, line 156: think how to simplified these checks. make it shorter and easy to understand at the same time.

min-xu-ai: at least same logic on sst/dst should not be repeated.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +108,81 @@ class SignalSparsity:
...
+
+        Args:
+            dense (Tensor):
+                Input dense tensor (no zeros).
+
         Returns:
             Same shaped tensor, still in dense format but in frequency domain and has zeros.
         """
-        pass
+        k = get_k_for_topk(dense, self._sst_top_k_percent, self._sst_top_k_element, self._sst_top_k_dim)
```
min-xu-ai, line 175: we can reduce the helper function's input to just k_ele, k_percent and top_k_total_size? That way, you don't need to pass both tensor and dim into it.
Have a different function that goes from (tensor, dim) -> top_k_total_size could be a reusable thing.

min-xu-ai (commented):
This is really a big improvement and getting better! More comments below to get it even better I think. :-)

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,66 @@
...
+    tensor: Tensor, topk_percent: Optional[float], top_k_element: Optional[int], topk_dim: Optional[int]
+) -> int:
+    """Converts the top_k_percent to top_k_element when top_k_percent is provided
+    as the criterion for top-k calculation. When, top_k_element is used as the criterion,
+    simply returns the value for k. Also, ensures k is never 0 to avoid all sparse tensors.
+
+    Args:
+        tensor (Tensor):
+            Input tensor.
+        topk_percent (float, optional)
```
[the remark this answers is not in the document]
min-xu-ai: only for internal funcs.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,66 @@
...
+            the value of k for use as top-k
+    """
+    if top_k_element is None:
+        dim_numel = tensor.numel() if topk_dim is None else tensor.shape[topk_dim]
+        k = int(dim_numel * topk_percent / 100.0)
+    else:
+        k = top_k_element
+
+    # ensure we never have 100% sparsity in tensor and always have 1 surviving element!
+    k = k if k != 0 else 1
```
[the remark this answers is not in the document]
min-xu-ai: math is better than if. :-)

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: naming is very important for understanding and cognitive load of the code.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: good common case is important.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: be a minimalist. self is basically an arg to a class function. If it is not needed, don't add it.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: DRY principle: don't repeat yourself.

fairscale/experimental/wgit/signal_sparsity.py
riohib: Could you make a comment on what you think about this? 
https://github.com/facebookresearch/fairscale/pull/1034#discussion_r927140788

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: I see. That's true. Ignore my other comment on the default value here then. I was thinking default to user not supplying any of the 2 args. But your idea here is potentially better. It is just that user must know one of the arg is "required". Default case will error out. Make sure the docstring is really clear for them and error msg direct them to how to fix the error when they use both None as the defaults.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: sorry, just did. I didn't see your comment there first. I think your idea is good.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,86 @@
...
+
+def _scatter_topk_to_sparse_tensor(
+    top_k_tensor: Tensor, to_be_sparsify_tensor: Tensor, k: int, dim: Optional[int]
+) -> Tensor:
+    """Scatter the topk values of the to_be_sparsify_tensor to a zero tensor of the same shape
+    at the top-k indices of the top_k_tensor. This function allows top-k computation with a
+    derived tensor from to_be_sparsify_tensor.
+
+    Args:
+        top_k_tensor (Tensor)
```
min-xu-ai, line 35: : after )

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,86 @@
...
+            The source tensor whose top-k "indices" are taken and used to extract
+            the corresponding "values" from the to_be_sparsify_tensor.
+        to_be_sparsify_tensor (Tensor)
+            The tensor whose values are gathered according to the top-k indices
+            of the top_k_tensor, and a zero tensor of same shape is populated with these
+            values at those indices and creates the sparse_tensor tensor.
+        k (int)
+            the value of k for top-k
+        dim (int, optional)
```
min-xu-ai, line 44: this is not optional. I think this is easy to confuse us.

There is an arg being optional, which means it has a default value.

Type hint can be "Optional" with a big O, that means the type is either None or the type inside Optional[]. 

Here, type is Optional, but arg is not optional. I got confused initially too.

min-xu-ai: I suggest change to:

`dim (Optional[int]):`

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,86 @@
...
+        to the top-k indices of the source tensor.
+    """
+    assert (
+        top_k_tensor.shape == to_be_sparsify_tensor.shape
+    ), "top_k_tensor and to_be_sparsify_tensor have different shapes!"
+
+    sparse_tensor = torch.zeros_like(to_be_sparsify_tensor)
+    if dim is None:
+        _, idx = top_k_tensor.flatten().topk(k)
+        sparse_tensor.flatten()[idx] = to_be_sparsify_tensor.flatten()[idx]
```
min-xu-ai, line 59: don't we need to reshape the result back? That a bug here? Not tested?

I suggest the following structure:

```
shape = top_k_tensor.shape
if dim is None and len(shape)>1:
  orig_shape = ...
  sparse_tensor = sparse_tensor.reshape([-1]
  top_k_tensor = top_k_tensor.reshape...
  to_be ... = to_be...reshape...
  dim = -1

# then do gather-scatter uniformly

if orig_shape is not None:
  sparse_tensor = sparse_tensor.reshape(orig_shape)

```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,86 @@
...
+        # gather the values from the to_be_sparsify_tensor and scatter to a tensor of zeros
+        sparse_tensor = sparse_tensor.scatter_(dim, idx, to_be_sparsify_tensor.gather(dim, idx))
+    return sparse_tensor
+
+
+def _top_k_total_size(tensor: Tensor, topk_dim: Optional[int]) -> int:
+    """Get the total size of the input tensor along the topk_dim dimension. When, the
+    dimension is None, get the number of elements in the tensor.
+    """
+    top_k_total_size = tensor.numel() if topk_dim is None else tensor.shape[topk_dim]
```
min-xu-ai, line 71: return directly, not need to assign to a var and then return that var

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -40,7 +120,7 @@ class SignalSparsity:
         dst_top_k_element (int, optional):
             Number of top-k elements to retain for DST. Default: None
         dst_top_k_percent (float, optional):
-            Percent of top-k elements to retain for DST. Default: 0.1
+            Percent of top-k elements to retain for DST. Default: None
```
min-xu-ai, line 123: line 129 the example need to be updated too? Otherwise, that will be incorrect example?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
+
+        self._sst_top_k_dim = sst_top_k_dim
+        self._sst_top_k_element = sst_top_k_element
+        self._sst_top_k_percent = sst_top_k_percent
+        self._dst_top_k_dim = dst_top_k_dim
+        self._dst_top_k_element = dst_top_k_element
+        self._dst_top_k_percent = dst_top_k_percent
+
         self._validate_conf()
+        self._transform = torch.fft.fft if algo == Algo.FFT else _dct_transform  # type: ignore
```
min-xu-ai, line 154: try not to type ignore. There should be a way to make the type error go away. Or add a TODO for me to address it.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
 
     def _validate_conf(self) -> None:
-        """Validating the config is valid.
+        """Validating if the config is valid.
 
-        For example, not both top_k_element and top_k_percent is set.
+        This includes asserting the following:
+        1. validating that both top_k_element and top_k_percent is not set.
```
min-xu-ai, line 160:
```suggestion
        1. validating that one and only one of top_k_element and top_k_percent is set.
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
+        2. Asserting that if top_k_percent is not None, it is a valid number for percentage.
```
min-xu-ai, line 161:
```suggestion
        2. Asserting that both element and percentage in valid ranges.
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
 
-        this should assert fail if checking fails.
+        This asserts fail if the checking fails.
         """
-        pass
+        # assert that both top_k_elements and top_k_percent aren't set for sst and dst
+        def one_and_only(a: Optional[int], b: Optional[float]) -> bool:
```
min-xu-ai, line 166: nice

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
+            self._dst_top_k_element, self._dst_top_k_percent
+        ), "One and only one of top_k_element and top_k_percent for each of sst and dst must be provided as argument."
+
+        # assert that, if top_k_percent is not None, it is a valid number for a percentage.
+        def none_or_in_range(a: Optional[float]) -> bool:
+            return a is None or (0.0 < a <= 100.0)
+
+        assert none_or_in_range(self._sst_top_k_percent) and none_or_in_range(
+            self._dst_top_k_percent
+        ), "top_k_percent values for sst and dst has to be in the interval (0, 100]"
```
min-xu-ai, line 179: need to check top_k_element > 0 too if it is not None.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
-        pass
+        top_k_total_size = _top_k_total_size(dense, self._sst_top_k_dim)
+        k = _get_k_for_topk(self._sst_top_k_percent, self._sst_top_k_element, top_k_total_size)
+        dense_freq = self._transform(dense)
+
+        # NOTE: real_dense_freq can potentially be magnitude of complex frequency components
+        # or DCT transformed components when using DCT (currently not implemented).
+        # TODO: In case of the FFT, the imaginary part can perhaps be quantized or pruning can be
+        # done on the smaller phases.
+        real_dense_freq = torch.abs(dense_freq.real)
```
min-xu-ai, line 206: use .abs() is shorter?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -52,28 +132,81 @@ class SignalSparsity:
...
+
+        sst = _scatter_topk_to_sparse_tensor(real_dense_freq, dense_freq, k, dim=self._sst_top_k_dim)
```
min-xu-ai, line 208: return directly?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,142 @@
...
+        return dict(zip(arg_key_list, vals_list))
+
+    # Validate assertion error is raised when, either
+    # 1. both of sst (or dst) percent and element is set to values (not None).
+    # 2. both of sst (or dst) percent and element is set to None.
+    element = random.randint(0, TENSOR_DIM_UPPER)
+    percent = random.uniform(0, 100)
+    dim = random.randint(-1, 3)
+
+    pytest.raises(AssertionError, SignalSparsity, **kwargs([element, percent, dim, element, None, dim]))
```
min-xu-ai, line 44: very nice!

min-xu-ai: I was wrong when I said we should assert fail on the wrong args. I think we should raise ValueError() instead since those cases are incorrect user input values. Assert is for internal logic errors. Here, this new class is not internal since it can be used outside of wgit eventually too. My bad for giving you wrong initial direction.

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,142 @@
...
+    pytest.raises(AssertionError, SignalSparsity, **kwargs([element, None, dim, element, percent, dim]))
+    pytest.raises(AssertionError, SignalSparsity, **kwargs([None, None, dim, element, None, dim]))
+    pytest.raises(AssertionError, SignalSparsity, **kwargs([element, None, dim, None, None, dim]))
+
+
+def test_sst_sparsity_level():
+    """Tests the dense_to_sst method of SignalSparsity in terms of generating SST at the correct
```
min-xu-ai, line 51: remove "of SignalSparsity" since this whole file is about that class anyway.

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,142 @@
...
+    sparsity level.
+    """
+
+    def get_sst(sst_top_k_percent, dst_top_k_percent=100):
+        sparser_2d = SignalSparsity(sst_top_k_percent=sst_top_k_percent, dst_top_k_percent=100)
+        sst = sparser_2d.dense_to_sst(tensor)
```
min-xu-ai, line 57: return directly.

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,142 @@
...
+    pytest.raises(AssertionError, SignalSparsity, **kwargs([element, None, dim, None, None, dim]))
+
+
+def test_sst_sparsity_level():
+    """Tests the dense_to_sst method of SignalSparsity in terms of generating SST at the correct
+    sparsity level.
+    """
+
+    def get_sst(sst_top_k_percent, dst_top_k_percent=100):
+        sparser_2d = SignalSparsity(sst_top_k_percent=sst_top_k_percent, dst_top_k_percent=100)
```
min-xu-ai, line 56: =100 here or should be =dst_top_k_percent?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,142 @@
...
+        sst = sparser_2d.dense_to_sst(tensor)
+        return sst
+
+    for _ in range(TRIALS):
```
min-xu-ai, line 60: ok, for the rest of test, I suggest we can simplify them and make them more targeted.

1. test for a fix input, expect a fixed output. Say a 5x5 input (not random, fixed in the test code, like 1,2,3,4,5, then 5,4,3,2,1, etc, etc. Then get the sst at element=2/percent=50, compare the output with a fixed expected output. This should be deterministic on all torch versions too. (we will find out)
2. test for some corner cases:
2.1. test a 2D input with dim=None, this should catch the bug where reshape not done on the result
2.2. test a 3D input with dim=1, which is the middle dim
2.3. more cases if we find more bugs in the future.

Again, in all case, fixed input and fixed expected output is good to test this kind of code. You can still do random test like you have done below, but it may not be efficient in catching bugs. 

In general, randomness in testing is not very good. If you have to have it, at least log all the random info and/or have a fixed seed so that random failure can be reproduced with a seed or with the logged info.

In general, testing should be tight, i.e. checking for exact expected value is better since any bug cause small change will be caught. Clearly, tight testing need determinism.

Finally, testing should cover all code paths. i.e. high code coverage. There are tools for measuring code coverage. But mentally you can check if your tests triggers all code paths to be executed when the code is small enough.

min-xu-ai (commented):
Nice improvement. I think there is a bug and I also suggested how to better test it.

tests/experimental/wgit/test_signal_sparsity.py
min-xu-ai: code / comment should be concise but not missing info. useless info increase the cognitive load for readers.

fairscale/experimental/wgit/signal_sparsity.py
riohib: I don't think this is a bug, as Tensor.flatten doesn't mutate the tensor itself, but does in fact perform the value assignment to the indices. Can be replicated by this short snippet:
```
t = torch.ones(5,5)
print(t.shape)
t.flatten()[1] = 7
print(t)
```
I can however make this more explicit according to your suggestion. There is tests covering some aspects of this, but that relies on this strategy as well, so I will add more explicit tests for this. Thank you for pointing the test out.

fairscale/experimental/wgit/signal_sparsity.py
riohib: Thanks for pointing out!

fairscale/experimental/wgit/signal_sparsity.py
riohib: Was considering whether to return directly or be more explicit about what is being returned. Will return directly, thanks! :))

tests/experimental/wgit/test_signal_sparsity.py
riohib: Sounds good. Can be adapted to raise value error!

tests/experimental/wgit/test_signal_sparsity.py
riohib: This test was for sst only, so `dst_top_k_percent=100` (no change as choosing top 100% or all). When we  do add dst implementation didn't want that to impact sst only tests.

tests/experimental/wgit/test_signal_sparsity.py
riohib: Thank you so much for outlining the testing strategy. WIll go ahead with this. I think some of your sub-points within 2. are covered in line 133 of the tests, where we test dims for linear. Will go ahead and integrate the rest of the testing strategy you suggested!

tests/experimental/wgit/test_signal_sparsity.py
min-xu-ai: I see. why dst_top_k_percent default to 100 on line 55? Shouldn't it like sst_top_k_percent that don't have a default?

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: I see. You are right. I missed the fact that left-hand-side expression is just a temp tensor with the same backing.

I think explicit is better here and reshape() doesn't cost much since it should only change the tensor metadata a bit. Any code that is more obviously correct is better.

fairscale/experimental/wgit/signal_sparsity.py
min-xu-ai: The docstring should already talked about what's being returned.

A var should be introduced if it is used at least twice. If it is used only once and the original expression could be used just as well. 

In some case, the code is super unclear with the expression only. In that case, using a comment is better than trying to come up with a var name to explain the expression IMHO.

riohib: That is very helpful, thank you :))

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,90 @@
...
+        to_be_sparsify_tensor (Tensor):
+            The tensor whose values are gathered according to the top-k indices
+            of the top_k_tensor, and a zero tensor of same shape is populated with these
+            values at those indices and creates the sparse_tensor tensor.
+        k (int):
+            the value of k for top-k
+        dim (Optional[int]):
+            dimension for top-k
+
+    Returns:
```
min-xu-ai, line 49: add return type too:

```suggestion
    Returns:
        (Tensor)
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,90 @@
...
+        Returns a sparse_tensor with the same shape as the top_k_tensor and to_be_sparsify_tensor,
+        and populated with the values of the to_be_sparsify_tensor at the indices corresponding
+        to the top-k indices of the source tensor.
```
min-xu-ai, line 52: indent those after adding the type above.

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -40,40 +124,102 @@ class SignalSparsity:
...
+
+        self._sst_top_k_dim = sst_top_k_dim
+        self._sst_top_k_element = sst_top_k_element
+        self._sst_top_k_percent = sst_top_k_percent
+        self._dst_top_k_dim = dst_top_k_dim
+        self._dst_top_k_element = dst_top_k_element
+        self._dst_top_k_percent = dst_top_k_percent
+
         self._validate_conf()
+        # TODO: [Min] Type checking for the following
```
min-xu-ai, line 158:
```suggestion
        # TODO (Min): Type checking for the following
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -40,40 +124,102 @@ class SignalSparsity:
...
-        """Validating the config is valid.
+        """Validating if the config is valid.
 
-        For example, not both top_k_element and top_k_percent is set.
+        This includes asserting the following:
+        1. validating that one and only one of top_k_element and top_k_percent is set.
+        2. Asserting that both element and percentage are in valid ranges.
 
-        this should assert fail if checking fails.
+        This asserts fail if the checking fails.
```
min-xu-ai, line 168:
```suggestion
    Throws:
        ValueError:
            If validation fails.
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -40,40 +124,102 @@ class SignalSparsity:
...
-        pass
+        # assert that both top_k_elements and top_k_percent aren't set for sst and dst
+        def one_and_only(a: Optional[int], b: Optional[float]) -> bool:
+            return (a is None) ^ (b is None)
+
+        if not (
+            one_and_only(self._sst_top_k_element, self._sst_top_k_percent)
+            and one_and_only(self._dst_top_k_element, self._dst_top_k_percent)
+        ):
+            raise ValueError(
```
min-xu-ai, line 178: put the conf var values in the error msg too, like:

raise ValueError(f".... Input values are: {self._sst_top_k_element} {...} {...} {...}")

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,90 @@
...
+        orig_shape = sparse_tensor.shape
+        sparse_tensor = sparse_tensor.reshape(-1)
+        _, idx = top_k_tensor.flatten().topk(k)
+        sparse_tensor[idx] = to_be_sparsify_tensor.flatten()[idx]
+        sparse_tensor = sparse_tensor.reshape(orig_shape)
+    else:
+        _, idx = top_k_tensor.topk(k, dim=dim)
+        # gather the values from the to_be_sparsify_tensor and scatter to a tensor of zeros
+        sparse_tensor = sparse_tensor.scatter_(dim, idx, to_be_sparsify_tensor.gather(dim, idx))
+    return sparse_tensor
```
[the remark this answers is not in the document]
riohib: Oh no, the suggested structure is great and I tried that first. However, was having trouble replicating the correct solutions. So resorted back to tackling which I initially thought was the salient point, discarding the `tensor.flatten()[idx] = ... `, assignment statement.

Upon going through it in more detail, the issue seems to be that scatter and gather on flattened array works with dim=0 only, and not dim=-1. The following changes have been made, which I believe closely replicates the suggested structure:

```
    sparse_tensor = torch.zeros_like(to_be_sparsify_tensor)    
    shape = top_k_tensor.shape
    orig_shape = sparse_tensor.shape
    if dim is None and len(shape) > 1:
        sparse_tensor = sparse_tensor.reshape(-1)
        to_be_sparsify_tensor = to_be_sparsify_tensor.reshape(-1)
        top_k_tensor = top_k_tensor.reshape(-1)
        dim = 0

    _, i = top_k_tensor.topk(k, dim=dim)
    sparse_tensor = sparse_tensor.scatter(dim, i, to_be_sparsify_tensor.gather(dim, i)).reshape(orig_shape)
    return sparse_tensor
```

One modification I am suggesting is 
1. taking out the ` orig_shape = sparse_tensor.shape` outside of the `if clause` and putting it before the `if clause`. This helps us get rid of the last if clause suggested (however now we do a reshape on all cases to `orig_shape` in the second to last line.) Do you think this is okay?

And on another note:
2. Getting rid of `and len(shape) > 1` in the if condition `if dim is None and len(shape) > 1:` makes the code adapt to cases of 1D tensor, which I am not sure if we want or not (the only statement we need is setting `dim=0` for 1D tensor though).

fairscale/experimental/wgit/signal_sparsity.py
riohib: Was wondering if we also want to map the provided inputs to the arguments as well? e.g.
```
raise ValueError(f".... Input values are: sst_top_k_element = {self._sst_top_k_element} , ....<arg name> = {...}  ..")
```
Or just inputs are fine?

min-xu-ai: sure, with names is better. using shorter names is also fine if the msg getting too long. It is mainly for easy diagnosis in case we receives an error report with this exception in the trace.

fairscale/experimental/wgit/signal_sparsity.py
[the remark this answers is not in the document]
min-xu-ai: > Upon going through it in more detail, the issue seems to be that scatter and gather on flattened array works with dim=0 only, and not dim=-1. 

I just did some experiment on this. It seems that dim=-1 works for both 1d and 2d tensors. Here is the commands I tried:

```
import torch
d1 = torch.emtpy(100)
d1 = torch.empty(100)
d2 = torch.empty(100,200)
d
d1
d2
_,i = d1.topk(10, -1)
_,i = d2.topk(10, -1)
_,i2 = d2.topk(10, -1)
_,i1 = d1.topk(10, -1)
i1.shape
i2.shape
d1.gather(-1, i1)
d2.gather(-1, i2)
d1.scatter(-1, i1, d1.gather(-1, i1))
d2.scatter(-1, i2, d2.gather(-1, i2))
```

then on flattened one:
```
In [19]: d3 = d1.reshape(-1)

In [20]: d3.shape
Out[20]: torch.Size([100])

In [21]: d3 = d2.reshape(-1)

In [22]: d3.shape
Out[22]: torch.Size([20000])

In [23]: _,i3 = d3.topk(10, -1)

In [24]: i3.shape
Out[24]: torch.Size([10])

In [25]: d3.scatter(-1, i3, d3.gather(-1, i3))
Out[25]: 
tensor([1.7225e+29, 4.5622e-41, 1.7225e+29,  ..., 3.6434e-44, 3.3631e-44,
        0.0000e+00])
```

Am I missing something?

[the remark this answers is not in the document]
riohib: You're absolutely correct about it! My bad! 
What do you think of the last two points?

fairscale/experimental/wgit/signal_sparsity.py
riohib: Sounds good. Thanks!

fairscale/experimental/wgit/signal_sparsity.py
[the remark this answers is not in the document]
min-xu-ai: Why not first change the code as you see fit. I can comment on it once I see the actual code?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,92 @@
...
+            The tensor whose values are gathered according to the top-k indices
+            of the top_k_tensor, and a zero tensor of same shape is populated with these
+            values at those indices and creates the sparse_tensor tensor.
+        k (int):
+            the value of k for top-k
+        dim (Optional[int]):
+            dimension for top-k
+
+    Returns:
+        (Tensor)
```
min-xu-ai, line 48:
```suggestion
        (Tensor):
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,92 @@
...
+            to the top-k indices of the source tensor.
+    """
+    assert (
+        top_k_tensor.shape == to_be_sparsify_tensor.shape
+    ), "top_k_tensor and to_be_sparsify_tensor have different shapes!"
+
+    sparse_tensor = torch.zeros_like(to_be_sparsify_tensor)
+    shape = top_k_tensor.shape
+    orig_shape = sparse_tensor.shape
+    if dim is None and len(shape) > 1:
```
min-xu-ai, line 60:
```suggestion
    if dim is None and len(orig_shape) > 1:
```

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,92 @@
...
+            Returns a sparse_tensor with the same shape as the top_k_tensor and to_be_sparsify_tensor,
+            and populated with the values of the to_be_sparsify_tensor at the indices corresponding
+            to the top-k indices of the source tensor.
+    """
+    assert (
+        top_k_tensor.shape == to_be_sparsify_tensor.shape
+    ), "top_k_tensor and to_be_sparsify_tensor have different shapes!"
+
+    sparse_tensor = torch.zeros_like(to_be_sparsify_tensor)
+    shape = top_k_tensor.shape
```
min-xu-ai, line 58: delete this line since we can just use orig_shape below?

fairscale/experimental/wgit/signal_sparsity.py
```diff
@@ -4,10 +4,92 @@
...
+    orig_shape = sparse_tensor.shape
+    if dim is None and len(shape) > 1:
+        sparse_tensor = sparse_tensor.reshape(-1)
+        to_be_sparsify_tensor = to_be_sparsify_tensor.reshape(-1)
+        top_k_tensor = top_k_tensor.reshape(-1)
+        dim = -1
+
+    _, i = top_k_tensor.topk(k, dim=dim)
+    sparse_tensor = sparse_tensor.scatter(dim, i, to_be_sparsify_tensor.gather(dim, i)).reshape(orig_shape)
```
min-xu-ai, line 67: return directly?

min-xu-ai (commented):
the main code looks great! You will change the tests, right?

fairscale/experimental/wgit/signal_sparsity.py
riohib: yes, can be deleted, they should be of the same shape.

fairscale/experimental/wgit/signal_sparsity.py
riohib: I have mostly changed the tests accordingly. Please take a look and provide feedback. Thanks!!

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,131 @@
...
+        """verifies if the topk element and top k percent returns the same results
+        for same sparsity in a fixed tensor along some dim.
+        """
+        _, sst_percent = sst_and_default_sps(tensor, percent, None, dim=dim)
+        _, sst_elem = sst_and_default_sps(tensor, None, element, dim=dim)
+        assert all((sst_percent == sst_elem).flatten())
+
+    for tensor in tensors:
+        dim_list = [None] + list(range(tensor.dim()))
+        for dim in dim_list:
```
min-xu-ai, line 108: instead of looping, perhaps parameterize this test?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,131 @@
...
+SPARSITY_ATOL = 8e-2
+FFT_RECONS_ATOL = 2e-3
+
+
+@pytest.fixture
+def sst_and_default_sps():
+    """A fixture returning a function that calculates the sst from pytorch operations
+    and compares it with the sst from dense_to_sst method."""
+
+    def assert_and_get_sps_sst(tensor, percent, element, dim):
```
min-xu-ai, line 22: why does this need to be a fixture? can't it be a helper function?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,131 @@
...
+        pytest.raises(ValueError, SignalSparsity, **kwargs(args))
+
+
+def test_dense_to_sst(sst_and_default_sps):
+    """Tests the dense_to_sst method with fixed inputs"""
+    tensors = list()
+    tensors.append(torch.arange(20).reshape(4, 5))
+    tensors.append(torch.arange(80).reshape(4, 5, 4))
+
+    def elem_percent_equality(tensor, percent, element, dim):
```
min-xu-ai, line 98: using torch.testing.close()?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -0,0 +1,131 @@
...
+        dim_list = [None] + list(range(tensor.dim()))
+        for dim in dim_list:
+            # Should return the FFT transformed tensor with top_100_percent for sst
+            sparser_2d = SignalSparsity(sst_top_k_percent=100, sst_top_k_dim=dim, dst_top_k_percent=100)
+            assert all((sparser_2d.dense_to_sst(tensor) == torch.fft.fft(tensor)).flatten())
+
+            # Test for different sparsity levels using top_k_percent and top_k_elements
+            for percent, element in zip([20, 40, 60, 80], [1, 2, 3, 4]):
+                sst_and_default_sps(tensor, percent=percent, element=None, dim=dim)
+                sst_and_default_sps(tensor, percent=None, element=element, dim=dim)
```
min-xu-ai, line 116: do we need to test those different values? I would say test only a common case + corner cases.

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
-        dim = None if (dim == -1) else dim
-        assert all((default_sst.abs().count_nonzero(dim) == k).flatten())
-
-        sst = sparser_2d.dense_to_sst(tensor)
-        assert all((default_sst == sst).flatten())
-        assert sst.shape == tensor.shape
-        return default_sst, sst
-
-    return assert_and_get_sps_sst
+from fairscale.experimental.wgit.signal_sparsity import SignalSparsity
```
min-xu-ai, line 9: nice, no more access of internal functions.

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
+
+
+def expected_io_vals():
```
min-xu-ai, line 12: expected_io_vals -> get_test_params?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
+        [
+            [[1.0 + 0.0j, -1.0 + 0.0j], [5.0 + 0.0j, -1.0 + 0.0j]],
+            [[9.0 + 0.0j, -1.0 + 0.0j], [13.0 + 0.0j, -1.0 + 0.0j]],
+            [[17.0 + 0.0j, -1.0 + 0.0j], [21.0 + 0.0j, -1.0 + 0.0j]],
+        ],
+        dtype=torch.complex64,
+    )
+
+    tensor_4x3 = torch.arange(12).reshape(4, 3)
+    tensor_2x2x3 = torch.arange(12).reshape(3, 2, 2)
```
min-xu-ai, line 58: move these before the expected ones?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
-        return default_sst, sst
-
-    return assert_and_get_sps_sst
+from fairscale.experimental.wgit.signal_sparsity import SignalSparsity
+
+
+def expected_io_vals():
+    """Helper function to create and return a list of tuples of the form:
+    (in_tensor, expected_tensor, dim, percent) to be used as parameters for tests. The top-k
+    value for the expected SST was set at 2.
```
min-xu-ai, line 15: not sure where is the hardcoded "2" this comment is talking about. Did I misunderstand it?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
+    """
+    # Expected SST output for 4x3 tensor of ascending ints
+    expected_4x3_None = torch.tensor(
+        [
+            [0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j],  # with dim=None, top-2
```
min-xu-ai, line 20: nice comment so that we know only 2 values are non-zero

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
+        ],
+        dtype=torch.complex64,
+    )
+
+    expected_4x3_0 = torch.tensor(
+        [
+            [0.0000000 + 0.0000000j, 0.0000000 + 0.0000000j, 0.0000000 + 0.0000000j],  # with dim=0, top-2
+            [0.0000000 + 0.0000000j, 0.0000000 + 0.0000000j, 0.0000000 + 0.0000000j],
+            [21.0000000 + 0.0000000j, -1.5000000 + 0.8660254j, -1.5000000 - 0.8660254j],
+            [30.0000000 + 0.0000000j, -1.5000000 + 0.8660254j, -1.5000000 - 0.8660254j],
```
min-xu-ai, line 33: did you double check the second and third cols should have the same top-2 values? I suppose the input cols are different there, right?

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
+        ],
+        dtype=torch.complex64,
+    )
+
+    expected_4x3_1 = torch.tensor(
+        [
+            [3.0000000 + 0.0000000j, -1.5000000 + 0.8660254j, 0.0000000 + 0.0000000j],  # with dim=1, top-2
+            [12.0000000 + 0.0000000j, -1.5000000 + 0.8660254j, 0.0000000 + 0.0000000j],
+            [21.0000000 + 0.0000000j, -1.5000000 + 0.8660254j, 0.0000000 + 0.0000000j],
+            [30.0000000 + 0.0000000j, -1.5000000 + 0.8660254j, 0.0000000 + 0.0000000j],
```
min-xu-ai, line 43: it is interesting that the first elements are different but second elements are the same between the rows. Seems to make sense but worth double checking the math.

tests/experimental/wgit/test_signal_sparsity.py
```diff
@@ -6,50 +6,63 @@
...
+        ],
+        dtype=torch.complex64,
+    )
+
+    expected_2x2x3 = torch.tensor(
+        [
+            [[1.0 + 0.0j, -1.0 + 0.0j], [5.0 + 0.0j, -1.0 + 0.0j]],
```
min-xu-ai, line 50: add a comment on what dim/top is for this output? Also, double check numerics please.

min-xu-ai (commented):
so cool! much better IMHO. Hope you agree too. I have some questions.

min-xu-ai:
btw, how many tests does it run and long does it report when you run:

`pytest -s tests/experimental/wgit/test_signal_sparsity.py`

tests/experimental/wgit/test_signal_sparsity.py
riohib: Oh right. I was trying to express that these expected output tensors were created when we select top_k_element = 2 and dim=dim from the input tensor. Basically, these outputs are only expected when top_k_element = 2 (or equivalent top_k_percents) is set on Signal Sparsity. Hence, the 2. Do you suggest a better way to express this? May be: `The top-k value for the expected SST "should be" set at 2.` ?


```

**Comment by riohib** [comment]
> btw, how many tests does it run and long does it report when you run:
> 
> `pytest -s tests/experimental/wgit/test_signal_sparsity.py`

It runs 16 tests and I assume you are asking for how long it takes to run? It's taking a little less than 2 seconds to run.

File: tests/experimental/wgit/test_signal_sparsity.py (hunk 55 above)
Comment (min-xu-ai) [reply tests/experimental/wgit/test_signal_sparsity.py:15]: Why not just omit this info here but return 2 as the number of elements in top-k at the end of this function as a test_param? That way, there is no hardcoding and 2 is close to the expected value here in the same function at least. thoughts?

**Comment by min-xu-ai** [comment]
> > btw, how many tests does it run and long does it report when you run:
> > `pytest -s tests/experimental/wgit/test_signal_sparsity.py`
> 
> It runs 16 tests and I assume you are asking for how long it takes to run? It's taking a little less than 2 seconds to run.

pytest should print the time at the end BTW.

File: tests/experimental/wgit/test_signal_sparsity.py (hunk 58 above)
Comment (riohib) [reply tests/experimental/wgit/test_signal_sparsity.py:43]: Yeah, I noticed that too. Here is the output from the pytorch fft on this tensor. I am only guessing it might be due to the nice ascending structure of the input tensor.

```
In [4]: tensor = torch.arange(12).reshape(4, 3)
   ...:    torch.fft.fft(tensor)
Out[4]:
tensor([[ 3.0000+0.0000j, -1.5000+0.8660j, -1.5000-0.8660j],
        [12.0000+0.0000j, -1.5000+0.8660j, -1.5000-0.8660j],
        [21.0000+0.0000j, -1.5000+0.8660j, -1.5000-0.8660j],
        [30.0000+0.0000j, -1.5000+0.8660j, -1.5000-0.8660j]])
        ```

tests/experimental/wgit/test_signal_sparsity.py
riohib: I think this is relevant here as well: https://github.com/facebookresearch/fairscale/pull/1034#discussion_r930498816

riohib:
> > > btw, how many tests does it run and long does it report when you run:
> > > `pytest -s tests/experimental/wgit/test_signal_sparsity.py`
> > 
> > 
> > It runs 16 tests and I assume you are asking for how long it takes to run? It's taking a little less than 2 seconds to run.
> 
> pytest should print the time at the end BTW.

Here is the output from the pytest:
`pytest -s ./tests/experimental/wgit/test_signal_sparsity.py  2.23s user 5.56s system 259% cpu 2.999 total`

tests/experimental/wgit/test_signal_sparsity.py
riohib: Sounds much better tbh. Making the change. Thanks!

tests/experimental/wgit/test_signal_sparsity.py
min-xu-ai: 👍perhaps worth checking with numpy or scipy

riohib: Better idea, taking a look!

riohib: scipy also seems to agree with the results for this:
```
In [5]: from scipy.fft import fft
   ...: array = np.arange(12).reshape(4,3)
   ...: fft(array)
Out[5]:
array([[ 3. -0.j       , -1.5+0.8660254j, -1.5-0.8660254j],
       [12. -0.j       , -1.5+0.8660254j, -1.5-0.8660254j],
       [21. -0.j       , -1.5+0.8660254j, -1.5-0.8660254j],
       [30. -0.j       , -1.5+0.8660254j, -1.5-0.8660254j]])
```

min-xu-ai (approved):
LGTM!

min-xu-ai:
Seems all working fine

Issue #1042, opened 2022-07-26
[Defect in 0.4.7] pygit2/pgzip not automatically installed.


AlexanderVanEck:
Good morning! Thank you for publishing another hotfix for `fairscale` 🙇‍♂️ I'm sure this will be well received.

# Bug description
Due to the inclusion of entry_points the `pygit2/pgzip` packages are now required by fairscale even though its experimental and optional.

```
entry_points={"console_scripts": ["wgit = fairscale.experimental.wgit.__main__:main"]},
```
https://github.com/facebookresearch/fairscale/blob/main/setup.py#L72

# Steps to reproduce
(in a virtualenv)
```
$ pip install fairscale==0.4.7
$ python -c "import fairscale"
...
Error: please pip install pygit2 module to use wgit
```

# Expected behavior
(in a virtualenv)
```
$ pip install fairscale==0.4.7
$ python -c "import fairscale"
$ # no error
```

# Workaround
(in a virtualenv)
```
$ pip install fairscale!=0.4.7
$ python -c "import fairscale"
$ # no error
```

Would the solution here be to add `pygit2` & `pgzip` to requirements.txt or optionally import them in https://github.com/facebookresearch/fairscale/blob/main/fairscale/experimental/wgit/__init__.py ?

anupambhatnagar:
@min-xu-ai @riohib Can you please have a look at this?

min-xu-ai:
Thanks for the report! I am on it.

min-xu-ai:
Now I see why experimental wgit is imported

```
/home/m1n/git/fairscale/fairscale/experimental/nn/offload.py:435: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at  ../torch/csrc/utils/tensor_numpy.cpp:68.)
  offload_device: torch.device = torch.device("cpu"),
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/m1n/git/fairscale/fairscale/__init__.py", line 12, in <module>
    from . import nn
  File "/home/m1n/git/fairscale/fairscale/nn/__init__.py", line 9, in <module>
    from .data_parallel import FullyShardedDataParallel, ShardedDataParallel
  File "/home/m1n/git/fairscale/fairscale/nn/data_parallel/__init__.py", line 8, in <module>
    from .fully_sharded_data_parallel import FullyShardedDataParallel, OffloadConfig, TrainingState, auto_wrap_bn
[...]

PR #1045, opened 2022-07-26
[test] setup.py dependency testing

min-xu-ai:
## What does this PR do?
Help catch issues like #1042 

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,6 @@
 # FairScale should only depends on torch, not things higher level than torch.
+# Note1: setup.py automatically reads this file to setup install time dependencies.
+# Note2: we use >= in this file but == in requirements-dev.txt for determinism
+#        in testing.
 torch >= 1.8.0
+numpy >= 1.22.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,6 +273,10 @@ jobs:
       - <<: *check_test_list
       - <<: *setup_venv
 
+      # Do this first to test repo dependencies. Basic import should work after this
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1,2 +1,6 @@
 # FairScale should only depends on torch, not things higher level than torch.
+# Note1: setup.py automatically reads this file to setup install time dependencies.
+# Note2: we use >= in this file but == in requirements-dev.txt for determinism
+#        in testing.
 torch >= 1.8.0
+numpy >= 1.22.0
```
anupambhatnagar, line 6: Since torch depends on numpy, we don't need to add this here. https://github.com/pytorch/pytorch/blob/master/requirements.txt#L5

min-xu-ai: I thought that too. But pip doesn’t do that. If this line is removed. The test would fail. You can try pip install fairscale without any packages already installed. Import would fail.

anupambhatnagar: How did the installation of other packages needed for pytorch work earlier? How come we are finding this now?

anupambhatnagar (approved):

requirements.txt
anupambhatnagar: Found the answer. In CI we are downloading the pytorch wheel which contains all the dependencies. I think either we should list all the dependencies in requirements.txt or just leave it at torch>=1.8. What do you think @min-xu-ai ?

PR #1085, opened 2022-09-27
Fsdp ptd

ruanslv:
- Disable prefetching (re-enable 2 PG backward)
- Check queue at the beginning of all-gather
- Update queue after a free event


min-xu-ai:
Just curious, is this PR trying to the "limiter" thing as PTD's fsdp here?

ruanslv:
Yes exactly. `events[-1].synchronize()` doesn't seem to be working as is, however.

PR #1086, opened 2022-10-04
Fix gradient accumulation

### LINKED ISSUE #1062: AdaScale with gradient accumulation: Sum or average of gradients?
Hi!

I am wondering what is the intended way to implement gradient accumulation when using AdaScale. Usually, for gradient accumulation, I would use a training loop like this

    for i, (X, y) in enumerate(train_loader):
        if (i+1) % num_accum_steps == 0:
            loss = loss_fn(model(X), y) / num_accum_steps
            loss.backward()
            opt.step()
            opt.zero_grad()
        else:
            with model.no_sync():
                loss = loss_fn(model(X), y) / num_accum_steps
                loss.backward()

That is, when the optimizer step is applied, the gradient stored in `p.grad` is the **average** gradient across accumulation steps. However, this [line](https://github.com/facebookresearch/fairscale/blob/main/fairscale/optim/adascale.py#L460) in the AdaScale code makes me think that AdaScale expects `p.grad` to contain the **sum** of gradients across accumulation steps.

Could somebody clarify what's the intended way?

Thanks!

cyugao:
Add ``is_scaled_loss`` flag to support both scaled / unscaled loss Fix ``test_grad_accum`` and``test_set_num_gradients_to_accumulate``

## What does this PR do?
Fixes #1062.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/optim/adascale.py
+++ b/fairscale/optim/adascale.py
@@ -130,16 +130,17 @@ class AdaScale(Optimizer):
             between each optimizer step. This can be changed during
             training as long as the train loop changes gradient accumulation
             accordingly.
+            The loss in each pass can be either scaled or unscaled. See `is_scaled_loss` below.
             Default to 1, which does not accumulate gradients.
+        is_scaled_loss (bool):
+            If True, assume that the loss is scaled by `num_gradients_to_accumulate`.
+            If False, the loss is not scaled.
+            Default: True.
         debias_ewma (bool):
             (experimental) Use debias exponential moving average
             for smoothing and mu and sigma variables. False will
             use the method in the paper's Appendix B.3.
             Default: True, which is what have been validated so far.
-        is_scaled_loss (bool):
-            If True, assume that the loss is scaled by ``num_gradients_to_accumulate``.
-            If False, the loss is not scaled.
-            Default: True.
     """
 
     def __init__(
@@ -149,8 +150,8 @@ class AdaScale(Optimizer):
         scale: Optional[float] = None,
         smoothing: float = None,
         num_gradients_to_accumulate: int = 1,
-        debias_ewma: bool = True,
         is_scaled_loss: bool = True,
+        debias_ewma: bool = True,
     ):
         # Init hook_handles list, otherwise, a partial init'ed object may fail in ``__del__``.
         self._hook_handles: List[Any] = []
@@ -468,8 +469,11 @@ class AdaScale(Optimizer):
         if self._num_grads_to_accum > 1:
             # Handle scaling for for gradient accumulation
             if self._is_scaled_loss:
+                # If loss is scaled down, we need to scale the local gradients back by a factor of _num_grads_to_accum squared;
+                # total_grad_sqr is already scaled by _num_grads_to_accum squared.
                 local_grad_sqr *= self._num_grads_to_accum**2
             else:
+                # If loss is not scaled, local gradients are correct, but we need to scale the total_grad_sqr down to account for gradient accumulation.
                 total_grad_sqr /= self._num_grads_to_accum**2
 
         # See appendix B.3 of the paper.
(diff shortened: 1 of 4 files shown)


fairscale/optim/adascale.py
```diff
@@ -136,6 +136,10 @@ class AdaScale(Optimizer):
             for smoothing and mu and sigma variables. False will
             use the method in the paper's Appendix B.3.
             Default: True, which is what have been validated so far.
+        is_scaled_loss (bool):
```
min-xu-ai, line 139: Can you add to line 132 above and refer to this option so that people know there are 2 ways to do grad_accum?

fairscale/optim/adascale.py
```diff
@@ -453,17 +459,19 @@ def _final_callback(self) -> None:
...
-            # np array doesn't support /=.
-            total_grad_sqr = total_grad_sqr / (self._num_grads_to_accum**2)
 
         # Wait for all_reduce to be done and move it to cpu & np.
         if work:
             work.wait()
         local_grad_sqr = self._local_grad_sqr.cpu().numpy()
 
+        if self._num_grads_to_accum > 1:
+            # Handle scaling for for gradient accumulation
```
min-xu-ai, line 469: Can you expand this comment and document the reason behind line 471 and 473? It is not easy to understand and we may forget its reasons without detailed explanation here. :)

fairscale/optim/adascale.py
```diff
@@ -453,17 +459,19 @@ def _final_callback(self) -> None:
...
+            if self._is_scaled_loss:
+                local_grad_sqr *= self._num_grads_to_accum**2
+            else:
+                total_grad_sqr /= self._num_grads_to_accum**2
```
min-xu-ai, line 473: interesting, `/=` not worked in the past, maybe newer numpy works with it now?

fairscale/optim/adascale.py
```diff
@@ -509,7 +517,12 @@ def step(self, *args: Any, **kwargs: Any) -> Optional[float]:
         original_lr = []
         for idx, param_group in enumerate(self._optimizer.param_groups):
             original_lr.append(param_group["lr"])
-            param_group["lr"] = self.gain(pg_idx=idx) * param_group["lr"]
+            if self._num_grads_to_accum > 1 and not self._is_scaled_loss:
+                # Divide by num_grads_to_accum to account for gradient accumulation.
```
min-xu-ai, line 521: this comment only states the same info as the code, which is not very useful. Let's explain that when scaled loss is not used, the gain computed is correct, but in order to match what DDP does, we need to scale the grad back by N.

However, with momentum, scale LR by N here is not the same as scale the param.grad by N in the DDP case, right?

tests/optim/test_single_node_adascale.py
```diff
@@ -194,16 +195,23 @@ def test_lr_scheduler():
 
 @skip_if_no_cuda
 @pytest.mark.parametrize("debias_ewma", [True, False])
-def test_add_param_group(debias_ewma):
-    """Test AdaScale supports add_param_group() API."""
+@pytest.mark.parametrize("is_scaled_loss", [True, False])
```
min-xu-ai, line 198: very nice to introduce a bool for this

tests/optim/test_single_node_adascale.py
```diff
@@ -82,14 +82,14 @@ def test_grad_accum(test_case, cpu):
         in_data_0 = Tensor(in_data[0])
         if not cpu:
             in_data_0 = in_data_0.cuda()
-        out = model(in_data_0)
-        out.sum().backward()
+        loss = model(in_data_0).sum() / 2
```
min-xu-ai, line 85: in this test, can we add expected grad to adascale_test_data on line 63? That would explicitly check that the grad in all three cases below are the same:
1. ddp
2. grad with loss scale
3. grad without loss scale but with a final grad avg step

This should ensure there is no bug I suspected above in the step() function.

min-xu-ai (commented):
Very nice! I have a few minor suggestions and a one question below.

fairscale/optim/adascale.py
cyugao: I think if you use `a /= 2` when `a` is an integer array it will break since the result will be float, otherwise it should be fine?

min-xu-ai (approved):
LGTM! Good luck merge it.

PR #1106, opened 2023-02-17
Remove `torch._six` from `__init__.py`

malfet:
Just use `torch.inf`, as torch._six is gone after https://github.com/pytorch/pytorch/pull/94709


min-xu-ai:
Thanks!

PR #1110, opened 2023-03-08
Fix bibtex entry

mrbaozi:
I was using FairScale for a publication and noticed that the bibtex entry leads to a reference like `authors, FairScale` or `authors, F.`. I'm guessing this is not intended and should rather read as `FairScale authors`. For this, the author field needs to be wrapped in an additional pair of braces (so as to be interpreted as one unit by (bib)latex).

This PR simply fixes that.


min-xu-ai:
Thanks!

PR #1125, opened 2023-05-20
added option to do backward AG over smaller set of gpus instead of full DDP world

ngoyal2707:


awgu (commented):
This makes sense to me!

awgu:
One more comment: I am not familiar with how the model checkpointing works in Fairscale FSDP, but one concern might be what happens if a user tries to checkpoint the model only after forward (and backward has not run yet). Will this work out of the box?

Perhaps, this is not a major concern for your use case since may no one will save a state dict after only forward.

PR #1136, opened 2023-09-17
Fp8 all gather hack

jspark1105:
This is based on ngoyal_added_zero2_shard_modelparams_multiple_gpus and adding hacks to use fp8 all-gather with Nvidia's transformer engine.

This depends on transformer engine changes in https://github.com/facebookresearch/TransformerEngine/pull/20

To use fp8 allgather, set compute_dtype=torch.int8 (due to the lack of native fp8 type support yet in PyTorch) and mixed_precision=True
We separate out precision critical parameters like affine weights for norm as non flattened params and hard-code we use bf16.
We have to workaround the lack of native fp8 type support in a couple of places, and we hard code special handling for a couple of TransformerEngine layers like Linear, LayerNormLinear, and LayerNormMLP in _cast_fp32_param_shards_to_fp16 to access their fp8 meta data to quantize with right scales (TODO: we may want to extract this as a user customizable call back functions?)


jspark1105:
Will merge main_grad related changes with https://github.com/facebookresearch/fairscale/pull/1142

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1689,23 +1771,31 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
 
-            if self.mixed_precision and self.fp32_reduce_scatter:
-                # Cast grad to FP32.
-                param.grad.data = param.grad.data.to(param.dtype)
+            if self.mixed_precision:
+                if self.fp32_reduce_scatter:
+                    # Cast grad to FP32.
+                    grad_or_main_grad.data = grad_or_main_grad.to(param.dtype)
+                elif self._is_fp8_dtype():
+                    # Use bf16 wgrad for fp8 weights (TODO: handle fp8 wgrad)
```
jianyuh, line 1781: Currently this is not working with the latest FP8 wgrad ?

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1393,7 +1447,11 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
 
         # For root and mixed precision, we convert the input to FP16 (no_grad is needed for
         # the conversion).
-        is_bf16 = self.compute_dtype == torch.bfloat16
+        is_bf16 = self.compute_dtype in [
```
jianyuh, line 1450: nit: is_bf16_or_fp8

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
jspark1105: This meant to be for future work when we have fp8 reduce-scatter. I'll update the comment.

fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -2265,8 +2361,7 @@ def local_metadata_dict(self) -> Dict[str, Any]:
                         backing_param_name = m.module.flat_param_names[i]
                         names, shapes, numels = m.module.metadata(i)
                     else:
-                        assert len(m._param_name_groups[i]) == 1
-                        backing_param_name = m._param_name_groups[i][0]
+                        backing_param_name = m._param_name_groups[m._num_flatten_params][i - m._num_flatten_params]
```
jspark1105, line 2364: Need to make sure checkpointing works properly with this.

PR #1132, opened 2023-08-28
Fix typo in argument ParallelEmbedding argument assignment

hessamb:
## What does this PR do?
This fixes a small typo in ParallelEmbedding where `scale_grad_by_freq` was assigned to `self.norm_type`.

PR #1138, opened 2023-09-21
Fix the parameter in ParallelEmbedding

taowangcheng:
The parameter passing is wrong in ParallelEmbedding. It should be the same as that in class VocabParallelEmbedding. https://github.com/facebookresearch/fairscale/blob/164cc0f3170b4a3951dd84dda29c3e1504ac4d6e/fairscale/nn/model_parallel/layers.py#L111 This may be a typo error.

cc @ezyang @likethesky @myleott @froody @Celebio


Truewarlock (approved):

awgu:
I think this is the same as https://github.com/facebookresearch/fairscale/pull/1132.

PR #1144, opened 2023-10-10
Added fns for manual free, reduce-scatter; removed stream sync if event sync

awgu:
This PR:
1. Adds `_run_all_post_backward_hooks()` to run `_post_backward_hook()` on all parameters of all FSDP modules in the calling module's `.modules()`.
    - This is used for manually reduce-scattering gradients after PP send/recv for input gradients.
2. Adds `_free_all_full_params()` to run `_free_full_params()` on all parameters of all FSDP modules in the calling module's `.modules()`.
    - This is used for freeing parameters (and hence blocking the CPU thread if using CUDA stream synchronization) before PP send/recv for input gradients.
3. Disables the CUDA stream synchronization in `_free_full_params()` if both `limit_all_gather_events=True` and `limit_reduce_scatter_events=True` (requiring both just to be conservative).
    - On the first order, we do not need the CUDA stream synchronization if using the CUDA event synchronization. This change allows us to delay the CPU blocking until later when truly needed.
4. Adds profiler `record_function()` calls for the all-gather CPU sync and reduce-scatter CPU sync.

### DIFF
--- a/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
+++ b/fairscale/nn/data_parallel/fully_sharded_data_parallel.py
@@ -1930,6 +1930,16 @@ class FullyShardedDataParallel(nn.Module):
                     assert self._output_pre_backward_hook_registered is not None  # make mypy happy
                     self._output_pre_backward_hook_registered.clear()
 
+    def _run_all_post_backward_hooks(self) -> None:
+        for module in self.modules():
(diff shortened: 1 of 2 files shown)


awgu:
Changes planned: only call `torch.cuda.current_stream().synchronize()` if freeing parameters is not a no-op.

awgu:
Close for now since not useful.

PR #1147, opened 2023-10-29
[Not for merge] fp8allgather debug

jiecaoyu:
WIP


fairscale/nn/data_parallel/fully_sharded_data_parallel.py
```diff
@@ -1428,22 +1471,70 @@ def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
...
+                    continue
+                d = {info[0]: info[1] for info in p._param_infos}
+                for n, m in d.items():
+                    # Previous iteration was grad_enabled
+                    if m.fp8_meta.get("update_amax_and_scale_fwd", False):
+                        if m.fp8_meta["recipe"].reduce_amax:
+                            FP8GlobalStateManager.copy_amax_from_global_buffer(
+                                m.fp8_meta, forward=True
+                            )
+                            # FIXME update_weight_scale_inv is only True for the first micro-batch
```
jspark1105, line 1526: As discussed it's OK to update weight_scale_inv multiple times but it can be still annoying if we see numerical differences btw no-PP and PP (actually just micro-batching). Actually I wonder we can check is_first_microbatch in kwargs to skip this.

PR #1158, opened 2023-12-28
Add support for `torch.set_default_device` when initializing model parameters

fshp971:
## What does this PR do?
Replace `torch.Tensor` with `torch.empty` when initializing model parameters.

The reason is that `torch.empty` is compatible with `torch.set_default_device` (but `torch.Tensor` is not), thus enable one to efficiently initialize a large ML model with specified GPUs.

(An example is that initialize a Llama2 model with the default `torch.nn.init.xavier_normal_` method is extremely slow on CPUs, but very fast on GPUs.)

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review


mhjacobson:
I just hit this myself, and while I am not an expert in this particular project, I think it would be a good idea.

PR #1174, opened 2024-04-03
add context parallel group init to mp init

amylittleyang:
## What does this PR do?
follow up from [D55538929](https://www.internalfb.com/diff/D55538929).
update model_parallel/initialize.py to also initialize context parallel groups.
original change made in this PR: https://github.com/fairinternal/llm_inference/pull/333/files

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/fairscale/nn/model_parallel/initialize.py
+++ b/fairscale/nn/model_parallel/initialize.py
@@ -34,17 +34,21 @@ _MODEL_PARALLEL_GROUP = None
 _DATA_PARALLEL_GROUP = None
 # Pipeline parallel group that the current rank belongs to.
 _PIPELINE_PARALLEL_GROUP = None
-
 _PIPELINE_PARALLEL_RANKS = None
+
+_CONTEXT_PARALLEL_GROUP = None
+_CONTEXT_PARALLEL_GROUP_RANKS = None
 
 
 def initialize_model_parallel(
-    model_parallel_size_: int,
+    model_parallel_size: int,
+    context_parallel_size: int = 1,
     pipeline_length: int = 1,
     *,
     model_parallel_backend: Optional[str] = None,
+    cp_backend: Optional[str] = None,
     pipeline_backend: Optional[str] = None,
-    ddp_backend: Optional[str] = None
+    ddp_backend: Optional[str] = None,
 ) -> None:
     """
     Initialize model data parallel groups.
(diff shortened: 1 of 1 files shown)


fairscale/nn/model_parallel/initialize.py
```diff
@@ -88,32 +89,119 @@ def initialize_model_parallel(
...
+                    _PIPELINE_PARALLEL_GROUP = group
+                    _PIPELINE_PARALLEL_RANKS = ranks
+
+
+    # Build the context parallel groups.
+    global _CONTEXT_PARALLEL_GROUP
+    global _CONTEXT_PARALLEL_GROUP_RANKS
+    global _CP_PREV_RANK
+    global _CP_NEXT_RANK
+    global _CP_ZERO_RANK
```
jasonjk-park, line 130: wondering why we add these?
all these can be derived with CP GROUP RANKS (in fact, PP does that for similar functionalities)

we can always utilize @functools.lru_cache if performance is concern?

fairscale/nn/model_parallel/initialize.py
```diff
@@ -88,32 +89,119 @@ def initialize_model_parallel(
...
+                    _CP_PREV_RANK = ranks[(index + 1) % len(ranks)]
+                    _CP_NEXT_RANK = ranks[index - 1]
+                    _CP_ZERO_RANK = ranks[0]
+
+
+def model_parallel_is_initialized() -> bool:
+    """Check if model and data parallel groups are initialized."""
+    if _MODEL_PARALLEL_GROUP is None or _DATA_PARALLEL_GROUP is None or _PIPELINE_PARALLEL_GROUP is None:
+        return False
+    return True
```
jasonjk-park, line 153: defined twice here and below?

also, should we add context parallel group to condition?

fairscale/nn/model_parallel/initialize.py
```diff
@@ -88,32 +89,119 @@ def initialize_model_parallel(
...
+
+
+def get_context_parallel_group():
```
jasonjk-park, line 156: add return type annotation for all newly added functions?

fairscale/nn/model_parallel/initialize.py
```diff
@@ -183,6 +271,5 @@ def destroy_model_parallel() -> None:
     _DATA_PARALLEL_GROUP = None
     global _PIPELINE_PARALLEL_GROUP
     _PIPELINE_PARALLEL_GROUP = None
-
-    global _PIPELINE_PARALLEL_RANKS
-    _PIPELINE_PARALLEL_RANKS = None
+    global _CONTEXT_PARALLEL_GROUP
+    _CONTEXT_PARALLEL_GROUP = None
```
jasonjk-park, line 275: should we also make all other CP based global variables None here?

fairscale/nn/model_parallel/initialize.py
```diff
@@ -88,32 +89,119 @@ def initialize_model_parallel(
...
+    ), "context parallel group is not initialized"
+    return _CP_ZERO_RANK
+
+
+def get_context_parallel_world_size():
+    """Return world size for the context parallel group."""
+    if torch.distributed.is_available() and torch.distributed.is_initialized():
+        return torch.distributed.get_world_size(group=get_context_parallel_group())
+    else:
+        return 0
```
jasonjk-park, line 196: is 0 right default value? usually world size returns 1 by default?

fairscale/nn/model_parallel/initialize.py
amylittleyang: Sounds good let me move this part into fbcode. It's probably for convenience in the original PR where they put all comms related changes in this file.

fairscale/nn/model_parallel/initialize.py
amylittleyang: Good catch!

fairscale/nn/model_parallel/initialize.py
amylittleyang: We should return torch.distributed.get_world_size(...) directly here. `get_context_parallel_group()` already checks the CP pg is initialized.

fairscale/nn/model_parallel/initialize.py
```diff
@@ -88,41 +94,87 @@ def initialize_model_parallel(
...
+
+
+def get_context_parallel_world_size() -> int:
+    """Return world size for the context parallel group."""
+    return torch.distributed.get_world_size(group=get_context_parallel_group())
+
+
+def get_context_parallel_rank() -> int:
+    """Return my rank for the context parallel group."""
+    return torch.distributed.get_rank(group=get_context_parallel_group())
```
jasonjk-park, line 175: looks like line 161-165 is the same function - maybe remove this?

jasonjk-park (approved):

Issue #1195, opened 2025-04-09
VocabParallelEmbedding fails if torch.set_default_device() was used: "Expected all tensors to be on the same device"


dvrogozh:
With:
* (fairscale) https://github.com/facebookresearch/fairscale/commit/bbce559da058ef34cd9da6d9d0d9320e7b380a07
* https://github.com/pytorch/pytorch/commit/cfab04d01b734ae48e0dd1ce957a4fad793ea3a2 (pytorch 2.8 candidate)

On:
* Nvidia A10 GPU
* Intel Data Center GPU Max (PVC)

Consider the following script (for Intel `s/cuda/xpu` and `s/nccl/xccl/` is needed). Note `torch.set_default_device()` usage which alters device where tensors are allocated by default.
```
$ cat test.py
import os
import torch
from fairscale.nn.model_parallel.initialize import initialize_model_parallel
from fairscale.nn.model_parallel.layers import VocabParallelEmbedding

torch.distributed.init_process_group("nccl")
initialize_model_parallel(2)
torch.set_default_device("cuda:0")
local_rank = int(os.environ.get("LOCAL_RANK", 0))
torch.cuda.set_device(local_rank)
print(f">>> device={torch.cuda.current_device()}")
v = VocabParallelEmbedding(10, 10, init_method=lambda x: x)
torch.distributed.destroy_process_group()
```

Running this example >=2 GPU devices (won't reproduce on 1 GPU due to different code paths executed in fairscale) will fail with tensor placement error ("Expected all tensors to be on the same device"):
```
$ torchrun --nproc-per-node 2 test.py

[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/dvrogozh/tmp/a.py", line 13, in <module>
[rank0]:     v = VocabParallelEmbedding(10, 10, init_method=lambda x: x)
[rank0]:   File "/home/dvrogozh/git/fairscale/fairscale/nn/model_parallel/layers.py", line 125, in __init__
[rank0]:     _initialize_affine_weight(
[rank0]:   File "/home/dvrogozh/git/fairscale/fairscale/nn/model_parallel/layers.py", line 78, in _initialize_affine_weight
[rank0]:     torch.cat(my_weight_list, dim=partition_dim, out=weight)
[rank0]:   File "/home/dvrogozh/git/pytorch/pytorch/torch/utils/_device.py", line 100, in __torch_function__
[rank0]:     return func(*args, **kwargs)
[...]

PR #1196, opened 2025-04-09
fix: use torch.empty instead of torch.Tensor

dvrogozh:
Fixes: #1195

`torch.Tensor` is a lower level constructor which allocates tensors on CPU. `torch.set_default_device()` don't change allocation preference. To avoid issues with expected tensors placement they can be created with `torch.empty()` instead.

CC: @ashwinb, @raghotham


ashwinb (approved):
yes I had observed similar issue. thanks for fixing.

ashwinb:
I don't have write permissions on this repository so will have to wait for the maintainers.

